PR #15, opened 2017-11-10
feature/documentation

abelardojarab:
This branch places the documentation source code under the doc directory.

Documentation source code includes:
* doxygen setup for the API
* markdown source files for Sphinx
* Sphinx configuration files.

TODO:
* Sphinx configuration file needs to be updated to point to right directories.
* Add cmake script to generate Sphinx HTML output as a cmake target (make sphinx_html? and make sphinx_pdfs?)
* Seems that best approach is NOT to use readthedocs.org but github.io pages; travis script can generate the HTML and upload it to the associated github.io pages for OPAE project. So this seems a better solution that using readthedocs. Readthedocs has their own installation of Sphinx in contrast as travis where we can install sphinx exactly as we want it


luebbers:
Hm. Is there something in our documentation setup that forces us to use a special sphinx setup? Or can we rather use standard tools (and enable readthedocs.io)? That would mean one less toolchain to maintain.

abelardojarab:
No there is no true need to use our sphinx installation; we can make it depend only on readthedocs.

abelardojarab:
Ok, I have removed the Intel-specific files and the documentation for the tools that are not included with OPAE-SDK

luebbers:
Nice work - that documentation tree looks very clean.

PR #16, opened 2017-11-10
Update Travis CI config to use gcc 5

luebbers:
Travis builds seem to fail when compiling some of the tools on
gcc 4.8.

This installs gcc and g++ 5 on Travis and sets up CC and CXX
environment variables. Done according to
https://docs.travis-ci.com/user/languages/c/#GCC-on-Linux

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,13 @@ script:
   - scripts/test-codingstyle.sh
 addons:
   apt:
+    sources:
+      - ubuntu-toolchain-r-test
     packages:
+      - g++-5
       - libboost-all-dev
       - libjson-c-dev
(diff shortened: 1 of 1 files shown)


southerngs:
What fails with gcc 4.8?  That's what we use internally on CentOS.

luebbers:
Travis errors out with

    g++-4.9: internal compiler error: Killed (program cc1plus)

It builds fine on a separate (non-Travis) VM with the same Ubuntu and compiler versions, so you should be fine.

michael-adler:
It's what I use on our standard EC Linux (SUSE, something from the Pleistocene). It was triggering an internal g++ compiler failure in Travis.

PR #18, opened 2017-11-13
drv_arch: Add note about process termination

tswhison:
and freeing resources (file descriptors).


doc/src/drv_arch/drv_arch.md
```diff
@@ -37,6 +37,14 @@ Function (PF).
 User-space applications can acquire exclusive access to the FME using `open()`,
 and release it using `close()` as a privileged user (root).
 
+.. note::
+
+```
+    In the case that an errant application terminates without freeing
+    the FME/Port resources, Linux closes all file descriptors owned by
+    the terminating process, freeing those resources.
+```
```
r-rojo: What about buffers prepared with a resource? Are those freed too?

tswhison: I was trying to be vague in my use of the word "resources".  Yes, everything that a process does with the device after opening the device node is associated with that process and tracked in data structures within the driver.  The memory is freed when the device handle (file descriptor) is being released.

r-rojo: Just saying'... It might be worth noting somewhere about buffers not being freed until last associated resource is closed (when they're opened in shared mode). This may not be the right spot for that but I don't know if that's documented anywhere. I did for `buffer` in all markdown files and found no mention of that.

tswhison:
Github says that I'm not authorized to merge this pull request.  Can somebody help me out with access?

doc/src/drv_arch/drv_arch.md
luebbers: Could we add that in the Doxygen of the `fpgaOpen()` call (that's where `FPGA_OPEN_SHARED` is documented), and also add a link in the Doxygen of `fpgaClose()`?

luebbers: PR #31 is taking care of that, so I'll go ahead and merge this.

PR #19, opened 2017-11-14
ASE: Code refactoring --changing global variables to local variables to static local variables

lzhan55:
- Changed global variables in protocol_backend.c such as session_empty, glbl_session_id, f_usrclk, glbl_umsgmode, umsg_mode_msg and app_run_cmd to be local varialble or static local variables; 
- Moved global variable cfg declaration from ase_common.h to protocol_backend.c;
- Fixed a memory leak issue in variable app_run_cmd

Passed regression test on Linux VM locally

### DIFF
--- a/ase/sw/ase_common.h
+++ b/ase/sw/ase_common.h
@@ -259,11 +259,6 @@ extern char *ase_workdir_path;
 
 // Timestamp IPC file
 #define TSTAMP_FILENAME ".ase_timestamp"
-extern char tstamp_filepath[ASE_FILEPATH_LEN];
-extern char *glbl_session_id;
-
-// CCIP Warnings and Error stat location
-extern char ccip_sniffer_file_statpath[ASE_FILEPATH_LEN];
 
 // READY file name
 #define ASE_READY_FILENAME ".ase_ready.pid"
@@ -670,7 +665,6 @@ struct ase_cfg_t {
 	int usr_tps;
 	int phys_memory_available_gb;
 };
-struct ase_cfg_t *cfg;
 
 // ASE config file
 // #define ASE_CONFIG_FILE "ase.cfg"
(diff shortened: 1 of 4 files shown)


michael-adler:
Have you gone too far with some of the externs moving out of ase_common.h or am I missing something? For example, removing "cfg" from ase_common.h forces you to declare extern versions in multiple modules. Does the linker check that they are the same type? The same is true of some of the other variables, like tstamp_filepath. Some you removed are a good choice, like glbl_session_id, which appears to be used only in one module.

lzhan55:
I moved cfg from ase_common.h to protocol_backend.c. Global variables usually should not be declared in header files, but it could be declared as an extern global variable in a header file though. Global variable cfg is only used by two modules --- protocol_backend.c and mem_model.c. I added "extern struct ase_cfg_t *cfg; " in mem_model.c. 

I did ASE regression tests on Linux by rebuilding everything and running the regression test, and everything worked fine. 

Thank you for your feedbacks!
-Tracy

michael-adler:
The problem is that the linker does no type checking. At least that's how it used to be. When you have an "extern" variable in a header because the variable is shared, the C compiler enforces the type equivalence of all uses of the variable. If you just put extern in every file that uses the variable and a declaration in one of them then you've lost that level of type safety.

I'm all for your removing non-global variables form ase_common.h. Just observing that you likely went a little overboard.

lzhan55:
Hi Michael,

I made some changes for "extern char tstamp_filepath[ASE_FILEPATH_LEN]; " based on your suggestions. Thank you!

Could you take a look at the changes again?

Thanks,
Tracy

rahulrs:
Has this been verified to work on both VCS and Modelsim-SE and Modelsim 32-bit?

PR #20, opened 2017-11-14
Enable warnings when building OPAE

southerngs:
I noticed that warnings are not enabled when building OPAE. I
recommend that we enable -Werror so that we are clean with all
warnings at all time. However, there are too many warnings right
now to enable.

In the interim I recommend at least enabling -Wall and probably
-Wextra.

It looks like this was the original intent, but at least on
the systems I've tested the various types of flags don't set
the ones that are used during compile.  Those come from
CMAKE_C_FLAGS and CMAKE_CXX_FLAGS. For now I'm setting them
explicity to -Wall -Wextra.  I would like to see -Werror
enabled as soon as possible.


luebbers:
This change is already included in commit 21f6836.

PR #21, opened 2017-11-15
ASE: Refactoring 3 -- Removed redundant buffer definition of wsmeta_t in app_backend.c

lzhan55:
Removed redundant buffer definition of wsmeta_t in app_backend.c  for MMIO and UMAS buffers, fixed some memory leak issues related to buffers, and changed some global variables to be local variables.

Passed regression tests on Linux with VCS.


ase/sw/app_backend.c
```diff
@@ -1340,56 +1342,76 @@ bool deallocate_buffer_by_index(int search_index)
 {
     FUNC_CALL_ENTRY;
     bool value;
-    uint64_t *bufptr = (uint64_t *) NULL;
-    struct wsmeta_t *wsptr;
+	struct buffer_t *bufptr = (struct buffer_t *) NULL;
```
rahulrs: 'wsid' may not necessarily be buffer_t->index. Is this difference being accounted for?

For eg.
**In MCP mode,** 
OPAE API buffer wsid=0 maps to ASE API Buffer 2 (this offset is due to MMIO and UMAS regions)
**In DCP mode,** 
OPAE API buffer wsid=0 = ASE API Buffer 1 (this offset is due to MMIO region)

This can cause the wrong region to get deallocated.

ase/sw/app_backend.c
```diff
@@ -1340,56 +1342,76 @@ bool deallocate_buffer_by_index(int search_index)
...
+    trav_ptr = buf_head;
     while (trav_ptr != NULL) {
-	if (trav_ptr->index == wsid) {
-	    bufptr =
-		(struct buffer_t *) trav_ptr->buf_structaddr;
-	    break;
-	} else {
-	    trav_ptr = trav_ptr->next;
-	}
+		if (trav_ptr->index == wsid) {
```
rahulrs: Application 'wsid' does not equal trav_ptr->index.
For eg. from simulation log of hello_fpga/NLB0, the DSM, (WSID=0) reads:

**Buffer 2** Allocated  (located /dev/shm//buf0.2671733310964392) =>
                Host App Virtual Addr  = 0x7f6a5d605000
                HW Physical Addr       = 0xc3ca00000
                HW CacheAligned Addr   = 0x30f28000
                Workspace Size (bytes) = 2097152

ase/sw/ase_common.h
```diff
@@ -404,7 +392,7 @@ extern uint64_t *umsg_umas_vbase;
 // ASE buffer valid/invalid indicator
 // When a buffer is 'allocated' successfully, it will be valid, when
 // it is deallocated, it will become invalid.
-#define ASE_BUFFER_VALID        0xFFFF
+#define ASE_BUFFER_VALID        0x1
```
rahulrs: Enum?

ase/sw/ipc_mgmt_ops.c
```diff
@@ -38,6 +38,9 @@
 // IPC control list
 static char ipclist_filepath[ASE_FILEPATH_LEN];
 
+// Local IPC log
+FILE *local_ipc_fp;
```
rahulrs: Unrelated change

ase/sw/protocol_backend.c
```diff
@@ -443,6 +430,13 @@ void initialize_fme_dfh(struct buffer_t *buf)
 {
 	FUNC_CALL_ENTRY;
 
+	// Capability CSRs
```
rahulrs: Unrelated changes

ase/sw/app_backend.c
```diff
@@ -45,9 +45,9 @@ struct buffer_t *mmio_region;
 // UMAS region
 struct buffer_t *umas_region;
 
-// Workspace metadata table
-struct wsmeta_t *wsmeta_head = (struct wsmeta_t *) NULL;
-struct wsmeta_t *wsmeta_end = (struct wsmeta_t *) NULL;
+// Record of MMIO and UMAS memory
+struct buffer_t *buf_head = (struct buffer_t *) NULL;
+struct buffer_t *buf_end = (struct buffer_t *) NULL;
```
luebbers: Can you elaborate on the reason for this change? Are you renaming wsmeta to buffer for consistency? Or is there already a buffer definition that can cover what wsmeta does, and you're consolidating?

I'm not saying it's not a good idea, I'd just like to understand your reasoning behind it. It's always a good idea to outline the "why" in your commit messages so that developers not familiar with the changes can understand the reasons for them.

You can also add comments to your on code in GitHub to explain changes if you think that's worthwhile for making the review go smoothly.

ase/sw/app_backend.c
```diff
@@ -1231,12 +1234,7 @@ void allocate_buffer(struct buffer_t *mem, uint64_t *suggested_vaddr)
     ase_buffer_info(mem);
 #endif
 
-    // book-keeping WSmeta // used by ASEALIAFU
-    struct wsmeta_t *ws;
-    ws = (struct wsmeta_t *) ase_malloc(sizeof(struct wsmeta_t));
-    ws->index = mem->index;
-    ws->buf_structaddr = (uint64_t *) mem;
-    append_wsmeta(ws);
+    append_buf(mem);
```
luebbers: This looks like you're tracking "mem" directly in a list, instead of allocating a new `wsmeta_t` entry referencing it. Are they equivalent?

ase/sw/app_backend.c
luebbers: `wsptr->index` contains the same value as `buf->index` (see l. 1237 in the original), so shouldn't a search for either work?

ase/sw/app_backend.c
luebbers: Can you point me to where that differentiation between application WSIDs and ASE workspace indices is made in the original code? I couldn't see it being removed in the diff.

ase/sw/ase_common.h
luebbers: Is this a magic sequence to detect unitialized memory (in which case it should have values that look a bit more magic)? Or is it just a boolean?

ase/sw/ipc_mgmt_ops.c
luebbers: Agree. This should be in a separate commit.

ase/sw/protocol_backend.c
luebbers: See above

ase/sw/app_backend.c
rahulrs: In allocate_buffer() asebuf_index_count keeps track of all buffers allocated in the application. ASE internally counts MMIO and UMSG regions as 0 and 1 respectively. App index starts from 1 or 2 depending on platform type.

lzhan55: From the original code, there is no difference regarding index between ws->index and buf->index. Please check the line at 1237 of app_backend.c. No matter how it is interpreted on different platforms, the index is same in wsmeta and buffer as shown in line 1237.. So it is not necessary to create another data structure of wsmeta_t to keep track of the index.

ase/sw/app_backend.c
lzhan55: I could not see any difference of application WSIDs and buffer index. There is no special handling of index regarding different use cases. At line 1202 in allocate_buffer() function, buffer index was incremented by asebuf_index_count. No matter what the use case of buffers are, MMIO or user buffers, it all invokes allocate_buffer() function to get its index based on global counter of asebuf_index_count.

ase/sw/ase_common.h
lzhan55: Boolean is enough for this case. I changed it to be 0x01 because wsmeta_t's valid field use a '0' and '1'. So I didn't want to break any code in case it is used in some special way.

ase/sw/ipc_mgmt_ops.c
lzhan55: I can make two commits. One commit for wsmeta_t changes and another commit for the rest.

ase/sw/protocol_backend.c
lzhan55: Agree.

ase/sw/app_backend.c
lzhan55: I think that struct wsmeta_t definition is not needed: 
1. struct buffer_t has all the information needed to keep track of the buffers. Adding another layer of buffer definion with wsmeta_t make the code not readable. 
2. There were memory leak issues since all the memories allocate for wsmeta_t were not released.  
Please check the buffer_t and wsmeta_t definition below. Writing simple software code while achieving the same goal is better. 

// Buffer information structure --
//   Be careful of alignment within this structure!  The layout must be
//   identical on both 64 bit and 32 bit compilations.
struct buffer_t			//  Descriptiion                    Computed by
{				// --------------------------------------------
	int32_t index;		// Tracking id                     | INTERNAL
	int32_t valid;		// Valid buffer indicator          | INTERNAL
	uint64_t vbase;		// SW virtual address              |   APP
	uint64_t pbase;		// SIM virtual address             |   SIM
	uint64_t fake_paddr;	// unique low FPGA_ADDR_WIDTH addr |   SIM
	uint64_t fake_paddr_hi;	// unique hi FPGA_ADDR_WIDTH addr  |   SIM
	int32_t is_privmem;	// Flag memory as a private memory |
	int32_t is_mmiomap;	// Flag memory as CSR map          |
	int32_t is_umas;	// Flag memory as UMAS region      |
	uint32_t memsize;	// Memory size                     |   APP
	char memname[ASE_FILENAME_LEN];	// Shared memory name              | INTERNAL
	struct buffer_t *next;
};


/*
 * Workspace meta list
 */
struct wsmeta_t {
	int index;
	int valid;
	uint64_t *buf_structaddr;
	struct wsmeta_t *next;
};Leave a comment

ase/sw/app_backend.c
lzhan55: Yes, it is equivalent.

ase/sw/app_backend.c
rahulrs: Can you quickly confirm this by doing these steps:
1. Add "#define ASE_DEBUG 1" to top of sw/ase_common.h
2. Rebuild ASE. This will generate a few more log files, one of them will be $ASE_WORKDIR/ase_pagetable.log
3. Correlate the buffer allocation index numbers in ccip_transactions and the file in 2.

ase/sw/ase_common.h
rahulrs: I suggest leave this as 32-bit number for 2 reasons:
1. This has to be packed into a struct and communicated over a named pipe, so things need to be predictable.
2. Future expansion of stateful data as needed.

lzhan55: In struct buffer-t definition, "int32_t valid;' is used for buffer->valid bit. No matter what value you assign it to this field, it will use a 32 bit data for it.

lzhan55:
I enabled "#define ASE_DEBUG 1 " at the beginning of ase_common.h , and hello_fpga test still passes. We should merge this changes today.

ase/sw/ase_common.h
rahulrs: I have had trouble with certain old versions of VCS, using bool(s) caused me problem. I suggest to keep the refactoring to coding styles. There are some portions of these structs that map to SV-DPI, and they can be particularly tricky and may render the code inoperable. It is extremely difficult to narrow down problems to particular versions of Modelsim and VCS.

diptisherlekar:
@lzhan55  ASE_DEBUG is used to debug ASE ,if an error occurs. It is not used for Validation of ASE .

rahulrs:
I suggest waiting till the CI infrastructure is completely regressing the code.

ase/sw/app_backend.c
lzhan55: When I enabled "#define ASE_DEBUG 1", it generated a zero size ase_pagetable.log file. Is there anything missing here?

lzhan55:
What else is needed for CI infrastructure enabling? To enable CI to automatically do regression tests on branches? Nakul sent an email yesterday saying that CI is ready now.

diptisherlekar:
@lzhan55 If Jenkins has passed the test and if the DCP tests pass , you are good to go . Please check with Modelsim 32 bit and 64bit on your end .

ase/sw/ase_common.h
lzhan55: OK, will keep the "#define ASE_BUFFER_VALID        0xFFFF ". Thank you for your feedbacks.

diptisherlekar:
@lzhan55 CI infrastructure checks for VCS simulation with different AFU's . It doesnt check Modelsim . Developer is needed to check for this on his end , till it gets enabled on the CI.

ase/sw/app_backend.c
lzhan55: @rahulrs Please provide code evidence for "'wsid' may not necessarily be buffer_t->index". If evidence is not provided by Sunday (11/26) and DCP tests pass, we will merge this pull request branch into master branch.

ase/sw/app_backend.c
lzhan55: @rahulrs Please provide code evidence for "'wsid' may not necessarily be buffer_t->index". If evidence is not provided by Sunday (11/26) and DCP tests pass, we will merge this pull request branch into master branch

deepakcu84:
Full DCP run: http://sw-pert.altera.com/pert/pert.php?test_run_id=2050289

Looks good. There are few failures due to infrastructure issues.

lzhan55:
ASE: Removed redundant buffer definition of wsmeta_t in app_backend.c

- Replaced struct wsmeta_t list with struct buffer_t in app_backend.c to make it consistant with simulator side buffer list 
- Fixed buffer_t memory leak issue at app_backend.c
- Changed some global variables to local variables in protocol_backend.c

PR #22, opened 2017-11-15
Add package release version customization

rpSTTD:
Add package release version customization


luebbers:
Okay, changes are in.

PR #29, opened 2017-11-20
Fix pthreads compilation problem and compiler flags

michael-adler:
Changing the order of include files in event.c causes pthreads.h to be
included with _GNU_SOURCE set (libopae/src/common_int.h), which adds
the definition of PTHREAD_MUTEX_RECURSIVE.

Drop some compiler warnings that were generating lots of errors in
system includes.

### DIFF
--- a/cmake/modules/compiler_config.cmake
+++ b/cmake/modules/compiler_config.cmake
@@ -103,8 +103,8 @@ if (C_SUPPORTS_NO_FORMAT_TRUNCATION)
 endif()
 
 check_c_compiler_flag("-Wno-unused-parameter"
-  C_SUPPORTS_NO_FORMAT_TRUNCATION)
-if (C_SUPPORTS_NO_FORMAT_TRUNCATION)
+  C_SUPPORTS_NO_UNUSED_PARAMETER)
+if (C_SUPPORTS_NO_UNUSED_PARAMETER)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unused-parameter")
 endif()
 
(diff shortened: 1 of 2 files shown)


cmake/modules/compiler_config.cmake
```diff
@@ -102,6 +102,12 @@ if (C_SUPPORTS_NO_FORMAT_TRUNCATION)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-format-truncation")
 endif()
 
+check_c_compiler_flag("-Wno-unused-parameter"
+  C_SUPPORTS_NO_FORMAT_TRUNCATION)
+if (C_SUPPORTS_NO_FORMAT_TRUNCATION)
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unused-parameter")
+endif()
+
```
luebbers: I believe the CMake variable should be named something like `C_SUPPORTS_NO_UNUSED_PARAMETER` instead of `C_SUPPORTS_NO_FORMAT_TRUNCATION`?

cmake/modules/compiler_config.cmake
```diff
@@ -126,6 +132,12 @@ if (CXX_SUPPORTS_NO_UNKNOWN_PRAGMAS)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unknown-pragmas")
 endif()
 
+check_cxx_compiler_flag("-Wno-unused-local-typedefs"
+  CXX_SUPPORTS_NO_UNKNOWN_PRAGMAS)
+if (CXX_SUPPORTS_NO_UNKNOWN_PRAGMAS)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-local-typedefs")
+endif()
```
luebbers: Same here (`-CXX_SUPPORTS_NO_UNUSED_LOCAL_TYPEDEFS` instead of `CXX_SUPPORTS_NO_UNKNOWN_PRAGMAS`)?

PR #30, opened 2017-11-20
[ASE] Transaction ordering config change for integrated and discrete

rahulrs:
This change modifies the way in which transaction order control is selected. CCI-P specification does not imply ordering of responses in any platform. ASE mimics that behavior.
An infinite bandwidth mode is set to override the out-of-order behavior for high-throughput AFU tuning.

Signed-off-by: Rahul R Sharma <rahul.r.sharma@email.com>

### DIFF
--- a/ase/rtl/ase_pkg.sv
+++ b/ase/rtl/ase_pkg.sv
@@ -278,8 +278,6 @@ package ase_pkg;
     */
    // Number of transactions in latency scoreboard
    parameter LATBUF_NUM_TRANSACTIONS = 32;
-   // Radix of latency scoreboard radix
-   parameter LATBUF_COUNT_WIDTH      = $clog2(LATBUF_NUM_TRANSACTIONS) + 1;
    // ASE_fifo full threshold inside latency scoreboard
    parameter LATBUF_FULL_THRESHOLD   = LATBUF_NUM_TRANSACTIONS - 5;
    // Radix of ASE_fifo (subcomponent in latency scoreboard)
--- a/ase/rtl/ccip_emulator.sv
+++ b/ase/rtl/ccip_emulator.sv
@@ -1732,9 +1732,8 @@ module ccip_emulator
        .DEBUG_LOGNAME       ("latbuf_ch0.log"),
        .NUM_WAIT_STATIONS   (LATBUF_NUM_TRANSACTIONS),
        .NUM_STATIONS_FULL_THRESH (LATBUF_FULL_THRESHOLD),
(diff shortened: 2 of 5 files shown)


michael-adler:
I try this against the bug we had this afternoon and make sure almost full is now triggered.

diptisherlekar:
@michael-adler @rahulrs The channel 0 and 1 depths are different for DCP and MCP ?

michael-adler:
There is back pressure now with 1 beat requests, but not with multi-beat requests. I suspect this is because of the way READ and WRITE_get_latbuf_unroll_put_outfifo handle VH0/VH1 traffic. VHx traffic seems to come out of the latency FIFO all in one beat, independent of length. Bus bandwidth is better with 4 line requests, but not 4x better.

rahulrs:
@michael-adler : Do you have any suggestions for this commit.

michael-adler:
Was the motivation for treating VH0/VH1 traffic differently in READ/WRITE_get_latbuf_unroll_put_outfifo to attempt to model bandwidth? Was it trying to account for the fact that PCIe bandwidth improves with MCL=2 or 4? If yes, then I suggest you try simply removing that and treating all packets like VL0. It's more important that we test code with OOO responses and back pressure than bandwidth ratios. The current model isn't correct anyway.

Another option would be to adopt the VC round robin algorithm actually used by the BBS: only process a PCIe packet one out of every 4 cycles and QPI every other cycle. I looked at the code trying to figure out where that would go. The easiest is probably in the same functions as above, though I didn't really think it through.

rahulrs:
Are you referring to outoforder_wrf_channel.sv lines 428-430?

michael-adler:
No. Those seem to enforce fairness in a manner similar to BBS but only when there is contention among channels. I'm under the impression that somewhere in the BBS there is a MUX that is reading incoming buffers, going round-robin independent of traffic. That strategy might work here if applied in the rate place to rate limit each channel.

It's still likely simpler just to get rid of the special-case PCIe code 969-988 and 1428-1446.

rahulrs:
Or, if we turn this arbitration block into just VH1, would that be sufficient for the discrete FPGA case ?

michael-adler:
I don't think you should special case discrete. You'd have the same problem on Xeon+FPGA if an AFU used only a single PCIe channel.

luebbers:
Is this still under investigation?

luebbers:
Any updates?

rahulrs:
Lets close this for now -- I will respond with an update.

PR #31, opened 2017-11-20
doc: add note about freeing shared resources

tswhison:
to the doxygen of fpgaOpen().

### DIFF
--- a/common/include/opae/access.h
+++ b/common/include/opae/access.h
@@ -54,6 +54,8 @@ extern "C" {
  * @param[in]  flags    One of the following flags:
  *                        * FPGA_OPEN_SHARED allows the resource to be opened
  *                          multiple times (not supported in ASE)
+ *                          When the last handle is closed with fpgaClose(),
+ *                          all of the shared resources are released.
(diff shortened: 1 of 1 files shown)


common/include/opae/access.h
```diff
@@ -54,6 +54,8 @@ extern "C" {
  * @param[in]  flags    One of the following flags:
  *                        * FPGA_OPEN_SHARED allows the resource to be opened
  *                          multiple times (not supported in ASE)
+ *                          When the last handle is closed with fpgaClose(),
+ *                          all of the shared resources are released.
```
luebbers: Should we generalize that? Technically, you don't need to close it with `fpgaClose()` - if your app crashes, it'll still close the handle.

tswhison: with fpgaClose(), (or if the process terminates unexpectedly) all of..  ?

r-rojo: You could add:
_Shared resources (including buffers) are released when all associated handles have been closed (either explicitly with fpgaClose or by the process termination)._

PR #32, opened 2017-11-20
Allow MMIO access without explicit call to mapMMIO()

luebbers:
This commit changes the semantics of the MMIO API so that it is no
longer necessary to explicitly call fpgaMapMMIO() before using
fpgaMMIORead*() or fpgaMMIOWrite*().

Any of these functions will now create the respective mapping befor
performing its operation (read/write/return pointer). This simplifies
their usage, and allows future use of the read/write APIs on platforms
that do not support memory mapping.

For ASE, the same semantics have been implemented; the variable tracking
the "MMIO is mapped" state has been moved from a global variable into
the handle.


libopae/src/mmio.c
```diff
@@ -45,39 +45,40 @@
...
 
-fpga_result __FPGA_API__ fpgaWriteMMIO32(fpga_handle handle,
-					 uint32_t mmio_num,
-					 uint64_t offset,
-					 uint32_t value)
+static fpga_result port_get_region_info(fpga_handle handle,
+				 uint32_t mmio_num,
+				 uint32_t *flags,
+				 uint64_t *size,
+				 uint64_t *offset)
```
luebbers: This diff result is really unfortunate for a review - I simply moved `port_get_region_info()` and `port_mmap_region()` up top because the public API calls now depend on these.

libopae/src/mmio.c
```diff
@@ -128,16 +130,96 @@ fpga_result __FPGA_API__ fpgaReadMMIO32(fpga_handle handle,
...
+		}
+	}
+
+	return FPGA_OK;
+}
+
+/* Lazy mapping of MMIO region (only map if not already mapped) */
+static fpga_result find_or_map_wm(fpga_handle handle, uint32_t mmio_num,
+				struct wsid_map **wm_out)
+{
```
luebbers: This function will either return a valid wm pointer for the `mmio_num` supplied, or try to map the respective MMIO region and return the pointer then.

common/include/opae/mmio.h
```diff
@@ -138,24 +138,26 @@ fpga_result fpgaReadMMIO32(fpga_handle handle,
  * Map MMIO space
  *
  * This function will return a pointer to the specified MMIO space of the
- * target object in process virtual memory. Some MMIO spaces may be restricted
- * to privileged processes, depending on the used handle and type.
+ * target object in process virtual memory, if supported by the target. Some
+ * MMIO spaces may be restricted to privileged processes, depending on the used
+ * handle and type.
  *
  * After mapping the respective MMIO space, you can access it either through
```
michael-adler: There is no "or" to the "either" here. I assume it was supposed to say "or the function calls", though worded better than that.

Also, the Doxygen for the fpgaReadMMIO32, etc. calls still says that fpgaMapMMIO() must be called first.

libopae/src/mmio.c
```diff
@@ -87,38 +88,39 @@ fpga_result __FPGA_API__ fpgaWriteMMIO32(fpga_handle handle,
...
+	if (vaddr == MAP_FAILED) {
+		FPGA_MSG("Unable to map MMIO region. Error value is : %s",
+			 strerror(errno));
 		result = FPGA_INVALID_PARAM;
 		goto out_unlock;
 	}
 
-	*value = *((volatile uint32_t *) ((uint8_t *)wm->offset + offset));
+	/* Save return address */
+	if (vaddr)
```
tswhison: We no longer need this check with the addition of line 106.

libopae/src/mmio.c
```diff
@@ -87,38 +88,39 @@ fpga_result __FPGA_API__ fpgaWriteMMIO32(fpga_handle handle,
...
-		FPGA_MSG("Trying to access MMIO before calling fpgaMapMMIO()");
-		result = FPGA_INVALID_PARAM;
-		goto out_unlock;
-	}
-
-	if (offset > wm->len) {
-		FPGA_MSG("offset out of bounds");
+	/* Map MMIO memory */
+	addr = (void *) mmap(NULL, size, flags, MAP_SHARED, _handle->fddev, offset);
+	if (vaddr == MAP_FAILED) {
```
tswhison: Is this correct?  addr is catching the result of mmap, yet vaddr is compared to MAP_FAILED.  I think addr should be compared..

PR #35, opened 2017-11-21
Add packager installation support

rpSTTD:
Add packager installation support.


r-rojo:
This is really nice. However, I would like to see:
* A target defined that calls the process
  * This target has dependencies on the Python files that are to be included
* Explicitly use the same list of Python files instead of zipping everything and then excluding certain files
* A CMake macro defined that does all of this - so we can reuse

luebbers:
Any update?

michael-adler:
The only real problem with this method is the Python modules can't be used by other scripts. I was going to write a script that uses afu.py to load the JSON and emit header files with metadata, like the AFU UUID, to C and Verilog header files. Now I either have to add that feature to the packager binary, copy afu.py, or just import the AFU's JSON file as raw JSON.

I didn't make all the changes Rodrigo suggested, but did clean up the script so that it would be easier to turn into a macro. Specifically:

- The zipped "binary" is written to the CMake build's bin directory. This is valuable since it remains possible to run with that bin directory on PATH for testing instead of having to install every time.
- All building is now into the CMake tree. No more files written to the source tree.
- zipped all the needed files. Some subdirectories were missed in the earlier version.

michael-adler:
Please review again. All the objections have been addressed.

tools/packager/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+## INTERRUPTION)  HOWEVER CAUSED  AND ON ANY THEORY  OF LIABILITY,  WHETHER IN
+## CONTRACT,  STRICT LIABILITY,  OR TORT  (INCLUDING NEGLIGENCE  OR OTHERWISE)
+## ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
+## POSSIBILITY OF SUCH DAMAGE.
+
+# Generate executable python zip file
+zip_files(
+  ${CMAKE_CURRENT_BINARY_DIR}/packager.zip
+  ${CMAKE_CURRENT_SOURCE_DIR}
+  "*.py *.pyc metadata schema")
```
rpSTTD: we shall pass in exact python file list. No *.pyc should be packaged.

PR #36, opened 2017-11-21
fpgainfo: refactor use of object_id property/change case

r-rojo:
* Change `object_ID` to `object_id` to be consistent with other Python names.
* Move `object_id` property to `sysfs_resource` base class and remove printing of extra
information in `object_id` property. The property should return the value
only. Any other formatting should be done outside of the property.


r-rojo:
The `object_id` property should return the number and not necessarily a formatted string. I play to add a `type` property as well that can also be displayed when printing out the information.

r-rojo:
I planned (and would prefer) to do that in another PR as that requires adding the `object_type` property.

PR #37, opened 2017-11-21
cmake: fix build types; remove masking of warnings

tswhison:
Change the following cmake variables from CACHE'ed to uncached so that
cmake uses them to set the compiler flags:

CMAKE_C/CXX_FLAGS_DEBUG
CMAKE_C/CXX_FLAGS_RELWITHDEBUGINFO
CMAKE_C/CXX_FLAGS_RELEASE
CMAKE_C/CXX_FLAGS_MINSIZEREL

Remove set() directives that were attempting to silence certain classes
of warnings, as using them contradicts -Wall.

### DIFF
--- a/cmake/modules/compiler_config.cmake
+++ b/cmake/modules/compiler_config.cmake
@@ -47,33 +47,22 @@ function(SET_CACHED_VARIABLE var)
   set(_cached_vars ${_cached_vars} CACHE INTERNAL "")
 endfunction(SET_CACHED_VARIABLE)
 
-############################################################################
-## GCC specific options ####################################################
-############################################################################
+# Default flags to compiler when build user-space programs.
+# Should come before enabling language.
 
-if(CMAKE_COMPILER_IS_GNUCC)
-  # Default flags to compiler when build user-space programs.
-  # Should come before enabling language.
-  set(CMAKE_C_FLAGS_DEBUG "-g -O0 -Wall -Wextra"
-    CACHE STRING "Compiler flags for debug builds.")
-  set(CMAKE_C_FLAGS_RELWITHDEBINFO "-g -Wall -Wextra"
-    CACHE STRING "Compiler flags for Release With Debug Info builds.")
-  set(CMAKE_C_FLAGS_RELEASE "-Wall"
-    CACHE STRING "Compiler flags for release builds.")
-  set(CMAKE_CXX_FLAGS_DEBUG "-g -O0 -Wall -Wextra"
-    CACHE STRING "C++ compiler flags for debug builds.")
-  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-g -Wall -Wextra"
-    CACHE STRING "C++ compiler flags for Release with Debug Info builds.")
-  set(CMAKE_CXX_FLAGS_RELEASE "-Wall"
-    CACHE STRING "C++ compiler flags for release builds.")
-endif(CMAKE_COMPILER_IS_GNUCC)
+#TODO enable -Werror
 
-# Recommend enabling Werror. Can't enable it yet because we are not warning clean
-#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror")
-#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror")
+set(CMAKE_C_FLAGS_DEBUG            "-g -O0 -Wall -Wextra")
+set(CMAKE_CXX_FLAGS_DEBUG          "-g -O0 -Wall -Wextra")
 
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
+set(CMAKE_C_FLAGS_RELEASE          "-Wall -Wextra")
+set(CMAKE_CXX_FLAGS_RELEASE        "-Wall -Wextra")
+
+set(CMAKE_C_FLAGS_RELWITHDEBINFO   "-g -Wall -Wextra")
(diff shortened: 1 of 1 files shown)


cmake/modules/compiler_config.cmake
```diff
@@ -47,33 +47,22 @@ function(SET_CACHED_VARIABLE var)
...
 
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
+set(CMAKE_C_FLAGS_RELWITHDEBINFO   "-g -Wall -Wextra")
+set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-g -Wall -Wextra")
+
+set(CMAKE_C_FLAGS_MINSIZEREL       "-Os -Wall -Wextra")
+set(CMAKE_CXX_FLAGS_MINSIZEREL     "-Os -Wall -Wextra")
 
 # Check if support for C++ 11 is available
```
southerngs: What level of compiler support are requiring for C++ code?

cmake/modules/compiler_config.cmake
```diff
@@ -47,33 +47,22 @@ function(SET_CACHED_VARIABLE var)
...
-endif(CMAKE_COMPILER_IS_GNUCC)
+#TODO enable -Werror
+
+set(CMAKE_C_FLAGS_DEBUG            "-g -O0 -Wall -Wextra")
+set(CMAKE_CXX_FLAGS_DEBUG          "-g -O0 -Wall -Wextra")
 
-# Recommend enabling Werror. Can't enable it yet because we are not warning clean
-#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror")
-#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror")
+set(CMAKE_C_FLAGS_RELEASE          "-Wall -Wextra")
```
southerngs: I think we should move the setting for `-O2` here instead of having it in the `Defensive compilation for Release` settings at the bottom of the file.

cmake/modules/compiler_config.cmake
```diff
@@ -47,33 +47,22 @@ function(SET_CACHED_VARIABLE var)
...
+set(CMAKE_CXX_FLAGS_RELEASE        "-Wall -Wextra")
```
southerngs: same as previous comment

cmake/modules/compiler_config.cmake
```diff
@@ -47,33 +47,22 @@ function(SET_CACHED_VARIABLE var)
...
 
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
+set(CMAKE_C_FLAGS_RELWITHDEBINFO   "-g -Wall -Wextra")
```
southerngs: Needs `-O2` (or whatever optimization level we want for release)

cmake/modules/compiler_config.cmake
```diff
@@ -47,33 +47,22 @@ function(SET_CACHED_VARIABLE var)
...
+set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-g -Wall -Wextra")
```
southerngs: same as previous

cmake/modules/compiler_config.cmake
tswhison: C++11/14 is required for fpgadiag and will continue to be going forward.

cmake/modules/compiler_config.cmake
tswhison: done

cmake/modules/compiler_config.cmake
tswhison: done

cmake/modules/compiler_config.cmake
tswhison: done

cmake/modules/compiler_config.cmake
tswhison: done

cmake/modules/compiler_config.cmake
luebbers: Are we using any C++14 features that are not supported in GCC 4.9?

From https://gcc.gnu.org/projects/cxx-status.html:

Variable templates | N3651 | 5 | __cpp_variable_templates >= 201304
-- | -- | -- | --
Relaxing requirements on constexpr functions | N3652 | 5 | __cpp_constexpr >= 201304
Member initializers and aggregates | N3653 | 5 | __cpp_aggregate_nsdmi >= 201304
Sized deallocation | N3778 | 5 | __cpp_sized_deallocation >= 201309

r-rojo: All my development has been tested and compiled C++11. I don't think we are using any >= C++14 features.

tswhison: But we have always been building with -std=c++14 afaik. So, I think we would need to perform the exercise of turning it off and seeing what breaks.

r-rojo: All by personal builds/tests have been using
* gcc 4.8.5 (RHEL)
* gcc 4.8.4 (Ubuntu) 

Neither has C++14 so CMake will detect that not allow building with C++14.
Travis CI also uses gcc 4.8.4 by the way...

southerngs: I think we should change the CMake file to alway set `-std=c++11` rather than trying to detect what the compiler supports.  It sounds like we need C++11 features but we aren't using any C++14 features yet.  I'd prefer not to see the settings change depending on the compiler.  Instead this setting should be set based on what our requirements are for the project.  We don't want someone to start adding something that requires a C++14 compiler just because they happen to be using a newer compiler.

However, I'll still approve this PR since this is orthogonal to the other cleanup changes.  But I would prefer setting `std=c++11` all of the time, either in this PR or some other one.

PR #38, opened 2017-11-21
Enable compilation of mock driver and test binary (gtapi)

abelardojarab:
Build gtapi and mock driver; define basic test using combination of mock driver and gtapi.


tests/CMakeLists.txt
```diff
@@ -0,0 +1,181 @@
...
+  message(STATUS "Trying to fetch gtest throught git")
+  find_package(Git REQUIRED)
+  execute_process(COMMAND "${GIT_EXECUTABLE}"
+    "clone"
+    "--branch"
+    "release-1.8.0"
+    "https://github.com/google/googletest.git"
+    "gtest" WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+  set (gtest_root "${CMAKE_BINARY_DIR}/gtest")
+  message(STATUS "gtest locatet at: ${gtest_root}")
```
r-rojo: misspelled `locateet`

cmake/modules/Findjsoncpp.cmake
```diff
@@ -0,0 +1,36 @@
+# - Try to find libjson-c
```
r-rojo: Copy/paste error?

tests/common_test.h
```diff
@@ -0,0 +1,308 @@
+/*common_test.h */
+
```
rpSTTD: this needs to be refactored to work with rpm files.  Please only select the functions that are needed for the tests.

tests/fpgaopen_app.cpp
```diff
@@ -0,0 +1,82 @@
+#include <getopt.h>
```
rpSTTD: This file is not needed for the initial checkin.

tests/CMakeLists.txt
```diff
@@ -0,0 +1,181 @@
...
+add_dependencies(mock mock-sysfs-prepare)
+target_link_libraries(mock dl safestr)
+
+############################################################################
+## Add testing targets #####################################################
+############################################################################
+
+add_test(NAME test_enum_mock_drv
+  WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+  COMMAND gtapi -j ${CMAKE_BINARY_DIR}/tests/configuration.json --gtest_filter=*EnumCommon*enum*)
```
rpSTTD: Need to have the capability to modify the --gtest_filte.

tests/function/gtOpenClose.cpp
```diff
@@ -0,0 +1,269 @@
...
+using namespace common_test;
+using namespace std;
+
+class LibopaecOpenFCommonHW : public common_test::BaseFixture,
+                              public ::testing::Test {};
+
+class LibopaecCloseFCommonHW : public common_test::BaseFixture,
+                               public ::testing::Test {};
+
+TEST_F(LibopaecOpenFCommonHW, 03) {
```
rpSTTD: All tests that have integer only as test name need to be verified with mock driver first.

tests/CMakeLists.txt
r-rojo: Calling `make test` should result in all unit tests (that don't require hardware) being called and are used to assess that the build is not broken. Any other tests that are more in depth can always be called manually by invoking either `gtapi` itself or a wrapper script that maps the kind of test to a specific filter. Those can also be encapsulated by a build target.
If more tests need to be called to assess the health of the build, they should be added to either:
* another call to add_test
* an longer filter

It may still be a good idea to set the filter in a CMake variable though.

tests/COPYING
```diff
@@ -0,0 +1,37 @@
+If not otherwise noted in the individual file header, the OPAE test source code,
+located in the directories
+
+  * unit,
+  * function,
+  * stress,
+  * doc,
+
+is distributed under the following license:
+
```
luebbers: Aren't these covered with the original COPYING file in the top level directory?

cmake/modules/Findjson-c.cmake
```diff
@@ -27,6 +27,7 @@ find_library(libjson-c_LIBRARIES
   /usr/local/lib
   /usr/lib
   /lib
+  /usr/lib/x86_64-linux-gnu
```
luebbers: Is this related to tests?

abelardojarab:
Still over-bloated; next version will be trimmed down.
Apologies.

PR #39, opened 2017-11-21
hssi: add multi-byte byte address support for I2C

r-rojo:
Change i2c class:
* Change constructor to take in device address width
  * This is in bytes with a default value of 1
* Add i2c::send_byte_address function that sends the byte
  address over i2c and uses the byte_address_size_ member var
  Note: The byte address is in big-endian format so we start
  with the upper bytes of the byte address
* Change i2c::read and i2c::write to use i2c::send_byte_address

Change config_app:
* Add a --byte-address-size command line argument
  and default it to 1 byte
* Instantiate an i2c device with the byte address size

### DIFF
--- a/tools/hssi/i2c.cpp
+++ b/tools/hssi/i2c.cpp
@@ -46,7 +46,7 @@ i2c::i2c(przone_interface::ptr_t przone, size_t byte_addr_size)
 
 bool i2c::send_byte_address(uint32_t instance, uint32_t byte_addr)
 {
-    uint32_t ctrl = i2c_ctrl_trigger  | i2c_ctrl_transmit | i2c_ctrl_send_start;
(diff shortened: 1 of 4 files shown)


tools/hssi/i2c.cpp
```diff
@@ -38,28 +38,37 @@ namespace hssi
...
-i2c::i2c(przone_interface::ptr_t przone)
+i2c::i2c(przone_interface::ptr_t przone, size_t byte_addr_size)
 : przone_(przone)
+, byte_addr_size_(byte_addr_size)
 {
 }
 
+bool i2c::send_byte_address(uint32_t instance, uint32_t byte_addr)
+{
+    uint32_t ctrl = i2c_ctrl_trigger  | i2c_ctrl_transmit | i2c_ctrl_send_start;
```
luebbers: Redundant space.

PR #41, opened 2017-11-21
fpgainfo: refactor sysfs properties

r-rojo:
Refactor classes/functions is sysfs.py for better property enumeration and better
handling of dynamic properties. 
These changes also fixes an issue with errors not being accurately reflected from sysfs nodes.

# Refactor static properties
Add a function called `add_static_property` that adds a class property
and uses the passed in path to
* Read and parse the sysfs file when getting
* Write to the sysfs file when setting

# Refactor dynamic properties
Add function `add_dynamic_property` that is used to add properties to
objects (rather that classes) depending on whether the relative sysfs
path exists. This should be used in the __init__ method of
sysfs_resource subclasses to add properties only if the relative path
exists.

This also changes `fme_errors` and `port_errors` to use this method.

# Refactor to_dict and print_info
Add new method, `enum_props` that enumerates object/class properties
using the `inspect.getmembers` from the `inspect` Python module.
`enum_props` will also filter out any members that
* are a member of the base class, `sysfs_resource` as these are handled
  separately
* are methods
* begin with underscore character (`_`).

Finally, refactord `errors_feature` and subclasses to use underscore
character for "private" member variables so that they won't be included
in the `enum_props` method. Added a `name()` method to get the name of
the feature.

# Change formatting func. for some props
Change formatting function from `repr` to  `int` for:
* threshold1_policy

This will allow these properties to be printed as integers.

# Update fpgaerr for refactored sysfs
Update calls in fpgaerr.py into objects from sysfs.py to accomodate
changes in sysfs.py

### DIFF
--- a/tools/fpgainfo/fpgaerr.py
+++ b/tools/fpgainfo/fpgaerr.py
@@ -263,8 +263,8 @@ error_classes = {
 def set_error_parsers(err_feature, err_classes):
     for name, err_class in err_classes.iteritems():
         if hasattr(err_feature, name):
-            err_value = getattr(err_feature, name).fget(err_feature)
-            setattr(err_feature, name, err_class(err_value))
+            value = getattr(err_feature, name)
+            setattr(err_feature, name, value)
 
 
 class errors_command(fpga_command):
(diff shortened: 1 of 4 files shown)


tools/fpgainfo/fpgaerr.py
```diff
@@ -263,8 +263,8 @@ class fatal_errors(error_mask):
 def set_error_parsers(err_feature, err_classes):
     for name, err_class in err_classes.iteritems():
         if hasattr(err_feature, name):
-            err_value = getattr(err_feature, name).fget(err_feature)
-            setattr(err_feature, name, err_class(err_value))
+            value = getattr(err_feature, name)
+            setattr(err_feature, name, value)
```
luebbers: Huh? Why retrieve a value and then set it right back?

PR #43, opened 2017-11-27
Add afu_sim_setup: a wrapper script for instantiating and configuring an ASE simulation

michael-adler:
The new script is installed in OPAE bin and is aware of both the installation and source tree locations. It copies a base ASE tree to a target directory and then invokes generate_ase_environment.py. The script only supports the new flow, in which sources are specified explicitly in a file that is parsed by rtl_src_config.

In addition, generate_ase_environment.py is updated to auto-detect the simulator when no --tool is specified explicitly. If VCS is found, it is chosen.


ase/scripts/afu_sim_setup
```diff
@@ -0,0 +1,192 @@
...
+
+    # Set ASE mode
+    os.system("sed -i 's/ASE_MODE.*/ASE_MODE = {0}/' ase.cfg".format(
+        args.ase_mode))
+    # Control transaction printing
+    os.system(("sed -i 's/ENABLE_CL_VIEW.*/ENABLE_CL_VIEW = {0}/' " +
+               "ase.cfg").format(int(args.ase_verbose)))
+
+
+def main(args=None):
```
r-rojo: Is args necessary?

ase/CMakeLists.txt
```diff
@@ -35,5 +38,23 @@ set(EXTRA_ASE_FILES
...
+        afu_sim_setup)
+
+foreach(SRC ${PLATFORM_SCRIPTS})
+  ## Substitute CMake variables of the form @<var>@ in the source file
+  ## and store the file in the build tree's bin directory.
+  configure_file(
+          scripts/${SRC} ${PROJECT_BINARY_DIR}/bin/${SRC}
+          @ONLY NEWLINE_STYLE UNIX)
+  ## This forces the configure_file to be built for all targets
+  add_custom_target(${SRC} ALL DEPENDS ${PROJECT_BINARY_DIR}/bin/${SRC})
```
r-rojo: Is this necessary? Isn't `configure_file` always run as part of the cmake configure step?

ase/scripts/afu_sim_setup
```diff
@@ -0,0 +1,192 @@
...
+
+#
+# Figure out the root of the ASE source directory.  ASE is installed
+# along with OPAE SDK, so either find it in the installation tree or in the
+# source tree.
+#
+def getASESrcPath():
+    # CMake will update any variables marked by @ with the proper values.
+    opae_install_dir = '@CMAKE_INSTALL_PREFIX@'
+    project_src_dir = '@CMAKE_CURRENT_SOURCE_DIR@'
```
r-rojo: Will this script work when installed from an RPM?

michael-adler: Good question, though not unique to this change since other scripts are using the same method. I will investigate and fix all of them, if necessary, as a separate commit.

ase/CMakeLists.txt
michael-adler: No, this is special. It performs variable substitution and puts the script in bin, where it is available to the testing environment even when not installed.

ase/scripts/afu_sim_setup
michael-adler: I meant separate pull, in case that wasn't clear. I'd like to pull this one now and fix the general problem next, if it exists.

michael-adler:
There seem to be no more objections and I'll fix the rpm problem for this and other scripts in another push.

Can someone please approve? I have downstream changes waiting on this.

Thank you.

PR #48, opened 2017-11-28
Fix fpgadiag help menu

zhangzhang10:
Fix a problem where `fpgadiag -h -m <mode>` or `fpgadiag --help -m <mode>`
fails to produce detail help message for the test specified by <mode>.

### DIFF
--- a/tools/fpgadiag/fpgadiag
+++ b/tools/fpgadiag/fpgadiag
@@ -40,19 +40,22 @@ cmd_map = {'lpbk1': ['nlb0'],
            'mb1': ['mb1']}
 
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(add_help=False)
     parser.add_argument('-t', '--target',
                         default='fpga',
                         choices=['fpga', 'ase'],
                         help='choose target')
     parser.add_argument('-m', '--mode',
                         choices=cmd_list,
-                        required=True,
                         help='choose test mode')
 
     args, leftover = parser.parse_known_args()
 
-    cmdline = cmd_map[args.mode] + leftover
+    if args.mode:
+        cmdline = cmd_map[args.mode] + leftover
+    else:
+        parser.print_help()
+        exit(1)
 
     cmdline[0] = os.path.join(cwd, cmdline[0])
     cmdline = cmdline + ['-t', args.target] + leftover
--- a/tools/fpgadiag/fpgadiag
+++ b/tools/fpgadiag/fpgadiag
@@ -44,10 +44,11 @@ if __name__ == "__main__":
     parser.add_argument('-t', '--target',
                         default='fpga',
                         choices=['fpga', 'ase'],
-                        help='choose target')
+                        help='choose target. Default: fpga')
     parser.add_argument('-m', '--mode',
                         choices=cmd_list,
-                        help='choose test mode')
+                        help='choose test mode. Combine this with -h or --help \
+                        to see detail help message for each mode.')
 
     args, leftover = parser.parse_known_args()
 
--- a/tools/fpgadiag/fpgadiag
+++ b/tools/fpgadiag/fpgadiag
@@ -31,13 +31,12 @@ import subprocess
 from subprocess import CalledProcessError
 
 cwd = os.path.dirname(os.path.realpath(__file__))
-cmd_list = ['lpbk1', 'read', 'write', 'trput', 'sw', 'mb1']
+cmd_list = ['lpbk1', 'read', 'write', 'trput', 'sw']
 cmd_map = {'lpbk1': ['nlb0'],
            'read': ['nlb3', '--mode=read'],
            'write': ['nlb3', '--mode=write'],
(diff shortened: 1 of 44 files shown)


tools/fpgadiag/fpgadiag
```diff
@@ -40,19 +40,22 @@ cmd_map = {'lpbk1': ['nlb0'],
...
     parser.add_argument('-m', '--mode',
                         choices=cmd_list,
-                        required=True,
                         help='choose test mode')
 
     args, leftover = parser.parse_known_args()
 
-    cmdline = cmd_map[args.mode] + leftover
+    if args.mode:
+        cmdline = cmd_map[args.mode] + leftover
```
luebbers: Does this print help if "-m" is set?

zhangzhang10: Yes. If "-m" is given, "-h" or "--help" print detailed help info for the test. If "-m" is not given, "-h" or "--help" print a high-level message about the command line syntax, and describing what modes (tests) and targets are available.

zhangzhang10: Github cuts off my sentence. But the complete sentence can be seen in the commit message.

buie1: When I run 
<fpgadiag -m -h>
should I expect a  more in depth description of lbk1, read, write, trput, sw, and mb1?
Right now I get the same thing for the aforementioned command line and
<fpgadiag -t -h>

usage: fpgadiag [-t {fpga,ase}] [-m {lpbk1,read,write,trput,sw,mb1}]

zhangzhang10: You will get an in depth description for a mode if the mode is specified. So for example,
`fpgadiag -m lpbk1 -h` will give a description for mode `lpbk1`.

buie1: so if i  understand correctly the expected behavior for running is as follows
```
[jabuie@sj-avl-d22-mc ~]$ fpgadiag -m write -h
Usage: nlb3 [options]

    --help, -h. Show help
    --config, -c. Path to test config file Default=nlb3.json
    --target, -t. one of { fpga, ase } Default=fpga
    --mode, -m. mode { read, write, trput } Default=write
    --begin, -b. where 1 <= <value> <= 65535 Default=1
    --end, -e. where 1 <= <value> <= 65535 Default=1
    --multi-cl, -u. one of {1, 2, 4} Default=1
    --strided-access, -a. where 1 <= <value> <= 64 Default=1
```
the only thing i would suggest would be to have a message a message showing the options for the help flag. Right now looking at just fpgadiag the only optional args are `-t` and `-m`. I think users would appreciate some indication of how helpful the -h flag can be.

zhangzhang10: Thanks for the comment. I just pushed another review that incorporates the feedback.

buie1:
@zhangzhang10  Thanks for addressing these changes. I think it will be worth while to explain in the documentation why when you run 
+ @shivarao who has been delegating documentation tasks
```
[jabuie@sj-avl-d22-mc bin]$ fpgadiag -m read -h
Usage: nlb3 [options]

    --help, -h. Show help
    --config, -c. Path to test config file Default=nlb3.json
    --target, -t. one of { fpga, ase } Default=fpga
    --mode, -m. mode { read, write, trput } Default=read
    --begin, -b. where 1 <= <value> <= 65535 Default=1
    --end, -e. where 1 <= <value> <= 65535 Default=1
```
the help says to use `nlb3` vs. continuing to use `fpgadiag`

similarly, 
```
[jabuie@sj-avl-d22-mc bin]$ fpgadiag -m sw -h
Usage: nlb7 [options]

    --help, -h. Show help
    --config, -c. Path to test config file Default=nlb7.json
    --target, -t. one of { fpga, ase } Default=fpga
    --begin, -b. where 1 <= <value> <= 65535 Default=1
    --end, -e. where 1 <= <value> <= 65535 Default=1
    --cache-policy, -p. one of { wrline-I, wrline-M, wrpush-I } Default=wrline-M
```
says the usage is for `nlb7`

I'm not as familiar with the underlying details of the fpgadiag tool so not sure why this is the case. but as long as this is the intended behavior i have no issues.

+ @r-rojo  
Do you agree?

shivarao:
It looks like the help text is being printed directly from the underlying script. Should clean that up.

Should say something like "Usage: fpgdiag -m sw [options]" instead of "Usage: nlb7 [options]".

zhangzhang10:
@shivarao @buie1 
Thanks for the comments. The latest changes remove mentioning of subprogram names (e.g. nlb0) from the help message.

buie1:
These changes are coming together great. Thanks for the work @zhangzhang10 . The last thing i want to look at is the guide for 
```
[jabuie@sj-avl-d22-mc build]$ fpgadiag -m mb1 -h
/bin/sh: /usr/local/bin/mb1: No such file or directory
```
Everything else is looking great. Lets address this and get this green check mark

zhangzhang10:
Just checked. 'mb1' shouldn't be part of fpgadiag tests. I removed it from the script. Thanks for catching this!

PR #52, opened 2017-11-29
c++: updates to c++ API and add unit tests

r-rojo:
* add libopae++ sub-directory to build (optional)
* add second constructor, `properties(fpga_objtype)` to `properties` class
* add unit tests
  * while the feature/c++ branch does not have the testing infrastructure (and libmock),
    I added a `tests` sub-directory that I have used with a version of the mock driver I had on my dev.
    machine. The plan is to eventually migrate these to the testing infrastructure once that is enabled


libopae++/include/opaec++/properties.h
```diff
@@ -45,6 +45,7 @@ class properties
     typedef std::shared_ptr<properties> ptr_t;
 
     properties();
+    properties(fpga_objtype objetype);
```
tswhison: objtype

PR #53, opened 2017-11-29
add mmio interface and mmio_region class

tswhison:


tswhison:
The first commit re-writes line endings for a couple of files that were saved in DOS format.

libopae++/include/opaec++/mmio_region.h
```diff
@@ -0,0 +1,69 @@
...
+{
+namespace io
+{
+
+class mmio_region : public mmio
+{
+public:
+    enum class id_t : int32_t
+    {
+        UAFU,
```
r-rojo: What's the `U` for?

libopae++/include/opaec++/mmio_region.h
```diff
@@ -0,0 +1,69 @@
...
+        STP
+    };
+
+    mmio_region(opae::fpga::types::handle::ptr_t h, id_t id);
```
r-rojo: Do we want a static factory function that returns a shared ptr? This way we can return a null pointer if mapping mmio is unsuccessful. 
If not, then we'll have to implement copy c'tor and assignment operator.

libopae++/include/opaec++/mmio_region.h
```diff
@@ -0,0 +1,69 @@
...
+
+    virtual bool write_mmio32(uint32_t offset, uint32_t value) override;
+
+    virtual bool write_mmio64(uint32_t offset, uint64_t value) override;
+
+    virtual bool read_mmio32(uint32_t offset, uint32_t & value) const override;
+
+    virtual bool read_mmio64(uint32_t offset, uint64_t & value) const override;
+
+    virtual uint8_t * mmio_pointer(uint32_t offset) const override { return mmio_base_; }
```
r-rojo: Should this be volatile?

libopae++/include/opaec++/mmio.h
```diff
@@ -0,0 +1,56 @@
...
+
+    virtual bool write_mmio32(uint32_t offset, uint32_t value) = 0;
+
+    virtual bool write_mmio64(uint32_t offset, uint64_t value) = 0;
+
+    virtual bool read_mmio32(uint32_t offset, uint32_t & value) const = 0;
+
+    virtual bool read_mmio64(uint32_t offset, uint64_t & value) const = 0;
+
+    virtual uint8_t * mmio_pointer(uint32_t offset) const = 0;
```
r-rojo: Should this be volatile?

tswhison: Interesting point.  That would probably save a cast to volatile on the receiving end. Will make the change.

libopae++/include/opaec++/mmio_region.h
tswhison: It just matches the nomenclature used in the shared header file "intel-fpga.h".  I can remove the U if you feel strongly about it.

libopae++/include/opaec++/mmio.h
michael-adler: Won't the underlying pointer be volatile? Isn't that enough?

libopae++/include/opaec++/mmio_region.h
tswhison: Agree.  The question is whether to provide the factory here or in the handle class, since the handle really owns the regions.  What do you think?

libopae++/include/opaec++/mmio_region.h
r-rojo: No, just curious

libopae++/include/opaec++/mmio_region.h
tswhison: I think so.

libopae++/include/opaec++/mmio.h
tswhison: I think that client code might typically do this:

volatile struct my_afu_mmio *p = (volatile struct my_afu_mmio *) region.mmio_pointer()

or else use the base+offset pattern directly. In the latter case, I think it would save a typecast.

libopae++/include/opaec++/mmio_region.h
```diff
@@ -0,0 +1,74 @@
...
+class mmio_region : public mmio
+{
+public:
+    enum class id_t : int32_t
+    {
+        AFU,
+        STP
+    };
+
+    static mmio::ptr_t region(opae::fpga::types::handle::ptr_t h, id_t id);
```
r-rojo: What do you think of calling this `map`? It would look like:
```c++
auto mmio = mmio_region::map(h, AFU);
```

Thoughts?

libopae++/include/opaec++/mmio_region.h
```diff
@@ -0,0 +1,74 @@
...
+
+    virtual bool write_mmio32(uint32_t offset, uint32_t value) override;
+
+    virtual bool write_mmio64(uint32_t offset, uint64_t value) override;
+
+    virtual bool read_mmio32(uint32_t offset, uint32_t & value) const override;
+
+    virtual bool read_mmio64(uint32_t offset, uint64_t & value) const override;
+
+    virtual volatile uint8_t * mmio_pointer(uint32_t offset) const override
```
r-rojo: Is `offset` not used?

libopae++/src/mmio_region.cpp
```diff
@@ -0,0 +1,98 @@
...
+namespace opae
+{
+namespace fpga
+{
+namespace io
+{
+
+mmio::ptr_t mmio_region::region(opae::fpga::types::handle::ptr_t h, mmio_region::id_t id)
+{
+    mmio_region *rptr = new mmio_region(h, id);
```
r-rojo: Should we call `fpgaMapMMIO` outside of the constructor? Here if result is FPGA_OK we would reset shared ptr
```c++
p.reset(new mmio_region(h, base));
```

This way we don't have to worry about deleting the pointer if the call to `fpgaMapMMIO` failed.

michael-adler:
How does this tie in with the discussion last week about MMIO (really CSR) access for devices without actual memory mapped I/O? I believe the result of the discussion was that the explicit read/write calls would work despite the CSRs not being mappable.

libopae++/include/opaec++/mmio_region.h
tswhison: sure

libopae++/include/opaec++/mmio_region.h
tswhison: Good catch. Thx!

libopae++/src/mmio_region.cpp
tswhison: Sure

tswhison:
@michael-adler In the case that MMIO is not supported, we would need a mechanism to swap the concrete class to an alternate implementation in the factory.  In that case, we might want the factory function in the MMIO class itself, and we might also want it to be given a flag of how to behave if MMIO is not supported (whether to fail to allocate an mmio or to give an alternate concrete).

tswhison:
An alternative implementation has been requested.

PR #56, opened 2017-11-29
Feature/gtapi base

abelardojarab:
this commit:

1. builds mock driver
2. creates fake sysfs structure under /tmp
3. build small gtest (gtapi) application
4. define a target 'make test' which runs gtapi; while using mock driver instead of real hardware fpga driver.
5. add travis script to execute steps 1-4 (previous steps)


tests/COPYING
```diff
@@ -0,0 +1,37 @@
+If not otherwise noted in the individual file header, the OPAE test source code,
```
luebbers: Please remove this - it should be covered by the top level COPYING file.

tests/jsonParser.cpp
```diff
@@ -0,0 +1,127 @@
+/*++
```
luebbers: Do we still need this file?

abelardojarab: i removed copying and added the license header to mock.c and gtmain.cpp

i think the json file is needed for specifying a list of green bitstreams; so it is not needed as of now and it can be removed but that removal requires to modify gtmain.cpp too as there is no longer an input json file.

luebbers:
@anandaravuri The idea of this first commit is not to include all the tests, but only the infrastructure and a minimal subset of the tests. We can add individual tests later one by one.

I'll revert the last commit.

tests/jsonParser.cpp
luebbers: Let's keep it for now in anticipation of tests that actually need JSON parsing.

cmake/modules/Findjson-c.cmake
```diff
@@ -27,6 +27,7 @@ find_library(libjson-c_LIBRARIES
   /usr/local/lib
   /usr/lib
   /lib
+  /usr/lib/x86_64-linux-gnu
   ${CMAKE_EXTRA_LIBS})
 
 if(libjson-c_LIBRARIES AND libjson-c_INCLUDE_DIRS)
```
r-rojo: Just out of curiosity, can this use `pkg-config` to find it and get include/link dirs?

cmake/modules/Findjsoncpp.cmake
```diff
@@ -0,0 +1,36 @@
...
+  PATHS ${JSONCPP_ROOT}/lib
+  /usr/local/lib
+  /usr/lib
+  /lib
+  /usr/lib/x86_64-linux-gnu
+  ${CMAKE_EXTRA_LIBS})
+
+if(jsoncpp_LIBRARIES AND jsoncpp_INCLUDE_DIRS)
+  set(jsoncpp_FOUND true)
+endif(jsoncpp_LIBRARIES AND jsoncpp_INCLUDE_DIRS)
```
r-rojo: Do we still need this file? I don't think we're using jsoncpp

cmake/modules/libraries_config.cmake
```diff
@@ -70,9 +70,12 @@ check_library_exists(rt clock_gettime "" HAVE_LIBRT)
 # uuid check
 find_package(UUID)
 
-#json-c check
+# json-c check
 find_package(json-c)
 
+# jsoncpp check
+find_package(jsoncpp)
```
r-rojo: I don't think we're using jsoncpp

tests/CMakeLists.txt
```diff
@@ -0,0 +1,163 @@
...
+endif()
+
+# build mock driver fake directory structure
+add_custom_target(mock-sysfs-prepare)
+add_custom_command(TARGET mock-sysfs-prepare POST_BUILD
+  COMMAND cmake -E copy ${CMAKE_CURRENT_SOURCE_DIR}/mock_sys_tmp-1socket-nlb0.tar.gz /tmp
+  COMMAND tar xzvf /tmp/mock_sys_tmp-1socket-nlb0.tar.gz -C /tmp --strip 1)
+
+# build mock driver
+add_library(mock SHARED mock.c)
```
r-rojo: mock is pretty generic... would it make sense to rename this library as `opae-mock` or `opae-mock-sysfs`?

tests/gtmain.cpp
```diff
@@ -0,0 +1,131 @@
...
+	     << endl
+	     << endl;
+}
+
+int main(int argc, char** argv) {
+	int getopt_ret;
+	int option_index;
+
+	const char* path;
+	path = "../tests/configuration.json";
```
r-rojo: I'd recommend making this a command line argument defaulted to "configuration.json" so that it will look for the config file in $PWD/configuration.json? We just need to:
* Copy/configure file to `${CMAKE_BINARY_DIR}`
* Run it from `${CMAKE_BINARY_DIR}`

We can also embed the default configuration in a JSON string in the binary.

PR #57, opened 2017-11-29
libopae-c: add NULL pointer check in event.c

jeffreyjin678:
libopae-c: add NULL pointer check in event.c

### DIFF
--- a/libopae/src/event.c
+++ b/libopae/src/event.c
@@ -583,6 +583,11 @@ out_free:
 
 fpga_result __FPGA_API__ fpgaDestroyEventHandle(fpga_event_handle *event_handle)
 {
+	//sanity check
+	if (!event_handle) {
+		return FPGA_INVALID_PARAM;
+	}
+
 	struct _fpga_event_handle *_eh = (struct _fpga_event_handle *) *event_handle;
 	fpga_result result = FPGA_OK;
 	int err = 0;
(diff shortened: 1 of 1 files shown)


libopae/src/event.c
```diff
@@ -583,6 +583,11 @@ fpga_result __FPGA_API__ fpgaCreateEventHandle(fpga_event_handle *event_handle)
 
 fpga_result __FPGA_API__ fpgaDestroyEventHandle(fpga_event_handle *event_handle)
 {
+	//sanity check
+	if (!event_handle) {
```
tswhison: This check is redundant.  It is performed by event_handle_check_and_lock().
I think the problem here is actually on line 591.  We should not be de-referencing event_handle in the assignment.  The correct assignment is:

struct _fpga_event_handle _*eh = (struct _fpga_event_handle *) event_handle;

Then, the call to event_handle_check_and_lock() on 595 will do the right thing.

@luebbers can you confirm?

libopae/src/event.c
```diff
@@ -583,6 +583,11 @@ fpga_result __FPGA_API__ fpgaCreateEventHandle(fpga_event_handle *event_handle)
...
+		return FPGA_INVALID_PARAM;
+	}
+
```
tswhison: The event handle is being passed by reference, so the dereference on 591 is actually correct.
To follow the convention that variables be declared before executable statements, please change the code to the following:

struct _fpga_event_handle *_eh;
fpga_result result = FPGA_OK;
int err = 0;

if (!event_handle) {
    return FPGA_INVALID_PARAM;
}

_eh = (struct _fpga_event_handle *) *event_handle;

PR #60, opened 2017-11-30
Rpan1/cmake install packager 1

rpSTTD:
Add packager installation support.


michael-adler:
I like the shape of your script better, but I'm pretty confident you didn't actually try to run packager. When I tried it, only afu.py is in the zipped file.

PR #61, opened 2017-11-30
opaec++: add MMIO support to handle class

tswhison:
Parameterize the open call and the handle constructor so that the
desired MMIO region can be specified.
Map the MMIO region during object creation and unmap it during
destruction.
Provide 32- and 64-bit read and write methods.
Provide MMIO base pointer query method.

### DIFF
--- a/libopae++/src/handle.cpp
+++ b/libopae++/src/handle.cpp
@@ -23,6 +23,7 @@
 // CONTRACT,  STRICT LIABILITY,  OR TORT  (INCLUDING NEGLIGENCE  OR OTHERWISE)
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
+#include <opae/mmio.h>
 #include "opaec++/handle.h"
 
 
@@ -33,8 +34,10 @@ namespace fpga
 namespace types
 {
 
-handle::handle(fpga_handle h)
+handle::handle(fpga_handle h, uint32_t mmio_region, uint8_t *mmio_base)
 : handle_(h)
+, mmio_region_(mmio_region)
+, mmio_base_(mmio_base)
 {
 }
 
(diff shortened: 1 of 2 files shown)


michael-adler:
From a software perspective, polymorphic read/write that appear to just do the right thing for 32 and 64 bit types is usually a good idea. When talking to hardware I think you are asking for trouble. The semantics of 32 vs. 64 bit accesses could be interpreted differently by the FPGA. For HW, we generally prefer that the function explicitly say what it will do with no room for interpretation by the compiler. Any HW writer looking at those functions will get nervous and wind up having to debug just to prove that the correct size was chosen.

tswhison:
Thanks @michael-adler. These changes were put in at Enno's request.  @luebbers can you comment?

libopae++/src/handle.cpp
```diff
@@ -33,41 +34,88 @@ namespace fpga
...
+    ptr_t p;
 
-handle::ptr_t handle::open(fpga_token token, int flags){
-    fpga_handle c_handle;
     auto res = fpgaOpen(token, &c_handle, flags);
-    if (res == FPGA_OK){
-        return handle::ptr_t(new handle(c_handle));
+    if (res == FPGA_OK) {
+
+        res = fpgaMapMMIO(c_handle, mmio_region, reinterpret_cast<uint64_t **>(&mmio_base));
```
r-rojo: Do we always want to map mmio? Or do it lazily, the first time an mmio operation is requested?

libopae++/src/handle.cpp
```diff
@@ -33,41 +34,88 @@ namespace fpga
...
 fpga_result handle::close(){
+    fpga_result res = FPGA_INVALID_PARAM;
+
     if (handle_ != nullptr){
-        auto res = fpgaClose(handle_);
-        if (res == FPGA_OK){
-            handle_ = nullptr;
-        }else{
+
+        res = fpgaUnmapMMIO(handle_, mmio_region_);
```
r-rojo: Should we only unmap if the base ptr is non null?

libopae++/src/handle.cpp
tswhison: I think we should always try.  FPGA_DEVICE has no MMIO, so the call will fail. (see latest commit) If someone is opening an FPGA_ACCELERATOR, they are most likely going to do MMIO.

libopae++/src/handle.cpp
tswhison: I don't think it's necessary. The handle tracks which regions have been allocated, and the call will fail if region mmio_region_ is not found.

libopae++/src/handle.cpp
```diff
@@ -33,41 +34,102 @@ namespace fpga
...
-    fpga_handle c_handle;
     auto res = fpgaOpen(token, &c_handle, flags);
-    if (res == FPGA_OK){
-        return handle::ptr_t(new handle(c_handle));
+    if (res == FPGA_OK) {
+
+        // This will fail for FPGA_DEVICE (no mmio region), so we need to continue when it does.
+        res = fpgaMapMMIO(c_handle, mmio_region, reinterpret_cast<uint64_t **>(&mmio_base));
+
+        if (res != FPGA_OK) {
```
r-rojo: Do we want to use the c++ API to get properties and check the type?
Also, wouldn't we want to call 'fpgaMapMMIO` only if type is `FPGA_ACCELERATOR`?

PR #64, opened 2017-12-02
c++: Add guid_t struct, implicit conversion ops

r-rojo:
Add a guid_t struct to wrap the guid type.
This uses std::array to store the guid data put also keeps a reference
to a props_ object. It uses the props_ variable to call C API functions
to get/set the GUID property.
It also copies the data when getting and setting so that it doesn't get
stale.

Add unit tests for properties and integrate those in CMake so that one
may use 'make test'


libopae++/include/opaec++/pvalue.h
```diff
@@ -36,6 +38,57 @@ namespace fpga
 namespace types
 {
 
+struct guid_t
+{
+    guid_t(fpga_properties * p)
```
tswhison: Why fpga_properties * ?  fpga_properties is the opaque type which is already a pointer.

libopae++/include/opaec++/pvalue.h
```diff
@@ -36,6 +38,57 @@ namespace fpga
...
+
+    friend std::ostream & operator<<(std::ostream & ostr, const guid_t & g){
+        char guid_str[84];
+        uuid_unparse(g.data_.data(), guid_str);
+        ostr << guid_str;
+        return ostr;
+    }
+
+private:
+    fpga_properties *props_;
```
tswhison: Same comment as above.

libopae++/include/opaec++/pvalue.h
```diff
@@ -55,44 +108,46 @@ struct pvalue
     {
     }
 
-    template<typename U>
-    pvalue<T>& operator=(U v){
-        T value = static_cast<T>(v);
-        auto res = set_(*props_, value);
-        if (res != FPGA_OK){
-            // TODO: Print error: std::cerr << "Could not set property";
+    template<typename V = typename std::enable_if<!std::is_same<T, guid_t>::value, T>::type>
```
tswhison: Help me understand what this is doing.

libopae++/include/opaec++/pvalue.h
```diff
@@ -55,44 +108,46 @@ struct pvalue
...
-        }else{
-            ostr << "null";
-        }
+        T value = p;
+        ostr << +(value);
         return ostr;
     }
 
 private:
     fpga_properties *props_;
```
tswhison: Why fpga_properties * ?

libopae++/include/opaec++/pvalue.h
r-rojo: These property wrapper structs are constructed as members of the `properties` class. They are all initialized with a pointer to the private member `props_` of the `properties` instance which is invalid until the ctor of `properties` which uses `fpgaGetProperties` to allocate the `props_`. That is why all of the property structs need to have a reference (pointer) to `props_`.

libopae++/include/opaec++/pvalue.h
r-rojo: Basically, it is enabling this templated function only when `T` is not guid_t. It also assigns the type, `T`, to `V`.
It was something I was experimenting with before when I had `guid_t` a typedef of `fpga_guid`. It's probably not needed anymore and I will most likely remove it an upcoming PR.
`::value` refers to the value of the enable_if (true or false). `::type` refers to the type argument (in this case `T`) of enable_if.

libopae++/include/opaec++/pvalue.h
tswhison: Ok, that makes sense.

PR #65, opened 2017-12-02
Fix method for finding the installed share/opae directory.

michael-adler:
The true installation path is unknown when users install from an RPM.
All we can count on is the relative positions of the installed
bin and share directories.

Also fixed an error in finding all the platforms and interfaces for the
--help message.

These are both needed for release/0.13.

### DIFF
--- a/platforms/scripts/afu_platform_config
+++ b/platforms/scripts/afu_platform_config
@@ -60,7 +60,6 @@ def errorExit(msg):
 #
 def getDBRootPath():
     # CMake will update any variables marked by @ with the proper values.
-    opae_install_dir = '@CMAKE_INSTALL_PREFIX@'
     project_src_dir = '@CMAKE_CURRENT_SOURCE_DIR@'
     db_root_dir = '@PLATFORM_SHARE_DIR@'
 
@@ -77,8 +76,11 @@ def getDBRootPath():
             # database.
             db_root = project_src_dir
         else:
-            # The script is installed.
-            db_root = os.path.join(opae_install_dir, db_root_dir)
+            # The script is installed.  The installation path isn't known
+            # since it can be changed when using rpm --prefix.  We do
+            # guarantee that the OPAE bin and share directories have the
+            # same parent.
+            db_root = os.path.join(parent_dir, db_root_dir)
     else:
         # Running out of the source tree
         db_root = parent_dir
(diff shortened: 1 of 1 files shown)


michael-adler:
Can someone please approve this so it makes its way to release/0.13 today?

platforms/scripts/afu_platform_config
```diff
@@ -782,17 +784,17 @@ def findPlatforms(db_path):
...
     # Walk all the directories
     for db_dir in db_path:
-        try:
-            # Look for JSON files in each directory
-            for fn in os.listdir(db_dir):
-                if fn.endswith('.json'):
+        # Look for JSON files in each directory
+        for fn in os.listdir(db_dir):
+            if fn.endswith('.json'):
+                try:
```
r-rojo: Would it be better to use glob?
Something like:
```python
for db_dir in  db_path:
    for json_file in glob.glob(os.path.join(db_dir, "*.json")):
...
```

platforms/scripts/afu_platform_config
```diff
@@ -804,19 +806,19 @@ def findAfuIfcs(db_path):
...
```
r-rojo: Same thing here, use glob?
Also, would it make sense to refactor the process of finding json files in a directory tree?

platforms/scripts/afu_platform_config
```diff
@@ -804,19 +806,19 @@ def findAfuIfcs(db_path):
...
                         # If it has a module-arguments entry assume the file is
                         # valid
                         if ('module-arguments' in db):
                             afus.add(fn[:-5])
-        except Exception:
-            # Give up on this file or directory if there is any error
-            None
+                except Exception:
+                    # Give up on this file or directory if there is any error
+                    None
```
r-rojo: Do we want to give up on a directory if there is an error in just one JSON file? Wouldn't it be better to give up on just that file? Perhaps a little more explicit exception handling could help?

platforms/scripts/afu_platform_config
```diff
@@ -782,17 +784,17 @@ def findPlatforms(db_path):
...
                     with open(os.path.join(db_dir, fn)) as f:
                         # Does it have a platform name field?
                         db = json.load(f)
                         platforms.add(db['platform-name'])
-        except Exception:
-            # Give up on this file or directory if there is any error
-            None
+                except Exception:
+                    # Give up on this file or directory if there is any error
+                    None
```
r-rojo: Do we want to give up on a directory if there is an error in just one JSON file? Wouldn't it be better to give up on just that file? Perhaps a little more explicit exception handling could help?

michael-adler:
@r-rojo - all of these are fine points but unrelated to the commit and the immediate problem that release/0.13 is freezing and the commit is fixing a bug already present when an rpm's --prefix is moved. Unless you have a problem with the specific change, please approve. I'll review the other comments later today or tomorrow.

r-rojo:
Sounds good.

platforms/scripts/afu_platform_config
michael-adler: It was the old code that gave up on the directory when one error was found. That's fixed in the commit, though I failed to change the comment. I will.

PR #66, opened 2017-12-03
The true installation path is unknown when users install from an RPM.

michael-adler:
All we can count on is the relative positions of the installed
bin and share directories.

### DIFF
--- a/ase/scripts/create_ase_simbuild_env.sh
+++ b/ase/scripts/create_ase_simbuild_env.sh
@@ -27,6 +27,12 @@
 
 set -e
 
+##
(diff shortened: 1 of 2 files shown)


ase/scripts/create_ase_simbuild_env.sh
```diff
@@ -27,6 +27,12 @@
 
 set -e
 
+##
```
r-rojo: Should this script be deprecated? If so, should we print out a deprecation notice when this script is run?

michael-adler: Probably. I already asked Dipti and Rahul about it.

PR #68, opened 2017-12-04
Use a simpler method for finding JSON files in a directory

michael-adler:


platforms/scripts/afu_platform_config
```diff
@@ -785,16 +786,15 @@ def findPlatforms(db_path):
...
-                    None
+        for json_file in glob.glob(os.path.join(db_dir, "*.json")):
+            try:
+                with open(json_file, 'r') as f:
+                    # Does it have a platform name field?
+                    db = json.load(f)
+                    platforms.add(db['platform-name'])
+            except Exception:
+                # Give up on this file if there is any error
+                None
```
r-rojo: Maybe this could be refactored to one common function?

```python
def find_json_db(db_path):
    for db_dir in db_path:
        for json_file in glob.glob(os.path.join(db_dir, "*.json")):
            try:
                with open(json_file, 'r') as fd:
                    db = json.load(fd)
            except Exception:
                pass
            finally:
                yield db
```

michael-adler: I did think about it, but it would get ugly. They are different classes of JSON files and the validation tests are different.

r-rojo: Just saying, in case you want to handle exceptions more explicitly, you would only do that in one spot rather than different ones.
Handling explicit exceptions might be a good idea too...
 `IOError` vs. `KeyError`/`ValueError`

michael-adler: You're welcome to make changes if you think they are important and for the better. I'm going to move on to getting MCP builds working on DCP, which was the goal of this.

michael-adler:
@r-rojo what do you want to do about this pull? Commit this part, which is improved by your suggestion to use glob, or do you want to add more exception code? The exception code could be a separate commit later, too.

PR #70, opened 2017-12-04
opaec++: add DMA buffer support

tswhison:
Provides methods for fill, compare, read, write, poll, wait, and split.
Tracks the buffer owner (handle) so that it is not released before the
buffer.

### DIFF
--- a/libopae++/include/opaec++/dma_buffer.h
+++ b/libopae++/include/opaec++/dma_buffer.h
@@ -66,16 +66,16 @@ public:
     /** Retrieve the handle smart pointer associated with
      * this buffer.
      */
-    handle::ptr_t get_owner() const { return handle_; }
+    handle::ptr_t owner() const { return handle_; }
 
     /** Retrieve the length of the buffer in bytes.
      */
-    size_t get_size() const { return len_; }
+    size_t size() const { return len_; }
 
     /** Retrieve the address of the buffer suitable for
      * programming into the accelerator device.
      */
-    uint64_t get_iova() const { return iova_; }
+    uint64_t iova() const { return iova_; }
 
     /** Divide a buffer into a series of smaller buffers.
      *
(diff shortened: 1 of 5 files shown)


libopae++/include/opaec++/dma_buffer.h
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    /** Write c to each byte location in the buffer.
+     */
+    void fill(int c);
+
+    /** Compare this dma_buffer (the first len bytes)
+     * to that held in other, using memcmp().
+     */
+    int compare(ptr_t other, size_t len) const;
```
r-rojo: Should we overload this function with a version that doesn't include `len` parameter. It would compare:
* both sizes of both buffers
* both buffers for the entire length

libopae++/include/opaec++/dma_buffer.h
```diff
@@ -0,0 +1,248 @@
...
+
+    /** Retrieve the length of the buffer in bytes.
+     */
+    size_t get_size() const { return len_; }
+
+    /** Retrieve the address of the buffer suitable for
+     * programming into the accelerator device.
+     */
+    uint64_t get_iova() const { return iova_; }
+
```
r-rojo: would it be better to leave out `get_` in the previous three methods?

libopae++/include/opaec++/dma_buffer.h
```diff
@@ -0,0 +1,248 @@
...
+     * @param[in] offset The byte offset from the start of the buffer.
+     * @param[in] each The number of microseconds to sleep each loop iteration.
+     * @param[in] timeout The maximum time to wait in microseconds.
+     * @param[in] mask A bit mask to apply to the value read from the buffer.
+     * @param[in] value The expected value after applying the mask.
+     * @retval true If the expected value was observed.
+     * @retval false If the timeout expired before seeing the expected value.
+     */
+    template <typename T>
+    bool wait(size_t offset, microseconds_t each, microseconds_t timeout, T mask, T value) const
```
r-rojo: I'm a little hesitant in including `poll` and `wait` in this API. What do you think? Should we keep this API simpler than that?

libopae++/include/opaec++/dma_buffer.h
```diff
@@ -0,0 +1,248 @@
...
+     */
+    virtual ~dma_buffer();
+
+    /** dma_buffer factory method - allocate a dma_buffer.
+     * @param[in] handle The handle used to allocate the buffer.
+     * @param[in] len    The length in bytes of the requested buffer.
+     * @return A valid dma_buffer smart pointer on success, or an
+     * empty smart pointer on failure.
+     */
+    static dma_buffer::ptr_t allocate(handle::ptr_t handle, size_t len);
```
r-rojo: What about using a pre-allocated buffer? Should we add a method called `use_buffer(uint8_t *addr)` 
?

libopae++/include/opaec++/dma_buffer.h
tswhison: In the spirit of keeping the API simple, why don't we save that for a higher-level class.

libopae++/include/opaec++/dma_buffer.h
tswhison: Added this as dma_buffer::attach().

r-rojo: 👍

libopae++/include/opaec++/dma_buffer.h
```diff
@@ -0,0 +1,271 @@
...
+        v.reserve(sizes.size());
+
+        for (const auto &sz : sizes) {
+            ptr_t p;
+            p.reset(new dma_buffer(handle_,
+                                   sz,
+                                   virt_ + offset,
+                                   wsid_,
+                                   iova_ + offset,
+                                   shared_from_this()));
```
r-rojo: 👍

PR #71, opened 2017-12-04
opae-c: link libopae-c with pthreads

tswhison:
Not linking opae-c with pthreads placed the burden on applications to
link with the library. This change adds -lpthread to the opae-c link
line where it rightly belongs.

### DIFF
--- a/libopae/CMakeLists.txt
+++ b/libopae/CMakeLists.txt
@@ -65,7 +65,8 @@ target_link_libraries(opae-c
   m
   safestr
   ${libjson-c_LIBRARIES}
-  ${libuuid_LIBRARIES})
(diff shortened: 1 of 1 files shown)


libopae/CMakeLists.txt
```diff
@@ -65,7 +65,8 @@ target_link_libraries(opae-c
   m
   safestr
   ${libjson-c_LIBRARIES}
-  ${libuuid_LIBRARIES})
+  ${libuuid_LIBRARIES}
+  ${CMAKE_THREAD_LIBS_INIT})
```
r-rojo: do we need to add `-pthread` compile option?

PR #73, opened 2017-12-05
c++: Update API and example to use new APIs

r-rojo:
Update handle class to use implicit conversion operator for handle and
for properties::type.

Update simple_example to use handle::read and handle::write and
handle::mmio_ptr.

### DIFF
--- a/libopae++/samples/simple_example.cpp
+++ b/libopae++/samples/simple_example.cpp
@@ -46,7 +46,7 @@ int main(int argc, char* argvp[])
         p.guid = uuid;
     }
     p.bbs_id = 0;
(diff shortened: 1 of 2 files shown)


libopae++/samples/simple_example.cpp
```diff
@@ -46,25 +46,19 @@ int main(int argc, char* argvp[])
         p.guid = uuid;
     }
     p.bbs_id = 0;
-    auto tokens = token::enumerate({ p });
+    auto tokens = token::enumerate({p});
```
tswhison: Maybe change p to props for clarity?  Whenever I see a variable named p, I default to thinking of it as a pointer variable.

PR #74, opened 2017-12-05
c++: Add static variable, properties::none

r-rojo:
Add a static variable, `properties::none` that is initialized to an empty vector of `properties`

### DIFF
--- a/libopae++/include/opaec++/token.h
+++ b/libopae++/include/opaec++/token.h
@@ -43,6 +43,9 @@ class token
 {
 public:
     typedef std::shared_ptr<token> ptr_t;
+
+    static std::vector<token::ptr_t> enumerate();
+
     static std::vector<token::ptr_t> enumerate(const std::vector<properties> & props);
 
     ~token();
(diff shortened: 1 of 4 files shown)


libopae++/include/opaec++/token.h
```diff
@@ -43,6 +43,9 @@ class token
 {
 public:
     typedef std::shared_ptr<token> ptr_t;
+
+    static std::vector<token::ptr_t> enumerate();
```
tswhison: Will this add some confusion?  Would a new user be tempted to use this form because it's simpler and then be confused as to why they got so many results.

r-rojo: I don't think so... so long as it's documented correctly.
It's merely a convenience function, so that users don't have to call `token::enumerate({})`
Should we require then to use it like that? Or, I just thought of this, maybe we can create a const empty filter list in the class so it would look something like:
`token::enumerate(properties::none)`
What do you think?

tswhison: Creating properties::none would do, because the user will be forced to pass something into the fn.  Good call.

PR #75, opened 2017-12-05
ASE: Solve inconsistency in pthreads for Socket_srv_tid

diptisherlekar:
- Pthread_join for socket_srv is now called in start_sim_kill function
to make sure the thread is destroyed .


lzhan55:
DCP regression tests result is available at 
http://sw-pert.altera.com/pert/pert.php?test_run_id=2107608

deepakcu84:
Looks good. You can do a reg_exe --farm --official-only to get a clean run.

I'll approve.

PR #80, opened 2017-12-06
c++: Enhance `guid_t` and add ctor to `properties`

r-rojo:
* Add a constructor to `properties` class with only one parameter of type
`fpga_guid`.
This contructor is to allow implicit conversion of an fpga_guid object
to a properties object.
* Add `get` method to `guid_t` class that returns a pointer to the guid
data.
* Add unit test for this too and fix other unit test to have a more
explicit comparison (`memcmpof`)  guid objects.

### DIFF
--- a/libopae++/include/opaec++/pvalue.h
+++ b/libopae++/include/opaec++/pvalue.h
@@ -43,6 +43,7 @@ struct guid_t {
             *props_, reinterpret_cast<fpga_guid *>(data_.data())) == FPGA_OK) {
       return data_.data();
     }
+    return nullptr;
   }
 
   const uint8_t* get() const {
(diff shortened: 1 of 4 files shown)


libopae++/include/opaec++/pvalue.h
```diff
@@ -52,6 +52,10 @@ struct guid_t
         return nullptr;
     }
 
+    const uint8_t* get() const {
+        return data_.data();
+    }
+
```
luebbers: Just curious - what happens to the guid the returned pointer is pointing to if the properties object gets destroyed?

r-rojo: Hmm... It becomes a dangling pointer

luebbers: Uh, okay. That documented anywhere? Or is that implicit?

I just wonder if someone wouldn't get a pointer to the guid and discard the original...

r-rojo: Not documented yet... 😄

PR #81, opened 2017-12-06
tools:Fix fpgaconf input file status

anandaravuri:
changes:
 - Check for input file is valid or not


tools/fpgaconf/fpgaconf.c
```diff
@@ -389,10 +390,21 @@ int read_bitstream(char *filename, struct bitstream_info *info)
 	FILE *f;
 	long len;
 	int ret;
+	struct stat file_mode = { 0 };;
```
tswhison: May want to memset file_mode to 0 instead of using the initializer syntax.  Klocwork is very picky about that.

luebbers: And you only need one ';'. We have a limited supply. ;)

tools/fpgaconf/fpgaconf.c
```diff
@@ -389,10 +390,21 @@ int read_bitstream(char *filename, struct bitstream_info *info)
...
 
 	if (!filename || !info)
 		return -EINVAL;
 
+	if (stat(filename, &file_mode) < 0) {
+		perror(filename);
+		return -1;
+	}
+
```
luebbers: Klocwork may also call out that the attributes / existence of the file referred to by `filename` may change between the `stat()` call and a subsequent `open()` (similar, though not exactly identical to [this](https://wiki.sei.cmu.edu/confluence/display/c/POS35-C.+Avoid+race+conditions+while+checking+for+the+existence+of+a+symbolic+link)). You could possibly use `fstat()` instead after `open()`.

tools/fpgaconf/fpgaconf.c
```diff
@@ -396,22 +396,22 @@ int read_bitstream(char *filename, struct bitstream_info *info)
...
+	if (!f) {
 		perror(filename);
 		return -1;
 	}
 
-	if (S_ISREG(file_mode.st_mode) == 0) {
-		fprintf(stderr, "Invalid input GBS file\n");
+	if (fstat(fileno(f), &file_mode) != 0) {
+		perror(filename);
 		return -1;
```
luebbers: You need to `goto out_close` here.

tools/fpgaconf/fpgaconf.c
```diff
@@ -396,22 +396,22 @@ int read_bitstream(char *filename, struct bitstream_info *info)
...
 
-	info->filename = filename;
-
-	/* open file */
-	f = fopen(filename, "rb");
-	if (!f) {
-		perror(filename);
+	if (S_ISREG(file_mode.st_mode) == 0) {
+		fprintf(stderr, "Invalid input GBS file\n");
 		return -1;
```
luebbers: `goto out_close`

PR #83, opened 2017-12-06
opaec++: add class except for runtime exceptions

tswhison:
except encapsulates a location in the source code along with an
fpga_result flag. except overrides std::exception::what() to provide an
informative message about the exception.

### DIFF
--- a/libopae++/src/except.cpp
+++ b/libopae++/src/except.cpp
@@ -1,0 +1,100 @@
+// Copyright(c) 2017, Intel Corporation
+//
+// Redistribution  and  use  in source  and  binary  forms,  with  or  without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of  source code  must retain the  above copyright notice,
+//   this list of conditions and the following disclaimer.
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+// * Neither the name  of Intel Corporation  nor the names of its contributors
+//   may be used to  endorse or promote  products derived  from this  software
+//   without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,  BUT NOT LIMITED TO,  THE
+// IMPLIED WARRANTIES OF  MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED.  IN NO EVENT  SHALL THE COPYRIGHT OWNER  OR CONTRIBUTORS BE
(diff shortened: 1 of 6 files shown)


libopae++/src/except.cpp
```diff
@@ -0,0 +1,100 @@
...
+    err = strcat_s(buf_, 34, ":");
+
+    err = strcat_s(buf_, 50, loc_.fn());
+
+    err = strcat_s(buf_, 53, "():");
+
+    err = snprintf_s_i(buf_+strlen(buf_), 64, "%d:", loc_.line());
+
+    err = strcat_s(buf_, sizeof(buf_), fpgaErrStr(res_));
+
```
luebbers: Do we really need to count the `dmax` parameter to `strcat_s` upwards? Can't we always say `strcat_s(buf_, sizeof(buf_), "add_me")`?

libopae++/tests/unit/gtExcept.cpp
```diff
@@ -0,0 +1,56 @@
...
+ * When the object is constructed with an fpga_result and src_location<br>
+ * Then then the fpga_result value matches the value passed to the constructor<br>
+ * And the string returned by what() represents the fpga_result and src_location.<br>
+ */
+TEST(CxxExcept, except_03){
+    except e(FPGA_INVALID_PARAM, OPAECXX_HERE);
+
+    EXPECT_EQ(FPGA_INVALID_PARAM, e);
+
+    EXPECT_STREQ("gtExcept.cpp:TestBody():49:invalid parameter", e.what());
```
luebbers: Let's just hope nobody adds or changes any test code above this...

libopae++/src/except.cpp
```diff
@@ -0,0 +1,100 @@
...
+    while ((p > file_) &&
+           (*p != '\\') &&
+           (*p != '/'))
+        --p;
+    if (('\\' == *p) ||
+        ('/' == *p))
+        ++p;
+
+    return p;
+}
```
r-rojo: Have you thought about using std::string operations to get the filename from the path? Like do a reverse find for `\\` and then call `substr` with that location?

libopae++/src/except.cpp
tswhison: I was trying to prevent a particular field from taking more than its fair share of the buffer.  Can change if you feel strongly about it.

r-rojo: Or maybe use `std::string +` operator to formulate the string?

libopae++/tests/unit/gtExcept.cpp
tswhison: Yes, new tests below, please.

libopae++/src/except.cpp
tswhison: I was trying to process the exception without allocating any memory, in case the exception was caused by a low memory condition.

tswhison:
Hmm, but wouldn't that just test the C++ runtime's ability to throw/catch?

libopae++/src/except.cpp
r-rojo: Will placement new work?

PR #86, opened 2017-12-07
c++: apply coding style

r-rojo:
I used clang-format to modify all .h|.cpp files in the libopae++
codebase to apply google's coding style for C++.

_NOTE:_ Because this coding style uses two-space indentation instead of
the four that had been used before, this change will touch most code.

### DIFF
--- a/libopae++/include/opaec++/dma_buffer.h
+++ b/libopae++/include/opaec++/dma_buffer.h
@@ -34,156 +34,137 @@
 #include <opae/buffer.h>
 #include <opaec++/handle.h>
 
-namespace opae
-{
-namespace fpga
-{
-namespace types
-{
+namespace opae {
+namespace fpga {
+namespace types {
 
-class dma_buffer : public std::enable_shared_from_this<dma_buffer>
-{
-public:
-    typedef std::size_t size_t;
-    typedef std::shared_ptr<dma_buffer> ptr_t;
+class dma_buffer : public std::enable_shared_from_this<dma_buffer> {
+ public:
+  typedef std::size_t size_t;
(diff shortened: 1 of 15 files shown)


tswhison:
I assume the tests are passing after the change?  Could you share a link to the clang-format you used and the command parameters one would pass to get the same result?

r-rojo:
Yes, the tests pass with this change :)
I [downloaded](http://releases.llvm.org/download.html) and installed llvm+clang [binaries](http://releases.llvm.org/5.0.0/clang+llvm-5.0.0-linux-x86_64-ubuntu14.04.tar.xz) on my dev machine to use clang-format. 
You can also use [cpplint.py](https://raw.githubusercontent.com/google/styleguide/gh-pages/cpplint/cpplint.py) to check it.
It does complain about our TODOs missing author or bug id though.

tswhison:
Are we not going to favor the #ifndef guards over #pragma once for headers?

r-rojo:
clang-format didn't muck with the `#pragma once` and I'd rather use those. The only thing I configured for the clang-format config is to set `SortIncludes` to `false`.
You can see the config I used here:
https://gist.github.com/r-rojo/9b873fcd920ca627a87e770d776d865f

southerngs:
do we have the coding style(s) we are using documented anywhere?

r-rojo:
For C++, I've added the coding style to the documentation/proposal I've been working on. Currently, this documentation is part of the doc tree in feature/c++.
You can see it here:
https://github.com/OPAE/opae-sdk/blob/rrojo/c%2B%2Bdoc/doc/opae-c%2B%2B/opaecxx.md

southerngs:
@r-rojo thanks for the link

PR #87, opened 2017-12-07
Install safe string library with opae-devel

southerngs:
Safe string library is built and used with OPAE.  It is also needed for some AFUs now and probably more in the future.  This PR causes the safe string library to be installed with OPAE.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,6 +68,9 @@ install(DIRECTORY common/include/opae
   DESTINATION include
   COMPONENT libopaeheaders)
 
+install(DIRECTORY common/include/safe_string
+  DESTINATION include
+  COMPONENT safestrheaders)
 ############################################################################
 ## Add 'documentation' target ##############################################
 ############################################################################
@@ -189,6 +192,8 @@ define_pkg(devel
   docxml
   platform
   samplesrc
+  safestrlib
+  safestrheaders
   GROUP "devel"
   DISPLAY_NAME "opae-devel"
   DESCRIPTION "OPAE headers, sample source, and documentation"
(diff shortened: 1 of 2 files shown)


deepakcu84:
@diptisherlekar and @lzhan55, rebase your test branches on master after this pull request is merged and push to internal.

southerngs:
@luebbers let me know if you want to review this before it's merged.

r-rojo:
Should this be included in the "dev" package too?

southerngs:
@r-rojo what's the difference?  

Also I think I'm going to do a force push to update the branch.  I used the GitHub GUI to update the branch but that introduced a merge commit.  I had the commits logically grouped so I don't want to squash.  So what I wanted to do was rebase on master, but the GUI interface doesn't appear to do that.

southerngs:
@r-rojo I don't think we have a "dev" package.  Or if we do I don't understand how it's set up in CMake.  We've done our testing on CentOS with rpms thus far.

r-rojo:
The three main packages (excluding the meta package) created are:
* opae-<version>-<platform>-libs.rpm
  * This include the base libraries
* opae-<version>-<platform>-tools.rpm
  * This includes the tools (like fpgadiag) and their required libraries
* opae-<version>-<platform>-devel.rpm 
  * This includes the header files and documentation. Because the safestring library is a development library, it would probably make sense to put it here

southerngs:
I went ahead and merged because @deepakcu84 needs this for ASE testing.

southerngs:
@r-rojo I think it is in the devel.rpm because of this line: https://github.com/OPAE/opae-sdk/pull/87/files#diff-af3b638bc2a3e6c650974192a53c7291R195

r-rojo:
👍

PR #88, opened 2017-12-07
mux.py : Change mux.py to allow single configs

r-rojo:
Change logic in mux.py that will generate a config file compatible with
sw components nlb0, nlb3, nlb7, etc. for running standalone.
Also changes the default command line switch for specifying config
( from -m to -c ) as this is the default switch for sw components.

In particular, it now generates options dictionaries dynamically based
on the args.engines
If this list of dictionaries is just one, it will include just the
dictionary in the final configuration - leaving out fpgamux specific
meta data.

This also includes:
* A fix for validating nlb3 configurations
* A fix to call process object communicate (remove wait for timeout)
  * This was necessary for cases that generate a lot of stdout
    That was locking up the proces
  * This means processes may run forever - will put in correct timeout
    fix later


tools/fpgadiag/mux.py
```diff
@@ -259,29 +256,20 @@ def run(args):
...
-             "app": "nlb3",
-             "config": next(iterators[1])},
-            {"name": "mtnlb7",
-             "app": "mtnlb7",
-             "config": next(iterators[2])},
-            {"name": "mtnlb8",
-             "app": "mtnlb8",
-             "config": next(iterators[3])}
-        ]
+        options = [{"name": n,
```
zhangzhang10: I like this pythonic way of list comprehension!

tools/fpgadiag/mux.py
```diff
@@ -140,10 +141,17 @@ def validate(self, opts):
             return False
         if int(opts["begin"]) % int(opts["multi-cl"]) > 0:
             return False
+        if int(opts["end"]) % int(opts["multi-cl"]) > 0:
```
zhangzhang10: You don't need to check "end", because the actual end value is computed using "begin" and "multi-cl". This is what Nalini once told me. You can double check with her.

r-rojo: This is just filtering out invalid configuration combinations. I think it shouldn't hurt to keep this here.

PR #91, opened 2017-12-12
 Add git version breadcrumb

luebbers:
Adds a file ".id" that shows the current git version/hash whenever the
tree is checked out. This is useful for determining the origin commit for release artifacts.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+.id ident
+
--- a/.id
+++ b/.id
@@ -1,0 +1,1 @@
+$Id: please-fill-me-in, git$
--- a/.id
+++ b/.id
@@ -1,1 +1,1 @@
-$Id: please-fill-me-in, git$
+$Id$


southerngs:
How does this work?  I did a quick google search and saw something that said the recommended way is to use the build system: https://stackoverflow.com/questions/1792838/how-do-i-enable-ident-string-for-git-repos

Also I tested checking out this branch.  Then creating a new branch and committing a file and the `.id` file did not change.  The value in it also does not match the git hash of the HEAD.  So I'd like more information about what this does and how to use it before approving the PR.

luebbers:
The `.id` file is only updated on *checkout*, i.e. either when you check out a branch, or when you clone a new repository. If you change a file and check it in, it does not update files untouched by your checkin (i.e. the `.id` file).

You can try checking out a different branch (i.e. `master`) and then your test branch again - that should update the file.

We should also pull this into the build system at some point so that individual OPAE components can query it (like the version).

luebbers:
In other words, don't check in a change and immediately create a release artifact (RPM, tarball) from your working directory. Release artifacts should always be created from a separate repository (maybe even a freshly cloned one).

luebbers:
You're right - the $Id$ only stores the file's blob SHA, not the commit's SHA. Thus it

1. only changes when the `.id` file changes (not very useful)
2. doesn't let you track down a version of the repository, but only of the `.id` file (even less useful)

Let me try something else in a different PR.

PR #94, opened 2017-12-12
OPAE: Fix mmap mmio region

anandaravuri:
Changes:
 -Add mmap mmio region address to workspace list

### DIFF
--- a/libopae/src/mmio.c
+++ b/libopae/src/mmio.c
@@ -166,7 +166,7 @@ static fpga_result map_mmio_region(fpga_handle handle, uint32_t mmio_num)
 	wsid = wsid_gen();
 	if (!wsid_add(&_handle->mmio_root,
 		      wsid,
-		      (uint64_t) NULL,
+		      (uint64_t) addr,
 		      (uint64_t) NULL,
 		      size,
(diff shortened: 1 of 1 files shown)


libopae/src/mmio.c
```diff
@@ -166,7 +166,7 @@ static fpga_result map_mmio_region(fpga_handle handle, uint32_t mmio_num)
 	wsid = wsid_gen();
 	if (!wsid_add(&_handle->mmio_root,
 		      wsid,
-		      (uint64_t) NULL,
+		      (uint64_t) addr,
```
tswhison: Do we also need to add the iova, or is that done somewhere else?

rpSTTD:
iova is added in fpgaPrepareBuffer() API.

libopae/src/mmio.c
anandaravuri: Adding MMIO pointer to work space list , ivoa is not valid in mmio, this not buffer pointer

luebbers: For MMIO, that doesn't seem necessary (no need for user space or the accelerator to know that).

PR #99, opened 2017-12-13
Merge master into feature/c++

r-rojo:
Update feature/c++ with latest from master


r-rojo:
True, let me run the tests and then we'll take it from there...

r-rojo:
So the tests seem to pass so I'll go ahead and merge this.

PR #100, opened 2017-12-13
cmake: remove jsoncpp dependency

tswhison:
jsoncpp is no longer used and is not present on some build systems,
causing the configure step to fail. Removes dependency on jsoncpp in
favor of json-c.


tests/CMakeLists.txt
```diff
@@ -88,11 +88,10 @@ include_directories(
   ${OPAE_SDK_SOURCE}/tests)
 
 add_library(commonlib SHARED common_test.h common_test.cpp)
-target_link_libraries(commonlib opae-c ${GTEST_BOTH_LIBRARIES} ${jsoncpp_LIBRARIES})
+target_link_libraries(commonlib opae-c ${GTEST_BOTH_LIBRARIES})
```
r-rojo: Does commonlib need to link with json-c?

tswhison: Good call, I think so.

r-rojo:
We should probably remove jsoncpp from .travis.yml while we're at it 😄

tswhison:
@r-rojo Removed it from .travis.yml.

PR #101, opened 2017-12-13
API: Add functions to retrieve version information

luebbers:
Adds API calls to retrieve version and build information from the
library.


libopae/src/version.c
```diff
@@ -0,0 +1,73 @@
...
+}
+
+fpga_result __FPGA_API__ fpgaGetVersionString(char *version_str, size_t len)
+{
+	if (!version_str) {
+		FPGA_MSG("build_str is NULL");
+		return FPGA_INVALID_PARAM;
+	}
+
+	strncpy(version_str, INTEL_FPGA_API_VERSION, len);
```
tswhison: Klocwork will want strncpy_s.

libopae/src/version.c
```diff
@@ -0,0 +1,73 @@
...
+}
+
+fpga_result __FPGA_API__ fpgaGetBuildString(char *build_str, size_t len)
+{
+	if (!build_str) {
+		FPGA_MSG("build_str is NULL");
+		return FPGA_INVALID_PARAM;
+	}
+
+	strncpy(build_str, INTEL_FPGA_API_HASH, len);
```
tswhison: Klocwork will want strncpy_s.

libopae/src/version.c
luebbers: Right - will fix that. Thank you.

luebbers:
Yes, the substitution was done in PR #93.

libopae/src/version.c
```diff
@@ -0,0 +1,76 @@
...
+#include <config.h>
+#endif // HAVE_CONFIG_H
+
+#include "safe_string/safe_string.h"
+#include "common_int.h"
+#include "types_int.h"
+#include "config_int.h"
+
+#include <string.h>
+
```
r-rojo: Would it make sense to use the `fpga_version` struct we have defined in types? We could embed it here as a static variable initialized using the three macros...

```C
/**
 * Semantic version
 *
 * Data structure for expressing version identifiers following the semantic
 * versioning scheme. Used in various properties for tracking component
 * versions.
 */
typedef struct {
	uint8_t major;        /**< Major version */
	uint8_t minor;        /**< Minor version */
	uint16_t patch;       /**< Revision or patchlevel */
} fpga_version;
```

PR #103, opened 2017-12-14
Specify the __main__.py file explicitly when zipping Python executables

michael-adler:
This makes it possible to store multiple scripts that share sources in a single
directory while giving each one a unique __main__.py.


cmake/modules/packaging.cmake
```diff
@@ -77,16 +77,28 @@ function(DEFINE_PKG name)
...
     set(PACKAGER_BIN ${PROJECT_BINARY_DIR}/bin/${EXE_NAME})
+
+    # Generate a __main__.py that loads the target module
+    set(BUILD_DIR_MAIN "${CMAKE_CURRENT_BINARY_DIR}/${EXE_NAME}_main")
+    file(MAKE_DIRECTORY "${BUILD_DIR_MAIN}")
+    file(WRITE "${BUILD_DIR_MAIN}/__main__.py"
+        "import sys\n"
+        "from ${MAIN_MODULE} import main\n"
+        "if __name__ == '__main__':\n"
+        "    sys.exit(main())\n")
```
r-rojo: I'm guessing we probably don't need any sort of license text since this is going to be zipped up...

PR #105, opened 2017-12-14
add "ctor/dtor" functions in log.c

r-rojo:
Add a static function, `init_log`, with the `constructor`
attribute. This will cause the function to be called when the library is
loaded for the first time, thereby initializing the log level and the
log file if specified in the environment variables.

Also add a statc function `deinit_log` that will close the log file
is it has been opened


tswhison:
Thanks for the naming change. I suppose we can't easily place them in a separate file because they are referencing the logger's static globals..  Are others ok with leaving them in log.c?

r-rojo:
You can have more than one ctor/dtor by the way (at least with gcc).
https://gcc.gnu.org/onlinedocs/gcc-4.6.4/gcc/Function-Attributes.html


> constructor
destructor
constructor (priority)
destructor (priority)
The constructor attribute causes the function to be called automatically before execution enters main (). Similarly, the destructor attribute causes the function to be called automatically after main () has completed or exit () has been called. Functions with these attributes are useful for initializing data that will be used implicitly during the execution of the program.
_**You may provide an optional integer priority to control the order in which constructor and destructor functions are run. A constructor with a smaller priority number runs before a constructor with a larger priority number; the opposite relationship holds for destructors**_. So, if you have a constructor that allocates a resource and a destructor that deallocates the same resource, both functions typically have the same priority. The priorities for constructor and destructor functions are the same as those specified for namespace-scope C++ objects (see C++ Attributes).

libopae/src/log.c
```diff
@@ -36,35 +36,18 @@
 
 /* global loglevel */
 static int g_loglevel = FPGA_LOG_UNDEFINED;
+static FILE *g_logfile;
 /* mutex to protect against garbled log output */
 pthread_mutex_t log_lock = PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP;
 
-void __FIXME_MAKE_VISIBLE__ fpga_print(int loglevel, char *fmt, ...)
+__attribute__((constructor))
+static void init_opae(void)
```
luebbers: I'd call this `init_log()`, otherwise it runs the risk of attracting a lot of other, unrelated initialization code.

libopae/src/log.c
```diff
@@ -75,6 +58,44 @@ void __FIXME_MAKE_VISIBLE__ fpga_print(int loglevel, char *fmt, ...)
...
+	if (pthread_mutexattr_settype(&mattr, PTHREAD_MUTEX_RECURSIVE) ||
+	    pthread_mutex_init(&log_lock, &mattr)) {
+		fprintf(stderr, "Failed to create log mutex\n");
+	}
+
+	pthread_mutexattr_destroy(&mattr);
+}
+
+__attribute__((destructor))
+static void deinit_opae(void)
```
luebbers: same here (`exit_log()` or `deinit_log()`)

r-rojo:
I vaguely remember seeing somewhere that clang support these attributes as well. I can't find the explicit documentation that lists this thought...

PR #109, opened 2017-12-15
Check poll() return value in hello_events 

omkarhegde:
-Add integer variable to check return code of linux
    poll() call


tswhison:
The CI failure looks like an internal error.  @r-rojo, could you have a look?

southerngs:
@omkarhegde can you remove the reference to Klocwork in the commit message since we aren't using it on the public repo.  Just describe the problem that is being fixed instead.

omkarhegde:
@southerngs  done.

tswhison:
Talked to @r-rojo. We noticed that this failure is actually in the packager, which is strange since those files are not modified by this PR. It seems that there is a missing python dependency on 'jsonschema'. @rpSTTD, is this something you can help with?

tswhison:
@omkarhegde you could try adding the following to .travis.yml, under the other pip install lines:

  - pip install --user jsonschema

PR #112, opened 2017-12-16
c++: Add C++ logger class

r-rojo:
* Add a named logger class that is used to start a logging stream
sequence with an indicated log level. This returns a `wrapped_logger`
object which is used to stream the log message and eventually emit it to
its destination by calling the log methat in libopae-c

* Add a unit test to verify the error messages are sent to stderr in the
expected format

### DIFF
--- a/libopae++/include/opaec++/log.h
+++ b/libopae++/include/opaec++/log.h
@@ -1,0 +1,215 @@
+// Copyright(c) 2017, Intel Corporation
+//
+// Redistribution  and  use  in source  and  binary  forms,  with  or  without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of  source code  must retain the  above copyright notice,
+//   this list of conditions and the following disclaimer.
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+// * Neither the name  of Intel Corporation  nor the names of its contributors
+//   may be used to  endorse or promote  products derived  from this  software
+//   without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,  BUT NOT LIMITED TO,  THE
+// IMPLIED WARRANTIES OF  MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED.  IN NO EVENT  SHALL THE COPYRIGHT OWNER  OR CONTRIBUTORS BE
+// LIABLE  FOR  ANY  DIRECT,  INDIRECT,  INCIDENTAL,  SPECIAL,  EXEMPLARY,  OR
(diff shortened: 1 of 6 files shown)


libopae++/include/opaec++/log.h
```diff
@@ -0,0 +1,215 @@
...
+   *            information
+   *
+   * @return A wrapped_stream object to begin the stream sequence
+   */
+  wrapped_stream fatal(std::string str = "");
+
+
+ private:
+  std::string name_;
+  int pid_;
```
tswhison: Should we use pid_t ?

libopae++/include/opaec++/log.h
```diff
@@ -0,0 +1,215 @@
...
+class logger {
+ public:
+  /**
+   * @brief log level enumeration that roughly maps to the log levels in
+   * libopae-c
+   */
+  enum class level : int
+  {
+      none        = -1,
+      error       = 0,
```
tswhison: What's the significance of the level values, ie why not just -1..N ?

r-rojo: This was pretty much carried over from the first log.cpp in the other c++ libraries (except the order is now reversed). When I did that one, I modeled it after the Python logging library which skips numbers in between their logging levels which I think is done to allow others to define custom log levels. 
I defined these so that you can get to the level in `libopae-c` by dividing by 20.
This is how these levels map to the levels in the `libopae-c`:

C++ enum |C++ level   | C level   | C enum                     
-------------|------------|----------|-----------------------
error           |   0             |         0    | FPGA_LOG_ERROR
exception   | 10             |         0    | FPGA_LOG_ERROR 
fatal            | 15             |         0    | FPGA_LOG_ERROR 
wan            | 20             |         1    | FPGA_LOG_MESSAGE
info             | 25             |         1    | FPGA_LOG_MESSAGE
debug         | 40             |         2    | FPGA_LOG_DEBUG

libopae++/include/opaec++/log.h
r-rojo: We could but that would require us to include `unistd.h` in this header. I try to keep the C++ API generic (without OS specific includes) if possible.

libopae++/include/opaec++/log.h
```diff
@@ -0,0 +1,215 @@
...
+  /**
+   * @brief Generic log method that uses loglevel parameter
+   *
+   * @param l The level of the log message
+   * @param str An optional string that may include more contextual
+   *            information
+   *
+   * @return A wrapped_stream object to begin the stream sequence
+   */
+  wrapped_stream log(level l, std::string str = "");
```
tswhison: Do you ever want an empty log message? Why the default of "" ?

tests/CMakeLists.txt
```diff
@@ -170,6 +171,7 @@ add_gtfilter(CxxEnum "CxxEnum*" True)
 add_gtfilter(CxxOpenClose "CxxOpen*" True)
 add_gtfilter(CxxBuffer "CxxEnum*" True)
```
tswhison: This might be a typo..  You want it to run the Enum tests when CxxBuffer is given?

libopae++/include/opaec++/log.h
r-rojo: It's mean to provide more contextual information and will get formatted by surrounding it with brackets For example:
`[16605][ERROR][logger_name][context] This is the log message`

tests/CMakeLists.txt
r-rojo: Oh! You're right! Fix coming up...

PR #114, opened 2017-12-18
Add shims to the OPAE RTL platform between the top-level PR boundary and the AFU

michael-adler:
- JSON configurable clock domains for CCI-P and local memory.
- Platform-dependent auto-insertion of register stages to reduce timing pressure.


michael-adler:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2064142

This commit looks long, but the big part is mostly inclusion of some Qsys components for clock crossing.

This code affects ASE but not Quartus builds yet. Quartus builds will only be affected when they start using afu_platform_config and a new green_bs() that imports the generated configuration.

The most significant change here is the ability to select clocks in the AFU's JSON. It is now possible to map all DDR and CCI-P interfaces to a common clock without doing anything in the AFU other than connecting to the right clock.

platforms/afu_top_ifc_db/README_AFU
```diff
@@ -23,29 +21,44 @@ field:
 
 
 In addition to the name, some module-arguments parameters may be
-overridden in the AFU JSON database.  For example, the following makes
-local memory optional:
+overridden in the AFU JSON database.  For example, the following
+requests automatic registering of CCI-P signals on platforms that
+need more than one register stage, makes local memory optional and
```
deepakcu84: "makes local memory optional" -> does this mean that the interface isn't exposed to the AFU developer? or does it mean the interface is exposed, but using it doesn't have any effect?

platforms/afu_top_ifc_db/ccip_std_afu_avalon_mm_legacy_wires.json
```diff
@@ -7,6 +7,7 @@
          "use the SystemVerilog interface described in the parent."
       ],
    "parent": "ccip_std_afu_avalon_mm",
+   "platform-shim-module-name": null,
```
deepakcu84: Do you see a reason why customers may continue to want legacy wires for memory? If not, does it make sense to remove it (perhaps in a different review)? Since we dont expose legacy wires for CCIP, the natural question is why only support different interface types for local memory alone?

platforms/platform_db/discrete_pcie3.json
```diff
@@ -30,6 +30,7 @@
             "params":
                {
                   "vc-supported": "{ 1, 0, 1, 0 }",
+                  "num-timing-reg-stages": 2,
```
deepakcu84: Can I override platform's num-timing-reg-stages with that of AFU?

platforms/platform_db/discrete_pcie3.json
```diff
@@ -30,6 +30,7 @@
...
                   "max-bw-active-lines-c0": "{ 256, 256, 256, 256 }",
                   "max-bw-active-lines-c1": "{ 128, 128, 128, 128 }"
```
deepakcu84: what are these values? perhaps add comments?

platforms/platform_db/platform_defaults.json
```diff
@@ -59,10 +48,106 @@
...
+               "comment":
+                  [
+                     "CCI-P requires that an AFU register both the request and response",
+                     "signals before any combinational logic.  On some platforms, it",
+                     "is easier to meet timing when there is more than one register",
+                     "stage.  num-timing-reg-stages is a platform-specific recommendation",
+                     "to AFUs, suggesting the number of registered CCI-P request and",
+                     "response stages for timing across the partial reconfiguration boundary.",
+                     "",
+                     "Note: There is no extra platform-side buffering added for handling",
```
deepakcu84: I dont understand the explanation of "almost full" here. Can you clarify?

platforms/platform_db/platform_defaults.json
```diff
@@ -59,10 +48,106 @@
...
+                     "2 * (num-timing-reg-stages - 1) slots are lost to buffering.",
+                     "This accounts for the added latency of receiving almost full and",
+                     "also accounts for extra in-flight requests beyond the CCI-P",
+                     "allotted single stage."
+                  ],
+               "num-timing-reg-stages": 1,
+
+               "comment":
+                  [
+                     "This field requests that the platform register CCI-P requests",
```
deepakcu84: I don't understand the difference between num-timing-reg-stages and num-auto-timing-reg-stages. Why do you set it to null here?

platforms/platform_db/platform_defaults.json
```diff
@@ -59,10 +48,106 @@
...
+                     "responsibility for managing variations in num-timing-reg-stages",
+                     "from AFUs to the platform."
+                  ],
+               "num-auto-timing-reg-stages": null,
+
+               "comment":
+                  [
+                     "AFUs may request that the platform transform the CCI-P clock",
+                     "to a different clock domain by setting this value to the",
+                     "name of another incoming clock."
```
deepakcu84: an example in DCP would be DDR clock? Do they use the clock net name at the PR boundary? Perhaps clarify what name they need to use for the clock?

platforms/platform_db/platform_defaults.json
```diff
@@ -59,10 +48,106 @@
...
+                     "that the platform add pipeline stages to the local memory",
+                     "signals before passing them to the AFU.",
+                     "",
+                     "In addition to integer counts, the string 'auto' (use double",
+                     "quotes) will request num-timing-reg-stages auto-inserted",
+                     "stages.  Note the subtle difference from the CCI-P version.",
+                     "In CCI-P, one is subtracted in 'auto' mode because the AFU",
+                     "is expected to add a CCI-P register stage.  For local memory",
+                     "the AFU is not required to add a register stage, so the full",
+                     "count of num-timing-reg-stages is added by the platform."
```
deepakcu84: You appear to suggest that num-auto-timing-reg-stages specifies the number of pipeline stages added on the BBS side of the PR boundary. I dont understand how you can change BBS pipeline stages during AFU compile. Even if the AFU developer changes this value, it wont have any effect on the GBS. We dont let customers do BBS compiles.

platforms/platform_if/rtl/device_if/ccip_if_clock.sv
```diff
@@ -0,0 +1,60 @@
...
+    input  logic pck_cp2af_softReset,
+
+    output logic clk,
+    output logic reset
+    );
+
+`ifdef PLATFORM_PARAM_CCI_P_CLOCK_IS_DEFAULT
+    assign clk = pClk;
+`else
+    assign clk = `PLATFORM_PARAM_CCI_P_CLOCK;
```
deepakcu84: Why would you want to use PLATFORM_PARAM_CCI_P_CLOCK? Would you use this if you wanted pClkDiv2 or pClkDiv4?

platforms/platform_if/rtl/platform_shims/README
```diff
@@ -0,0 +1,15 @@
+Shims here are invoked by the top-level code that manages the partial reconfiguration
+boundary.  Instead of directly instantiating an AFU (e.g. ccip_std_afu), these
+intermediate shims may be mapped first.  The shims transform the interface and then
```
deepakcu84: "may be" or "will be"?

platforms/platform_if/rtl/platform_shims/README
```diff
@@ -0,0 +1,15 @@
...
+connect to the AFU.
+
+Shims names are defined in the JSON field "platform-shim-module-name" in
+AFU top-level interface descriptions: ../../../afu_top_ifc_db/
+
+Shims typically offer automatic clock crossing and automatic platform-specific
+register stage insertion to aid in timing closure.  See the various instances of
+"clock" and "num-auto-timing-reg-stages" in ../../../platform_db/platform_defaults.json.
+
+Some modules here are sub-shims, instantiated by larger shims.  For example,
```
deepakcu84: I dont expect anyone to work directly with sub-shims. Just a comment.

platforms/platform_if/rtl/platform_shims/utils/BBB_ccip_async/platform_utils_ccip_afifo_channel.sv
```diff
@@ -0,0 +1,112 @@
+// (C) 2001-2016 Altera Corporation. All rights reserved.
+// Your use of Altera Corporation's design tools, logic functions and other
```
deepakcu84: Why not pull these files from QII install? How do you keep these files upto date for QII 18.0 and beyond? Be aware of migration issues everytime you copy a standard IP file.

ase/rtl/ase_top.sv
```diff
@@ -172,10 +172,21 @@ module ase_top();
...
+   // Usual case: generate a unique clock for local memory
+   initial begin
+             ddr_pll_ref_clk = 0;
+   end
    always #1875 ddr_pll_ref_clk = ~ddr_pll_ref_clk; // 266.666.. Mhz
+`else
+   // For testing: tie local memory's clock to some other clock.  This will likely
+   // work only with simple memory model, since it uses the incoming reference
+   // clock without modification.
+   assign ddr_pll_ref_clk = `ASE_LOCAL_MEM_CLOCK;
```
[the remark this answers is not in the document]
michael-adler: This is only a convenience for debugging. It allows me to force the frequency of memory to match CCI-P without having to introduce a clock crossing. It's completely artificial.

platforms/afu_top_ifc_db/README
```diff
@@ -24,6 +24,19 @@ Each JSON database is a dictionary, supporting the following keys:
     The name of the AFU's top-level module.  The platform will instantiate a module
     with this name.
 
+  platform-shim-module-name:
+    String [option]
+
+    The name of a shim that will be instantiated automatically between the platform
```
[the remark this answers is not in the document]
michael-adler: I agree. The comments are still necessary for developers, no?

platforms/afu_top_ifc_db/README_AFU
michael-adler: When the AFU says memory is optional, the AFU must be prepared to deal with the local_mem interface maybe being there and maybe not. There are preprocessor macros generated that will tell it what happened.

platforms/afu_top_ifc_db/ccip_std_afu_avalon_mm_legacy_wires.json
michael-adler: We need them for now for compatibility with your existing AFUs. I agree we may choose to remove them over time.

platforms/platform_db/discrete_pcie3.json
michael-adler: No, the idea is to indicate what the platform believes to be needed. You can override the number auto-instantiated by setting num-auto-timing-reg-stages.

platforms/platform_db/discrete_pcie3.json
michael-adler: These are documented in platform_defaults.json in the same directory.

platforms/platform_db/platform_defaults.json
michael-adler: Buffering delays the AFU receiving almost full. During that delay it may be sending requests, which count against the almost full limit.

platforms/platform_db/platform_defaults.json
michael-adler: num-timing-reg-stages is a platform characteristic. num-auto-timing-reg-stages is a request to actually insert stages. I'll try to clarify on a later commit. There will be some. :-)

platforms/platform_db/platform_defaults.json
michael-adler: There isn't currently a name for the DDR clock. I thought about that, but it would be difficult. For now, you can pick any of the standard CCI-P clocks.

The other problem with using a DDR clock is there is no guarantee that all the DDR clocks from different banks are the same!

platforms/platform_db/platform_defaults.json
michael-adler: It's between ase_top (or green_bs) and the AFU, not in the BBS.

platforms/platform_if/rtl/device_if/ccip_if_clock.sv
michael-adler: This is the mechanism by which the JSON file picks a different clock.

platforms/platform_if/rtl/platform_shims/README
michael-adler: Will be. I'll change this.

platforms/platform_if/rtl/platform_shims/README
michael-adler: That's true.

platforms/platform_if/rtl/platform_shims/utils/BBB_ccip_async/platform_utils_ccip_afifo_channel.sv
michael-adler: Good point. This is already a problem since we have released the BBB! I'll work on this in a later commit/merge.

michael-adler:
Your top level comment was about waiting to release until it is all coordinated with the Quartus side of the release. Waiting would be a huge mess. Remember, there are multiple boards! I need a stable OPAE SDK and then to work with each board group to get this done.

The ASE version is already different! For example, it already accepts MCP syntax on DCP.

To make progress, I need to commit these intermediate changes and work with board and documentation groups.

ase/rtl/ase_top.sv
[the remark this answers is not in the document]
michael-adler: Is this too confusing. I find it convenient but will remove if you think it's important to remove it. I was imaging some test workloads that would script this and take advantage of it.

platforms/platform_db/platform_defaults.json
deepakcu84: Let me get it straight with an example.
If I set num-timing-reg-stages to 2, and num-auto-timing-reg-stages to null, no stages will be inserted
If I set num-timing-reg-stages to 2, and num-auto-timing-reg-stages to auto, 2 stages will be inserted
If I set num-timing-reg-stages to 2, and num-auto-timing-reg-stages to 4, 4 stages will be inserted

Is that what is expected?

michael-adler:
I rewrote the comments you suggested. Thank you for the feedback.
The Quartus version problem is on my list for later commits.

Anything else?

platforms/platform_db/platform_defaults.json
deepakcu84: Does this comment mean anything to the AFU developer other than that they should pay attention to almost full when  registered stages are present?

platforms/platform_db/discrete_pcie3.json
deepakcu84: Ok

platforms/platform_db/discrete_pcie3.json
deepakcu84: Ok.

platforms/afu_top_ifc_db/ccip_std_afu_avalon_mm_legacy_wires.json
deepakcu84: right.

platforms/afu_top_ifc_db/README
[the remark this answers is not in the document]
deepakcu84: The comments are useful here. My only concern is that the content is spread over multiple README files. I prefer to have everything under a single pdf/.html.

ase/rtl/ase_top.sv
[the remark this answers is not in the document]
deepakcu84: I'd remove it rather than have it there as a hack.

platforms/platform_db/platform_defaults.json
deepakcu84: Ok. So, the legal values are "pClkDiv2" and "pClkDiv4" and "pClk", default being pClk. I wonder if you want to mention that in the comment for clarity?

platforms/platform_db/platform_defaults.json
deepakcu84: ok. one more question. how is waitrequest handled if i set num-timing-reg-stages for memory to be non-zero value?

platforms/platform_if/rtl/platform_shims/README
deepakcu84: ok

deepakcu84:
I'v asked LeePing/Jake to look so they understand this change. If I don't hear any concerns by tomorrow afternoon, I'll approve it.

deepakcu84:
Also, can you verify my understanding of num-timing-reg-stages and num-auto-timing-reg-stages?

num-timing-reg-stages=2, num-auto-timing-reg-stages=null, no stages will be inserted
num-timing-reg-stages=2, num-auto-timing-reg-stages=auto, 2 stages will be inserted
num-timing-reg-stages=2, num-auto-timing-reg-stages=4, 4 stages will be inserted (override)

Is that the intent?

ase/rtl/ase_top.sv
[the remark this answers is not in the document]
michael-adler: Ok. I will.

platforms/afu_top_ifc_db/README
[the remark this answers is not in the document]
michael-adler: These are the detailed, low level comments. I agree with you there needs to be a higher-level summary. To some extend, that was what the README at https://github.com/OPAE/opae-sdk/tree/master/platforms is for, but it goes too far in the other direction.

platforms/platform_db/platform_defaults.json
michael-adler: No, that's it.

platforms/platform_db/platform_defaults.json
michael-adler: And the user clocks so you can have any frequency. Ok, I will.

platforms/platform_db/platform_defaults.json
michael-adler: It instantiates an Avalon-MM Pipeline Bridge that implements the protocol.

michael-adler:
Your questions about num-auto-timing-reg-stages are good and, I think, are highlighting something I could do better. The problem I'm having is the inconsistency between CCI-P and Avalon MM. CCI-P requires the AFU add a register stage. Avalon MM doesn't, because it can't. (Actually green_bs has a register stage for Avalon MM, but it uses a bridge that implements the protocol.)

Right now, the behavior of num-auto-timing-reg-stages is inconsistent. For Avalon-MM, it behaves exactly as you said. When num-auto-timing-reg-stages is a number, that number of stages is added. When num-auto-timing-reg-stages is "auto", num-timing-reg-stages are added.

For CCI-P, it adds num-auto-timing-reg-stages - 1 stages, assuming that the AFU is responsible for a stage.

Would it be better if the fields were suggested-extra-timing-reg-stages and auto-add-extra-timing-reg-stages? I'd make suggested-extra 0 on MCP and 1 on DCP for CCI-P. auto-add-extra could take true/false (insert exactly the suggested-extra number) or a number (insert exactly that number).

Would that be better?

deepakcu84:
It's fine. You can keep the original naming. Perhaps add my example in the comments to better illustrate how it works. Otherwise, looks good.

michael-adler:
Merge complete and tests pass:  http://sw-pert.altera.com/pert/pert.php?test_run_id=2067197

PR #115, opened 2017-12-19
libopae++: add logging and exceptions

tswhison:
Adds logging and exceptions for the current error paths.

### DIFF
--- a/libopae++/include/opaec++/pvalue.h
+++ b/libopae++/include/opaec++/pvalue.h
@@ -30,13 +30,16 @@
 #include <cstring>
 #include <uuid/uuid.h>
 #include <opae/properties.h>
+#include <opae/utils.h>
+#include "opaec++/log.h"
+#include "opaec++/except.h"
 
 namespace opae {
 namespace fpga {
 namespace types {
 
 struct guid_t {
-  guid_t(fpga_properties *p) : props_(p) {}
+  guid_t(fpga_properties *p) : props_(p), log_("guid_t") {}
 
   operator uint8_t *() {
     if (fpgaPropertiesGetGUID(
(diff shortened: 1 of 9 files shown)


libopae++/src/dma_buffer.cpp
```diff
@@ -38,15 +38,22 @@ dma_buffer::~dma_buffer() {
     fpga_result res = fpgaReleaseBuffer(handle_->get(), wsid_);
 
     if (res != FPGA_OK) {
-      // TODO log/throw error
+      log_.error() << "fpgaReleaseBuffer() failed with (" << res
+                   << ") " << fpgaErrStr(res);
```
r-rojo: doesn't except call fpgaErrStr?

libopae++/include/opaec++/pvalue.h
```diff
@@ -66,29 +72,38 @@ struct guid_t {
...
+    fpga_result res;
+    if ((res = fpgaPropertiesGetGUID(props, &guid_value)) == FPGA_OK) {
       char guid_str[84];
       uuid_unparse(g.data_.data(), guid_str);
       ostr << guid_str;
     } else {
-      // TODO: Log or throw
+      g.log_.error() << "fpgaPropertiesGetGUID() failed with (" << res
+                     << ") " << fpgaErrStr(res);
+      throw except(res, OPAECXX_HERE);
```
r-rojo: What if the `res` variable is `FPGA_NOT_FOUND`?

libopae++/include/opaec++/pvalue.h
```diff
@@ -126,16 +141,20 @@ struct pvalue {
...
     fpga_properties props = *p.props_;
-    if (p.get_(props, &value) == FPGA_OK) {
+    fpga_result res;
+    if ((res = p.get_(props, &value)) == FPGA_OK) {
       ostr << +(value);
     } else {
-      // TODO: Log or throw
+      p.log_.error() << "property getter failed with (" << res
+                     << ") " << fpgaErrStr(res);
+      throw except(res, OPAECXX_HERE);
```
r-rojo: What if the `res` variable is `FPGA_NOT_FOUND`?

libopae++/src/dma_buffer.cpp
```diff
@@ -38,15 +38,22 @@ dma_buffer::~dma_buffer() {
...
 
 dma_buffer::ptr_t dma_buffer::allocate(handle::ptr_t handle, size_t len) {
   ptr_t p;
+  opae::fpga::internal::logger log("dma_buffer::allocate()");
 
-  if (!len) return p;
+  if (!len) {
+    log.error() << "attempt to allocate buffer with len == 0";
+    throw except(OPAECXX_HERE);
+    return p;
```
r-rojo: does the compiler complain if we don't return from here?

libopae++/include/opaec++/pvalue.h
tswhison: Ah, yes. Thanks

libopae++/include/opaec++/pvalue.h
tswhison: Yes, thanks.

PR #116, opened 2017-12-19
add logger `warn_if` and `error_if` functions

r-rojo:
* Create a null_stream object that derives from wrapped_stream but does
  nothing at the end of the stream
* Change `logger` methods to `const`
* Change `wrapped_stream::sstream_` member variable to
  * protected
  * a pointer
* Check if `wrapped_stream::sstream_` member is null in the `<<` operator overaloads
* Add `warn_if` and `error_if` functions return `null_stream` instance
  when the condition argument is null

### DIFF
--- a/libopae++/src/log.cpp
+++ b/libopae++/src/log.cpp
@@ -51,8 +51,14 @@ static std::map<logger::level, std::string> s_level_map =
     {logger::level::fatal,      "FATAL"}
 };
 
+wrapped_stream::wrapped_stream()
+    : sstream_(nullptr)
+    , level_(-1)
+{
+}
+
 wrapped_stream::wrapped_stream(int level, bool new_line)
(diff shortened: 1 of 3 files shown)


libopae++/src/log.cpp
```diff
@@ -62,34 +68,53 @@ wrapped_stream::wrapped_stream(int level, bool new_line)
...
+    *sstream_ << manip;
+  }
   return *this;
 }
 
+class null_stream : public wrapped_stream {
+ public:
+  null_stream() : wrapped_stream() {}
+
+  virtual ~null_stream(){
```
tswhison: No need for this, because wrapped_stream::~wrapped_stream is virtual.

r-rojo: I suppose... 😈

PR #119, opened 2017-12-19
opaec++: make properties object copyable

tswhison:
Adds copy c'tor and copy assignment operator to class properties.
Previously, the compiler-generated copy c'tor was errantly creating
shallow copies of properties. This resulted in multiple calls to
fpgaDestroyProperties() for the same C properties pointer. Implementing
deep copy for properties avoids this issue.

### DIFF
--- a/libopae++/include/opaec++/pvalue.h
+++ b/libopae++/include/opaec++/pvalue.h
@@ -41,12 +41,16 @@ namespace types {
 struct guid_t {
   guid_t(fpga_properties *p) : props_(p), log_("guid_t"), is_set_(false) {}
 
+  fpga_result update() {
+    fpga_result res = fpgaPropertiesGetGUID(*props_,
+            reinterpret_cast<fpga_guid *>(data_.data()));
+    is_set_ = (FPGA_OK == res);
+    return res;
+  }
+
   operator uint8_t *() {
-    if (fpgaPropertiesGetGUID(
-            *props_, reinterpret_cast<fpga_guid *>(data_.data())) == FPGA_OK) {
-      is_set_ = true;
-      return data_.data();
-    }
+    if (FPGA_OK == update())
+        return data_.data();
     return nullptr;
   }
 
@@ -70,7 +74,7 @@ struct guid_t {
   }
 
   bool operator==(const fpga_guid &g) {
-    return 0 == std::memcmp(data_.data(), g, sizeof(fpga_guid));
+    return is_set() && (0 == std::memcmp(data_.data(), g, sizeof(fpga_guid)));
   }
 
   void parse(const char *str) {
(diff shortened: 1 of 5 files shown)


libopae++/include/opaec++/pvalue.h
```diff
@@ -104,9 +106,13 @@ struct guid_t {
     return ostr;
   }
 
+  bool is_set() const { return is_set_; }
+  void is_set(bool b) { is_set_ = b;    }
```
r-rojo: Shouldn't this be private and only set to `true` when `fpgaProperties(Set|Get).*` returns FPGA_OK?
Would it be better if we called this `invalidate()` that didn't take any arguments? All it would do is set `is_set_` to `false`.

libopae++/src/properties.cpp
```diff
@@ -69,6 +69,196 @@ properties::properties()
...
+  if (p.object_id.is_set()) {
+    object_id.operator uint64_t();
+  }
+  if (p.parent.is_set()) {
+    parent.operator fpga_token();
+  }
+  if (p.guid.is_set()) {
+    guid.operator uint8_t *();
+  }
+}
```
r-rojo: Would it make sense to add a function called `pvalue<T>::update` that would call the getter to update the copy?

tswhison: Yeah, in fact the type cast operators should really be const. Using them in this way is nonintuitive.

libopae++/include/opaec++/pvalue.h
tswhison: Sure, and we'd need to make properties a friend of pvalue..

r-rojo: why? properties shouldn't need to set `is_set_` to `true` - should it? all it needs to do is call `p.invalidate()` to set it to false... no?

tswhison: Yeah, but if invalidate() is to be private in pvalue, then properties will need to be a friend. Refer to the copy assignment operator.

r-rojo: why can't `invalidate` be public? I'm ok with making properties a `friend` too. It makes sense they should be friends 😸

libopae++/src/properties.cpp
r-rojo: 👍

libopae++/src/properties.cpp
```diff
@@ -69,6 +69,279 @@ properties::properties()
...
+    }
+  }
+  return *this;
+
+log_err:
+  log_.error("operator =") << "property getter failed with (" << res
+                           << ") " << fpgaErrStr(res);
+  throw except(res, OPAECXX_HERE);
+}
+
```
r-rojo: Would it make sense to include a `throw_if` in the log?
```C++
log_.throw_if(res, OPAECXX_HERE) << "property getter failed with ...";
```

tswhison: Probably worth thinking about, yes.. but in a separate pull.

PR #121, opened 2017-12-20
Fix compile warning in hello_events

r-rojo:
Compiler was warning about ignoring the return value of `read`
This assigns the return value to `bytes_read` and acts on it if zero

### DIFF
--- a/samples/hello_events.c
+++ b/samples/hello_events.c
@@ -113,6 +113,7 @@ int main(int argc, char *argv[])
 	struct pollfd pfd;
 	int timeout = 10000;
 	int poll_ret = 0;
+	ssize_t bytes_read = 0;
 
 	UNUSED_PARAM(argc);
 	UNUSED_PARAM(argv);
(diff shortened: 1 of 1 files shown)


samples/hello_events.c
```diff
@@ -174,7 +175,9 @@ int main(int argc, char *argv[])
 			 goto out_destroy_eh;
 		} else {
 			 printf("FME Interrupt occured\n");
-			 read(pfd.fd, &count, sizeof(count));
+			 bytes_read = read(pfd.fd, &count, sizeof(count));
+			 if (bytes_read == 0)
```
southerngs: We should probably handle the case where there is an error and bytes_read is negative.  `man 3 read` says:

````
RETURN VALUE
       Upon successful completion, these functions shall return a non-negative
       integer indicating the number of bytes actually  read.  Otherwise,  the
       functions shall return −1 and set errno to indicate the error.
````

r-rojo: Since this PR was merge, I'll put that in a new PR

PR #124, opened 2017-12-20
c++: Migrate dma_buffer functions to separate file

r-rojo:
This migrates functionality that is not intended to go into the core c++
library

* Move `poll` and `wait` template functions into new file `buffers.h`
* Create a `buffer_chunk` class that can be used to split buffers
* Add gtCxxBuffers to tests


tswhison:
I'm having some reservations about "buffer_chunk" as a name. Can we change it to something else?  buffer_segment, or segment, or block maybe?

r-rojo:
What about `slice` or `nugget`? 😉

tswhison:
Nugget is tempting, but how about buffer_slice?

PR #125, opened 2017-12-21
Feature/ase windows: initial version of ASE Windows support

lzhan55:
- This is the initial version of ASE Windows support
- Please follow the ase/README.txt for environment setups
- hello_fpga test passed on Linux
- hello_fpga test passed on Windows locally


michael-adler:
Good progress.

Why the .bat files? Can't you use Makefiles, just like on Linux? Especially the client one assumes a trivial compilation model for the SW that is unlikely to be generally useful. They key is linking in the right libraries, which is what we tell people on the Linux side. On the server side, what's wrong with Makefiles? More divergence == more maintenance, testing and bugs.

You had to do a lot of work on getting the right include files. Would it make sense to factor out most of the includes into a common header file and include that common file from the ASE C files? That way there are fewer points of divergence on the two operating systems.

lzhan55:
@michael-adler Thank you for your feedbacks. I will clean up the header files. Will work on makefile on Windows.

luebbers:
Hi @lzhan55 - are you going to update this or will there be another PR with the cleaned up header files / Makefiles?

lzhan55:
Hi Enno,

Aberlardo and I need to sync up on what to do. Probably we should create a new PR for header files/cmake changes, and later on we will port those changes to Windows after that.

-Tracy

luebbers:
Thanks; I'll close this PR, then. Please feel free to open a new one with your changes once they're ready.

PR #128, opened 2017-12-21
Emit more state as C/Verilog headers from AFU JSON

michael-adler:
Includes the requested user clock frequencies.

### DIFF
--- a/tools/packager/afu_json_mgr.py
+++ b/tools/packager/afu_json_mgr.py
@@ -28,6 +28,7 @@
 import os
 import sys
 import argparse
+import re
 import json
 import zipfile
 import uuid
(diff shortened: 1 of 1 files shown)


michael-adler:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2074180

The two failures in NLB are due to a different check in from Enno, breaking ASE. I'll send mail about that later.

This change is relatively uninteresting. It exposes user clock frequencies from AFU JSON. Not used by any workloads, yet.

deepakcu84:
What is your use case? As far as I know, only OpenCL cares about exposing user-clock frequency and we do have a test for it.

michael-adler:
Performance analysis -- converting cycles to time. My current code requires that you tell the SW what the user clock frequency is set to. This is an easy path to get the frequency automatically without involving the complicated (and sometimes privileged) clock calls in libopae.

PR #129, opened 2017-12-22
c++: Add event class and tests

r-rojo:
Add file `opae::fpga::types::event` class that wraps basic event
functions in OPAE C API.

Add two unit tests to test basic functionality

Add starting/stopping of `fpgad` in test script

### DIFF
--- a/libopae++/src/events.cpp
+++ b/libopae++/src/events.cpp
@@ -44,6 +44,11 @@ event::~event()
   }
 }
 
+event::operator fpga_event_handle()
+{
+  return event_handle_;
+}
+
 event::ptr_t event::register_event(handle::ptr_t h, event::type_t t, int flags)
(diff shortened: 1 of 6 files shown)


libopae++/src/events.cpp
```diff
@@ -0,0 +1,73 @@
...
+  evptr.reset(new event(h, t, eh));
+
+  return evptr;
+}
+
+event::event(handle::ptr_t h, event::type_t t, fpga_event_handle eh)
+    : handle_(h)
+    , type_(t)
+    , event_handle_(eh)
+    , log_("event"){}
```
tswhison: Need a newline between ) and {} ?

r-rojo: uh, not sure...

PR #131, opened 2017-12-26
Interpret a power setting of zero as maximum power

michael-adler:
An AFU's JSON file specifies maximum power for some platforms.  There was no cross-
platform way to indicate maximum power.  Zero used to be an error.  Change the
library to interpret zero as maximum power.


michael-adler:
@luebbers -- Enno, you're the main approver on this one. Waiting for you. The code change is simple -- it's the semantic change that's interesting. Note that DCP is already using a power value of 0 to mean don't care, which can reasonably be interpreted as max power.

luebbers:
Have you thought about what this would mean to the `coreidle` tool? Currently, that tool reads the GBS metadata to figure out how many Xeon cores to idle to free up power budget for the AFU. If the GBS power is 0, it will currently not do anything (0 usually means "discrete card / no shared power budget", so no need to idle anything). We might want to patch `coreidle` so that it interprets 0 as "max power" on integrated platforms.

michael-adler:
The update to coreidle is a good idea. Can we assign that to a power expert independent of this check in? This change is on a critical path for the platform interface manager integration.

PR #132, opened 2017-12-26
Fix an error in Tcl parsing of Quartus macros

michael-adler:


michael-adler:
This path isn't used yet in any flows other than mine, so no tests to post. It fixes an error when the first macro doesn't match the regular expression.

PR #134, opened 2018-01-03
c++: move headers to common dir

r-rojo:


r-rojo:
I don't know... good question. opaecxx would be consistent but I was thinking about how it would look when you include something from there. 
```C++
#include <opae/enum.h>
#include <opaecxx/token.h>
```
vs.
```C++
#include <opae/enum.h>
#include <opaec++/token.h>
```

Or another idea is to put them in a sub-directory of opae. Which is what @luebbers and I had originally thought of. 
```C++
#include <opae/enum.h>
#include <opae/c++/token.h>
```

What is your opinion?

tswhison:
I tend to gravitate towards opaecxx, because it would then match the guard sentinels if we were to move away from
* #pragma once
to
* #ifndef __OPAECXX_MYINCLUDE_H__

r-rojo:
What about?
```C++
#include <opae/enum.h>
#include <opae/cxx/token.h>
```

tswhison:
Yeah, I think that works, too.

PR #135, opened 2018-01-03
Improve CCI-P clock crossing shim

michael-adler:
- Better timing, tested to 500 MHz
- Simplify DC FIFO control logic and reduce pipeline stages
- Bug fixes for almost full overrun due to long pipelines
- Restructure for better separation of Quartus IP from platform logic


michael-adler:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2087197

OpenCL failures are due to test infrastructure still being broken.

PR #136, opened 2018-01-03
Add afu_synth_setup_std for configuring Quartus builds.

michael-adler:


michael-adler:
This script is not yet invoked in any build flow but will be needed soon, so populating the repositories.

platforms/scripts/afu_synth_setup_std
```diff
@@ -0,0 +1,251 @@
...
+    elif ('FPGA_RELEASE_DIR' in os.environ):
+        release_dir = os.environ['FPGA_RELEASE_DIR']
+    else:
+        # Parent directory of the running script
+        release_dir = os.path.dirname(
+            os.path.dirname(os.path.realpath(sys.argv[0])))
+
+    # Confirm that the path looks reasonable
+    if (not os.path.exists(os.path.join(release_dir,
+                                        'hw/lib/fme-ifc-id.txt'))):
```
deepakcu84: I assume this file is guaranteed to exist?

michael-adler: Yes. The packager depends on it, so it's a reasonable test that the user has pointed at a tree that makes sense.

platforms/scripts/afu_synth_setup_std
```diff
@@ -0,0 +1,251 @@
...
+
+    os.system(cmd)
+
+    # Extract JSON info for Verilog
+    cmd = 'afu_json_mgr json-info --afu-json=' + json
+    cmd += ' --verilog-hdr=../hw/afu_json_info.vh'
+    os.system(cmd)
+
+
+def main():
```
deepakcu84: Please verify my understanding of the end user flow. The user creates the AFU source specification file, points that to this top-level script along with release and platform info. This script generates the QII project and clean/run.sh scripts (used by the current flow?). What is the main value here? Ability to centrally manage AFU project settings for a particular release?

Can you clarify the term "platform"? Are variants of DCP (inspur, rushcreek) considered separate platforms?

platforms/scripts/afu_synth_setup_std
```diff
@@ -0,0 +1,251 @@
...
+                       OPAE base environment.""")
+
+    parser.add_argument('-s', '--sources', required=1,
+                        help="""AFU source specification file that will be
+                                passed to rtl_src_config.  See "rtl_src_config
+                                --help" for the file's syntax.  rtl_src_config
+                                translates the source list into either Quartus
+                                or RTL simulator syntax.""")
+    parser.add_argument('-p', '--plat', default=None,
+                        help="""FPGA platform to simulate.""")
```
deepakcu84: FPGA platform to "synthesize" ?

deepakcu84:
Looks reasonable. I'd like Chip to have a look at it.

platforms/scripts/afu_synth_setup_std
michael-adler: Good catch. Fixed.

platforms/scripts/afu_synth_setup_std
michael-adler: Platform is a physical platform for which there is a unique release. Inspur, Rush Creek, BDX Xeon+FPGA, etc.

This script is the analog of afu_sim_setup. It reads the same source specification file but, instead of constructing an ASE environment, constructs a Quartus build environment. Once this is checked in I will check in the script that transforms existing release to work with this script and the rest of the platform interface manager. The changes are pretty minor, especially for DCP, since I took Chip's structure as a starting point and rearranged the old Xeon+FPGA releases to match.

deepakcu84: Got it. Thanks.

platforms/scripts/afu_synth_setup_std
```diff
@@ -0,0 +1,286 @@
...
+        # a tarball along with a release.
+        f.write('#!/bin/sh\n')
+        f.write('SCRIPT_DIR="$(cd "$(dirname -- "$0")" 2>/dev/null && ' +
+                'pwd -P)"\n')
+        f.write('RELEASE_DIR="$(readlink -f ${SCRIPT_DIR}/' +
+                os.path.relpath(base_src, scripts_dst) +
+                ')"\n')
+        f.write('export BBS_LIB_PATH=' +
+                '${BBS_LIB_PATH:-"${RELEASE_DIR}/hw/lib"}\n')
+        f.write('${RELEASE_DIR}/bin/run.sh $*\n')
```
chipb: Any reason to use `$*` over `"$@"` here? Do we need the word splitting or globs in the individual arguments passed to the script?

platforms/scripts/afu_synth_setup_std
```diff
@@ -0,0 +1,286 @@
...
+    os.symlink(os.path.join(scripts_src, 'clean.sh'),
+               os.path.join(scripts_dst, 'clean.sh'))
+    run_sh = os.path.join(scripts_dst, 'run.sh')
+    with open(run_sh, 'w') as f:
+        # Use relative paths so generated scripts can be shipped in
+        # a tarball along with a release.
+        f.write('#!/bin/sh\n')
+        f.write('SCRIPT_DIR="$(cd "$(dirname -- "$0")" 2>/dev/null && ' +
+                'pwd -P)"\n')
+        f.write('RELEASE_DIR="$(readlink -f ${SCRIPT_DIR}/' +
```
chipb: I'd `"`-wrap the `$SCRIPT_DIR/foo` argument passed to readlink to avoid unintended word splitting.

platforms/scripts/afu_synth_setup_std
chipb: Should double-quote `${RELEASE_DIR}/bin/run.sh` here too.

platforms/scripts/afu_synth_setup_std
```diff
@@ -0,0 +1,286 @@
...
+    cmd = 'afu_platform_config --qsf --tgt platform'
+    cmd += ' --src=' + json + ' '
+
+    # Was the platform specified on the commadn line?
+    if (args.plat):
+        cmd += args.plat
+    else:
+        # Get the platform from the release
+        plat_class_file = os.path.join(getSysReleasePath(args),
+                                       'hw/lib/fme-platform-class.txt')
```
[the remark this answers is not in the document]
michael-adler: See line 103 of https://github.com/OPAE/intel-fpga-bbb/blob/master/samples/platform-db-setup/dcp_1_0_beta/install.sh#L103

The script here and platform manager are dependent on some environment updates.

michael-adler:
@chipb I made all suggested changes and made the other setup scripts consistent with the suggested --platform instead of --plat. I had copied it from whoever first added platform support to generate_ase_environment.py.

Thank you for the comments.

deepakcu84:
I don't see the approve button anymore :( 
Approved.

PR #139, opened 2018-01-04
fpgainfo: make object_id apply to port|fme only

r-rojo:
* Rename `sysfs_resource` to `sysfs_node`
* Add `sysfs_resource` class that currently only supports `object_id`
  property
* Change sysfs classes to derive from `sysfs_node` with the exception of
  `fme_info` and `port_info`

This prevents other sysfs objects from trying to read `object_id`
property


r-rojo:
I'd like to hold off on doing that. Most of the functionality related to sysfs reads/writes is here in sysfs.py and I'd like to keep it in one file until the time comes to create proper Python distributions.

PR #140, opened 2018-01-05
opaec++: update doxygen for the current classes

tswhison:


common/include/opae/cxx/handle.h
```diff
@@ -36,14 +36,26 @@ namespace opae {
 namespace fpga {
 namespace types {
 
+/** An allocated accelerator
```
r-rojo: Should we use `accelerator` or something more generic like `fpga resource`?
A `handle` can represent either logical accelerator object or a physical fpga object.
In other words, can a fpgaconf be rewritten to use this C++ API (upper layer included) and if so, what would it look like?

common/include/opae/cxx/properties.h
```diff
@@ -38,28 +38,62 @@ namespace types {
 
 class token;
 
+/** Wraps an OPAE fpga_properties object.
+ *
+ * properties are information describing an
+ * accelerator resource that is identified by
+ * its token.
```
r-rojo: Should we add something about properties also being used as a query for enumeration?

common/include/opae/cxx/properties.h
```diff
@@ -38,28 +38,62 @@ namespace types {
...
+  /** Get the underlying fpga_properties object.
+   */
   fpga_properties get() const { return props_; }
+
+  /** Get the underlying fpga_properties object.
+   */
   operator fpga_properties() const { return props_; }
 
+  /** Retrieve the properties for a given token object.
+   * @param[in] t A token identifying the accelerator resource.
```
r-rojo: Similar comment to a previous one. Should we use something more generic like `fpga resource`?

common/include/opae/cxx/token.h
```diff
@@ -37,16 +37,30 @@ namespace opae {
 namespace fpga {
 namespace types {
 
+/** Wraps the OPAE fpga_token primitive.
+ * token's are enumerated based on a set of
```
r-rojo: There's an apostrophe in tokens. What about changing this to read something like:
`tokens are created from an enumeration operation that uses properties describing an fpga resource`

common/include/opae/cxx/handle.h
tswhison: Hmm.. I was purposefully staying away from fpga so that we didn't limit ourselves.  How about "accelerator resource"?

r-rojo: Limit ourselves to just FPGA SW? The namespace already includes fpga...

common/include/opae/cxx/handle.h
```diff
@@ -36,14 +36,26 @@ namespace opae {
 namespace fpga {
 namespace types {
 
+/** An allocated accelerator resource
+ *
+ * Represents an accelerator resource that has
+ * been allocated by OPAE. The accelerator's
+ * MMIO regions are made available for reading
```
r-rojo: Hmm. MMIO is implementation for register access. If you want to keep things generic, maybe you could rephrase this part:

>Depending on the type of resource, its registers space may be read/written using a handle object.

  We may also think about this too for the documentation for the read/write functions.

common/include/opae/cxx/handle.h
```diff
@@ -80,9 +91,29 @@ class handle {
    */
   uint8_t *mmio_ptr(uint64_t offset) const { return mmio_base_ + offset; }
 
+  /** Allocate an accelerator, given a raw fpga_token
```
r-rojo: Allocate or Open? Also, should we use "accelerator resource"?

PR #141, opened 2018-01-05
Add postinst and prerm scripts to force ldconfig to run

abelardojarab:
This commit creates two extra CMAKE variables:

* BUILD_SPHINX_DOC, set to OFF by default so sphinx doesnt run automatically (even if we type make doxygen); unless we set -DBUILD_SPHINX_DOC=ON

* RUN_LDCONFIG, set to ON by default which adds extra postinst and prerm script inside both the RPM and DEB packages to force 'ldconfig' to run; this will create or remove a file under /etc/ld.so.conf.d which contains the path to the .so's


CMakeLists.txt
```diff
@@ -27,32 +27,14 @@
...
-endif(EXISTS ${GIT_EXECUTABLE})
-
 ############################################################################
 ## Add 'versioning' library ################################################
 ############################################################################
 
-set(INTEL_FPGA_API_VER_MAJOR 0  CACHE STRING "Intel FPGA API major version")
-set(INTEL_FPGA_API_VER_MINOR 13 CACHE STRING "Intel FPGA API minor version")
-set(INTEL_FPGA_API_VER_REV   0  CACHE STRING "Intel FPGA API revision version")
+set(INTEL_FPGA_API_VER_MAJOR 0 CACHE STRING "Intel FPGA API major version")
```
tswhison: Should we be changing the version in this PR?

CMakeLists.txt
```diff
@@ -107,22 +92,18 @@ if(BUILD_LIBOPAE_C)
   add_subdirectory(libopae)
 endif()
 
-add_subdirectory(platforms)
```
tswhison: Not sure we should be removing these. Can we add them back?

CMakeLists.txt
```diff
@@ -198,10 +188,7 @@ define_pkg(devel
   docrtf
   docman
   docxml
-  platform
```
tswhison: Do we want to remove these targets in this PR?

CMakeLists.txt
```diff
@@ -294,13 +301,13 @@ set(CPACK_RPM_EXCLUDE_FROM_AUTO_FILELIST
 set(CPACK_SOURCE_GENERATOR "TGZ")
 if (HASH_ARCHIVES)
   set(CPACK_SOURCE_PACKAGE_FILE_NAME
-    "${CMAKE_PROJECT_NAME}-${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}-${CPACK_PACKAGE_RELEASE}_git${GIT_COMMIT_HASH}")
+    "${CMAKE_PROJECT_NAME}-${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}_git${GIT_COMMIT_HASH}")
   set(DEFINE_RPM_NAME "%define _rpmfilename %%{ARCH}/%%{NAME}-%%{VERSION}-%%{RELEASE}_git${GIT_COMMIT_HASH}.%%{ARCH}.rpm")
-else()
+else(HASH_ARCHIVES)
   set(CPACK_SOURCE_PACKAGE_FILE_NAME
-    "${CMAKE_PROJECT_NAME}-${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}-${CPACK_PACKAGE_RELEASE}")
```
tswhison: Are you sure we want to remove this?

r-rojo:
It looks like these changes are based off an older base. You might want to either merge in master or rebase this on master.

CMakeLists.txt
abelardojarab: i forgot to rebase, those changes are not related to this fix

could you please review again please

PR #143, opened 2018-01-05
Creating dir "rtl" when creating ASE build dir

phhorrein:
When the ase build dir is generated through the provided scripts, the rtl
directory was not created. This very minor change creates the dir when calling the scripts, which prevents the python scripts used to configure the environment from complaining.

### DIFF
--- a/ase/Makefile
+++ b/ase/Makefile
@@ -44,6 +44,8 @@
 
 include ase_sources.mk
 
+ASE_OOD = 0
+
 ###############################################################
 ##                                                           ##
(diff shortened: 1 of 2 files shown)


luebbers:
@phhorrein Thank you for the contribution. As it happens, we have stumbled over the same dependency; it has already been addressed (see also #160). The updated documentation at https://github.com/OPAE/opae-sdk/blob/master/doc/src/ase_userguide/ase_userguide.md contains pointers to the new scripts that replace `create_ase_simbuild_env.sh`.

PR #148, opened 2018-01-05
 Add pre-removal and post-install script to RPM and DEB to run ldconfig

abelardojarab:
Add pre-removal and post-install script to RPM and DEB to run ldconfig


abelardojarab:
there was a ticket from Chris Rauer saying ldconfig was not running after installing the libraries and he wasnt able to compile OpenCL against OPAE libraries as OPAE libraries were appearing as missing

PR #151, opened 2018-01-09
Enable buiding ASE by default

rpSTTD:


southerngs:
@rpSTTD for some reasons I missed seeing this PR till now.  In general I recommend doing a rebase (`git rebase master` when your branch is checked out) followed by force push.  The change itself shouldn't conflict with current master and we don't really need merge commits associated with it.

@luebbers I think that `Rebase and merge` or `Squash and merge` should remove the merge commits so they don't show up in the history though.  If that's true should we approve the PR as is?

luebbers:
@southerngs Yes, we only do "Squash and merge", which will roll all commits into one, so the merge commits will not show up.

PR #152, opened 2018-01-09
Improve the CCI-P platform clock crossing code

michael-adler:
- Improve error detection by blocking on buffer overflow
- Reduce Tx FIFO sizes to reduce latency when requests back up. This doesn't affect bandwidth.


michael-adler:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2099285

PR #153, opened 2018-01-09
mb1: fix device status memory issue (#111)

tswhison:
mb1 declares class member variable dsm_. The run() function was using
a local variable to allocate the buffer, but then attempting to use dsm_
to run the actual test. This resulted in a segfault. The fix is to both
allocate and use dsm_.


tswhison:
Cherry-picking this fix from master.

PR #154, opened 2018-01-09
mmlink: fix command line options

tswhison:
Prior to this change, mmlink had the following errors in its
command-line parsing.
* The program did not respond to -h nor --help
* The names of the long options, eg --tcpport, did not match those
  appearing in the help message.
* The IP address field stored in the command params did not support a
  true string IP address. This meant that the IP address was required to
  be input as an integer value, accounting for the correct
  host/network byte ordering.
This change addresses these issues.

### DIFF
--- a/tools/mmlink/main.cpp
+++ b/tools/mmlink/main.cpp
@@ -28,10 +28,13 @@
 #include <getopt.h>
 #include <signal.h>
 #include <new>
+#include <arpa/inet.h>
 
 #include <opae/fpga.h>
 #include "mmlink_server.h"
 #include "mm_debug_link_interface.h"
+
+#include "safe_string/safe_string.h"
 
 // STP index in AFU
 #define FPGA_PORT_INDEX_STP               1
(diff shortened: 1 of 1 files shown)


omkarhegde:
Hi Tim, When running the mmlink tool as " sudo ./mmlink  -I=127.0.0.1", I see the following errrorL

/storage/shared/home_directories/omhegde/git/opae-sdk/tools/mmlink/main.cpp:374:ParseCmds() : Missing required argument for --ip-address
/storage/shared/home_directories/omhegde/git/opae-sdk/tools/mmlink/main.cpp:140:main() : Error scanning command line

tools/mmlink/main.cpp
```diff
@@ -124,27 +126,34 @@ int main( int argc, char** argv )
...
-	fpga_token port_token               = NULL;
+	fpga_token port_token              = NULL;
 	fpga_handle  port_handle           = NULL;
 	uint64_t *mmio_ptr                 = NULL;
+	int res;
 
 	// Parse command line
 	if ( argc < 2 ) {
 		MMLinkAppShowHelp();
 	return 1;
```
r-rojo: is the indentation off?

tools/mmlink/main.cpp
```diff
@@ -124,27 +126,34 @@ int main( int argc, char** argv )
...
-	} else if ( 0!= ParseCmds(&mmlinkCmdLine, argc, argv) ) {
-		PRINT_ERR( "Error scanning command line \n.");
+	} else if ( 0 != (res = ParseCmds(&mmlinkCmdLine, argc, argv)) ) {
+		if (res != -2)
+			PRINT_ERR( "Error scanning command line \n.");
 	return 2;
```
r-rojo: indentation?

tswhison:
Hi @omkarhegde, can you try -I 127.0.0.1 instead of -I=127.0.0.1?  Has -I=127.0.0.1 worked in the past?  I'm not accustomed to the form of the options.  Usually, only the long options require the = sign.

Hi @shivarao, it should't require any documentation update, so long as the docs reference the actual long options and not the ones described by the previous help output.  I will try to track down that doc to confirm.

tswhison:
@shivarao, yes the documentation is a mismatch.  Would you prefer me to change the names of the long options so that they match the doc?

omkarhegde:
@tswhison '-I <ip address>' works fine. Could you please remove the '=' from the help log?

tswhison:
@omkarhegde ok, sure, I'll remove mention of it from the help message.

PR #155, opened 2018-01-09
Updates to ASE documentation and examples

michael-adler:
- Begin to describe the flow using afu_sim_setup and rtl_src_config.
- Describe auto-discovery in generate_ase_environment.py as a legacy flow.
- Remove the NLB sample. It is complicated, doesn't use the new flow and specific to one platform.
- Add references to the BBB tutorial for an example of the new flow.
- Remove the old (broken) create_ase_simbuild.sh
- Rename ase_setup.sh to ase_setup_template.sh to make it clear that the script won't work until modified with site-specific details.


deepakcu84:
- How do we ensure that BBB tutorial examples are not broken (in simulation/hw)? Do we have regressions running against them?
- Is there any other channel to obtain NLB source?

diptisherlekar:
Arent the BBB owners responsible to check for the functionality of their AFU? Currently, the MCP ASE regressions do not support  BBB AFU's . We used to have regressions for BBB's them before we changed repos.

rahulrs:
I second @deepakcu84 
I fear that with this change, there wont be any other channel to get a sample NLB RTL. Is there a plan to make it available through some other route?

deepakcu84:
Do you need to update the README.md under tutorials - https://github.com/OPAE/intel-fpga-bbb/tree/master/samples/tutorial

Specifically, these instructions:
"Building RTL for simulation in ASE requires multiple steps.  The steps
are identical for all examples:"

Otherwise, your documentation changes look good.

michael-adler:
There aren't buttons to answer individually with the questions, so here is a concatenation. Sorry.

Chris is working on getting the BBB repository into the DCP testing environment. Once there, we can run regressions on the examples. I agree that is important. The tutorial examples are already quite public -- we steer people toward them on the HARP forums -- so we already have the burden of keeping them working.

NLB RTL has been released in every platform as an AFU example: SR-5.0.3, SR-6.3.0 and DCP 1.0. Those are run through regressions, which I believe isn't the case for the sample under ASE. I agree we should be releasing NLB but there is a path for that other than inside ASE. I'd like the ASE example to be trivial so that users aren't bogged down in details such as Qsys IP, multiple directories, etc. Hello world is one SystemVerilog file and one C file. There is also talk of changing the base self test module, so NLB may be even less relevant in the future.

@deepakcu84: You must be interpreting "the steps are identical for all examples" differently than I am. I meant for the 4 examples given. For that, the statement is currently true. How did you interpret it?

PR #156, opened 2018-01-09
mmlink: fix command line options (#154)

tswhison:
Prior to this change, mmlink had the following errors in its
command-line parsing.
* The program did not respond to -h nor --help
* The names of the long options, eg --tcpport, did not match those
  appearing in the help message.
* The IP address field stored in the command params did not support a
  true string IP address. This meant that the IP address was required to
  be input as an integer value, accounting for the correct byte
  host/network ordering.
This change addresses these issues.

### DIFF
--- a/tools/mmlink/main.cpp
+++ b/tools/mmlink/main.cpp
@@ -28,10 +28,13 @@
 #include <getopt.h>
 #include <signal.h>
 #include <new>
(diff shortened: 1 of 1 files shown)


tswhison:
Cherry-pick'ing this change back to master.

PR #159, opened 2018-01-11
change travis build to use matrix

r-rojo:
Use the `matrix` keyword to define four different job configurations to run in parallel.
These jobs run in parallel with their own dependencies.
For example, the job that only does tests the C coding style doesn't require any apt packages so they are omitted from it resulting in a faster job. The end result looks like this:
image


luebbers:
I see multiple jobs in all pull requests on travis, not just this one. How is this different?

r-rojo:
This is different because each job is configured explicitly whereas the previous job configuration used was more for stages where each job was configured the same. In other words, the old way ran `apt-get install` and `pip install` for each job. This configuration only execute those commands for the jobs that include them.

luebbers:
I see. That's not really evident in the Travis output is it?

r-rojo:
Only if you look at the run time for each job. Notice the first two faster than the others

PR #160, opened 2018-01-11
 Replace safe_string with inline protection in ASE RTL-side library

michael-adler:


michael-adler:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2104369

lzhan55:
Could you also run the ASE testing with Dipti on Jenkins? I agree with the changes since it makes the porting to Windows much easier.

michael-adler:
Having the central functions you created in ase_ops.c was instrumental in making this change easy, so it's great that you mapped all the string calls through common functions.

I thought it would help future developers to have the string functions specifically in a file of their own, especially when questioned about why the safe string library isn't being used. If everyone disagrees with the choice to add a new file for these "safe" functions I will put them back in ase_ops.c. I thought the change to a separate file made it clearer.

rahulrs:
Having a string functions file is good too.
If we want to do that, all of the string check functions in ase_ops.c need to be moved to the string functions file for cleanliness sake.

michael-adler:
Which other functions do you mean? The only possible candidate I see is ase_string_copy. I can move that. Anything else?

rahulrs:
I think this is the list:
remove_spaces
remove_tabs
remove_newline
ase_string_copy
Does anyone have anything else to add?

ase/sw/ase_ops.c
```diff
@@ -59,15 +59,14 @@ inline void set_loglevel(int level)
...
 {
-	errno_t err = EOK;
+	int res;
 
-	err = strncpy_s(name, strlen(SOCKNAME)+1, SOCKNAME,
-							strlen(SOCKNAME)+1);
-	if (err != EOK) {
-		ASE_ERR("%s: Error strncpy_s\n", __func__);
-		return err;
+	res = ase_strncpy(name, SOCKNAME, strlen(SOCKNAME)+1);
```
tswhison: If we are implementing our own version of these, could we follow the same parameter definitions as the safe_string call so that we perform identical checking?  I remember reading that baking your own string functions was frowned upon.  We'd probably raise fewer eyebrows if we could convincingly say that we ported the minimum set of safe_string fn's.

michael-adler: I'll try that. It will result in less change, which may be an improvement.

michael-adler:
Updated as suggested. ase_strings.c now has only the safe string code. @rahulrs, please dismiss the requested changes if you are satisfied.

https://sj-arc.altera.com/arc/dashboard/reports/show_job/649434113

PR #163, opened 2018-01-12
opaec++: update the documentation

tswhison:
Sync the plantuml diagrams with the current code implementation of the
first layer. Provide some explanatory text for some of the first layer
interactions. Make a few misc fixes.

### DIFF
--- a/doc/opae-c++/opaecxx.md
+++ b/doc/opae-c++/opaecxx.md
@@ -190,7 +190,7 @@ namespace opae::fpga::types {
 
 Once one or more `token`s have been enumerated, the application must choose which `token`s to request. The `token` is then converted to a `handle` by requesting that a `handle` object be allocated and opened for it.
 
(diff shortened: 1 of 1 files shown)


doc/opae-c++/opaecxx.md
```diff
@@ -125,51 +157,77 @@ namespace opae.fpga.types {
...
 }
-
 @enduml
 ```
 
+`properties` are used to narrow the search space for `token`s. Before enumerating the accelerator resources in the system, applications can produce one or more `properties` objects whose values are set to the desired characteristics for the resource. For example, an application may search for an accelerator resource based on its guid.
+
+Once one or more `token`s have been enumerated, the application must choose which `token`s to request. The `token` is then converted to a `handle` by requesting that a `handle` object be allocated and opened for it.
+
+Once a `handle` has been successfully opened, the application may use the `handle` to allocate `dma_buffer`s or to register `event`s. The `dma_buffer` and `event` objects retain a reference to their owning `handle` so that the `handle` does not lose scope before freeing the `dma_buffer` and `event` objects.
```
r-rojo: what about reading/writing registers?

PR #164, opened 2018-01-16
Update low-level platform manager documentation

michael-adler:
- Switch to Markdown
- More detail


lzhan55:
@michael-adler What is the difference between README and README.md? 
We would like to get an overview of the script flows for ASE sample from you on Thursday's meeting. Aberlado will send you emails about this soon.

michael-adler:
Markdown

PR #165, opened 2018-01-17
opaecxx: Add hello_fpga version using opaecxx-core

r-rojo:
Add hello_cxxcore which is a C++ version of hello_fpga using the opaecxx-core layer.

### DIFF
--- a/libopae++/samples/hello_fpga-1.cpp
+++ b/libopae++/samples/hello_fpga-1.cpp
@@ -34,31 +34,14 @@
 
 using namespace opae::fpga::types;
 
-#ifndef LOG2_CL
-# define LOG2_CL 6
-#endif // LOG2_CL
-
-static uint64_t cacheline_aligned_addr(uint64_t num) {
-  return num >> LOG2_CL;
-}
-
-uint64_t operator "" _CL(unsigned long long int num) {
-  return num * 64;
-}
-
-size_t operator "" _KB(unsigned long long int num) {
-  return 1024 * num;
-}
-
-size_t operator "" _MB(unsigned long long int num) {
-  return 1024 * 1024 * num;
-}
-
-static const char* NLB0_AFUID = "D8424DC4-A4A3-C413-F89E-433683F9040B";
-
-static const size_t LPBK1_DSM_SIZE                 = 4_KB;
-static const size_t LPBK1_BUFFER_SIZE              = 1_KB;
-static const size_t LPBK1_BUFFER_ALLOCATION_SIZE   = 4_KB;
+static const char* NLB0_AFUID                      = "D8424DC4-A4A3-C413-F89E-433683F9040B";
+static const uint64_t CL                           = 64;
+static const uint64_t KB                           = 1024;
+static const uint64_t MB                           = KB * 1024;
+static const uint64_t LOG2_CL                      = 6;
+static const size_t LPBK1_DSM_SIZE                 = 4*KB;
+static const size_t LPBK1_BUFFER_SIZE              = 1*KB;
+static const size_t LPBK1_BUFFER_ALLOCATION_SIZE   = 4*KB;
 static const uint64_t CSR_SRC_ADDR                 = 0x0120;
 static const uint32_t CSR_DST_ADDR                 = 0x0128;
 static const uint32_t CSR_CTL                      = 0x0138;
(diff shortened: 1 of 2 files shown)


libopae++/samples/hello_fpga-1.cpp
```diff
@@ -0,0 +1,131 @@
...
+// IMPLIED WARRANTIES OF  MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED.  IN NO EVENT  SHALL THE COPYRIGHT OWNER  OR CONTRIBUTORS BE
+// LIABLE  FOR  ANY  DIRECT,  INDIRECT,  INCIDENTAL,  SPECIAL,  EXEMPLARY,  OR
+// CONSEQUENTIAL  DAMAGES  (INCLUDING,  BUT  NOT LIMITED  TO,  PROCUREMENT  OF
+// SUBSTITUTE GOODS OR SERVICES;  LOSS OF USE,  DATA, OR PROFITS;  OR BUSINESS
+// INTERRUPTION)  HOWEVER CAUSED  AND ON ANY THEORY  OF LIABILITY,  WHETHER IN
+// CONTRACT,  STRICT LIABILITY,  OR TORT  (INCLUDING NEGLIGENCE  OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+#include "opae/cxx/properties.h"
```
tswhison: Seems like we should have a meta-header that includes each of these so that we could just do

#include <opae/cxx/opaecxx.h>

libopae++/samples/hello_fpga-1.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+  // open accelerator and map MMIO
+  auto accel = handle::open(tok, FPGA_OPEN_SHARED);
+
+  // allocate buffers
+  auto dsm = dma_buffer::allocate(accel, LPBK1_DSM_SIZE);
+  auto inp = dma_buffer::allocate(accel, LPBK1_BUFFER_ALLOCATION_SIZE);
+  auto out = dma_buffer::allocate(accel, LPBK1_BUFFER_ALLOCATION_SIZE);
+
+  // initialzie buffers
```
michael-adler: initialzie -> initialize

I don't like the dma_buffer::fill method. The data size is ambiguous. It takes an uint32_t and passes it to memset, so it's a byte value? Why should we have so many "helper" functions in the base dma_buffer class (like fill() and equal())?

libopae++/samples/hello_fpga-1.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+#ifndef LOG2_CL
+# define LOG2_CL 6
+#endif // LOG2_CL
+
+static uint64_t cacheline_aligned_addr(uint64_t num) {
+  return num >> LOG2_CL;
+}
+
+uint64_t operator "" _CL(unsigned long long int num) {
```
tswhison: Using the literal operator violates the [Google coding guidelines](https://google.github.io/styleguide/cppguide.html#Operator_Overloading).

libopae++/samples/hello_fpga-1.cpp
r-rojo: Thanks for your feedback. Helper functions was something that had been up for debate too. We started putting helper functions in another file to track them. We were thinking this other file would probably be part of the second layer so it shouldn't be considered as part of this first layer. The file can be found here for now (in feature/c++ branch): `common/include/opae/cxx/buffers.h`
Perhaps we'll move `fill` and `equal` to there too.

tswhison: I wonder if we should leave them in dma_buffer.  My thinking is that the base layer should provide what is necessary to perform a minimal AFU interaction (Lpbk1). In that case, if we removed these, then the user would need to acquire the raw buffer pointer and do the initialization / comparison on their own.

michael-adler: I think a helper function in dma_buffer to zero the buffer is perfectly reasonable. It's something likely to be common across applications. Setting a buffer to a repeated value of a particular byte is unlikely to be universal and the semantics of the call are not at all clear, since it isn't obvious that fill is byte sized. I don't think you should be catering to the needs of lpbk1 in the class.

PR #166, opened 2018-01-17
Detect packager app run time dependency information during build

rpSTTD:


michael-adler:
The check ignores the current PYTHONPATH environment variable. I have jsonschema on PYTHONPATH, which works, but the warning fires.

michael-adler:
Still not working. I'll look into what's going on tomorrow.

rpSTTD:
How did you setup $PYTHONPATH on your system?

michael-adler:
It's an environment variable in my shell.

rpSTTD:
if you have these folds $PYTHONPATH/jsonschema  & $PYTHONPATH/repoze on your system, the find_path macro shall work.

michael-adler:
It's loaded through an egg directory:

```shell
[opae-sdk(1925)] python
Python 2.7.5 (default, Oct 11 2013, 14:51:32) 
[GCC 4.7.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import jsonschema
>>> print jsonschema.__file__
/p/asim/common/lib/python/lib/python2.7/site-packages/jsonschema-2.6.0-py2.7.egg/jsonschema/__init__.pyc
>>> 
```

/p/asim/common/lib/python/lib/python2.7/site-packages is on PYTHONPATH.

michael-adler:
I debugged the problem. In my installation, jsonschema isn't a directory on PYTHONPATH. The "egg" directory is. If you check sys.path in Python, it expands all the egg directories onto the path and jsonschema is there.

How about the following simpler and faster code. I suspect you could turn it into a generic macro that would work for any Python module, so you could eliminate the separate file for each module in the current implementation:

```cmake
# - Try to find python jsonschema package
# Once done, this will define
#
#  jsonschema_FOUND - system has python jsonschema & repoze packages

execute_process (COMMAND python -c "import jsonschema" ERROR_QUIET RESULT_VARIABLE jsonschema_NOT_FOUND)

if(jsonschema_NOT_FOUND)
  message(WARNING "PACKAGER runtime depenency python jsonschema package is missing!!!")
else(jsonschema_NOT_FOUND)
  set(jsonschema_FOUND true)
endif(jsonschema_NOT_FOUND)
```

cmake/modules/FindPythonPackage.cmake
```diff
@@ -0,0 +1,9 @@
+# - Find python package installation status
+
+macro(FIND_PYTHON_PKG PKG_NAME )
```
michael-adler: How about an optional argument that controls the severity (warning vs. error)?

You've chosen to make these warnings. Why? packager and ASE configuration will fail without jsonschema. Wouldn't it be better to force the environment to be correct at config/build time instead of a more obscure message later?

tools/packager/CMakeLists.txt
```diff
@@ -27,6 +27,11 @@
 cmake_minimum_required(VERSION 2.8.12)
 include(packaging)
 
+####### check python dependency ##############################################
+include(FindPythonPackage)
+FIND_PYTHON_PKG("jsonschema" SEND_ERROR)
+FIND_PYTHON_PKG("repoze" SEND_ERROR)
```
michael-adler: Nice!

I just checked the DCP environment. Of course they have jsonschema, since they need it. I don't see repoze installed. Perhaps that should be a warning until you get them to install it? Otherwise, you'll break all their regressions.

Where is repoze used?

omkarhegde: Yeah, I don't think we have any dependency on repoze.

rpSTTD: latest commit removed the python repoze dependency check.

michael-adler:
DCP tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2123685

There are OpenCL, but don't appear to be related to this change.

PR #178, opened 2018-01-19
opaec++: add another version of hello_fpga

r-rojo:
This version uses the opae::fpga::application API to define a class, `nlb0` which derives from `component`.
The class is passed into the `accelerator::run` staic function as a template argument.
An instance of this class is passed into the callback.
This is another hypothetical API.
_This will not compile_

### DIFF
--- a/libopae++/samples/hello_fpga-4.cpp
+++ b/libopae++/samples/hello_fpga-4.cpp
@@ -1,0 +1,143 @@
+// Copyright(c) 2017, Intel Corporation
+//
+// Redistribution  and  use  in source  and  binary  forms,  with  or  without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of  source code  must retain the  above copyright notice,
+//   this list of conditions and the following disclaimer.
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+// * Neither the name  of Intel Corporation  nor the names of its contributors
+//   may be used to  endorse or promote  products derived  from this  software
+//   without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,  BUT NOT LIMITED TO,  THE
+// IMPLIED WARRANTIES OF  MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED.  IN NO EVENT  SHALL THE COPYRIGHT OWNER  OR CONTRIBUTORS BE
+// LIABLE  FOR  ANY  DIRECT,  INDIRECT,  INCIDENTAL,  SPECIAL,  EXEMPLARY,  OR
+// CONSEQUENTIAL  DAMAGES  (INCLUDING,  BUT  NOT LIMITED  TO,  PROCUREMENT  OF
+// SUBSTITUTE GOODS OR SERVICES;  LOSS OF USE,  DATA, OR PROFITS;  OR BUSINESS
+// INTERRUPTION)  HOWEVER CAUSED  AND ON ANY THEORY  OF LIABILITY,  WHETHER IN
+// CONTRACT,  STRICT LIABILITY,  OR TORT  (INCLUDING NEGLIGENCE  OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+#include "opae/cxx/accelerator.h"
+#include "opae/cxx/buffers.h"
+#include <uuid/uuid.h>
+#include <algorithm>
+#include <iostream>
+#include <chrono>
+#include <thread>
+
+using namespace opae::fpga::resources;
+using namespace opae::fpga::application;
+
+static const char* NLB0_AFUID = "D8424DC4-A4A3-C413-F89E-433683F9040B";
+static const uint64_t CL = 64;
+static const uint64_t KB = 1024;
+static const uint64_t MB = KB * 1024;
+static const uint64_t GB = MB * 1024;
+static const uint64_t LOG2_CL = 6;
+static const size_t LPBK1_DSM_SIZE = 4 * KB;
+static const size_t LPBK1_BUFFER_SIZE = 1 * KB;
+static const size_t LPBK1_BUFFER_ALLOCATION_SIZE = 4 * KB;
+static const uint32_t DSM_STATUS_TEST_COMPLETE = 0x40;
+
+static inline uint64_t cacheline_aligned_addr(uint64_t num) {
+  return num >> LOG2_CL;
+}
+
+
+struct nlb0 : public component
+{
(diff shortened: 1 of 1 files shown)


libopae++/samples/hello_fpga-4.cpp
```diff
@@ -0,0 +1,143 @@
...
+  // Call the accelerator::run static function with the following args:
+  // 1. The filter used to locate the accelerator
+  // 2. The flags used when opening the accelerator
+  // 3. The size of the initial buffer - more can be requested from accel
+  // 4. The callback called once an accelerator is found and opened
+  //    This callback takes as arguments the accelerator and the buffer allocated
+  return accelerator::run<nlb0>(filter,
+    FPGA_OPEN_SHARED,
+    1 * GB,
+    [](nlb0::ptr_t nlb)
```
michael-adler: Does everyone else like this style, with a callback and a lambda function? To me it seems as though you are heading in the direction of AAL, which wasn't a big hit. What's the advantage of this style over something more straight line?

tswhison: I like the style, but I'm also a fan of C++11 idioms. I can see that others may not share our enthusiasm. I wonder though, if we are considering the correct audience for this? Wouldn't those who might shriek at this kind of thing be more inclined to use the C-based API anyway? Do we want to design the C++ layer with those limitations in mind, or should we cater to a more C++ savvy crowd? I think AAL tended to cater to the more C++ savvy, but at that time there was no C-based alternative.

michael-adler: I'm confident that many will appreciate the sorts of things you've done with C++ to wrap C functions and hold state inside classes. They will appreciate the convenience of C++ for maintaining the handle inside a class, hiding multiple C calls in common idioms, etc.

However, I can't recall any user (at least the academic ones) who had anything good to say about AAL. In particular, they hated the callbacks and the complexity. I worry that you've gone too far here and there is a happy medium.

r-rojo: IMHO, AAL is good in spirit but some of the design decisions made it less flexible as one had to design their user specific code one way only and anything beyond SOA had to be done with wrappers. 

I like the overall direction we've been going with OPAE - starting from the bottom up. This allows us to create different layers giving end users/developers the choice to pick one that best suits their needs.

We've designed the first C++ layer with simplicity and interoperability in mind. What I'm prototyping for the second layer are higher level OO APIs that abstract out or hide some of the boilerplate while still being flexible enough to allow more explicit use of the object model.

This sample shows an interface that uses a callback (not necessarily a `lambda` - it can be any function pointer or` std::function` object)  that allows the end user/developer to focus on what he/she wants to do with an accelerator and not have to worry on the details on how to acquire one. It doesn't _have_ to be done as a callback as it can also be accomplished in the OO design of the `component` class. Say the concrete`component` (`nlb0` in this case) overrides a function called `run` with the logic in there instead of putting the logic in a callback.

These types of patterns are part of the Inversion of Control pattern or the Strategy Pattern and can be especially useful when designing something like an "acceleration engine" where the SW logic used for an accelerator can be loaded and executed at runtime.

I appreciate your feedback but my intent here is not to design the "one way" of doing things in C++ but rather show different ways that a C++ object model can be designed/used while still being flexible.

luebbers: When I first saw the design I had the same concerns (see #172). My "but... but... it's always been done like *this*" reflex was to ask for a pattern where I create a workload-specific subclass of the generic accelerator class and override its `run()` function (at least that's how I learned to specialize classes back in the 20th century). But it seems that this design doesn't preclude that approach (as long as the `run()` function used here doesn't do anything else except enumerate, open, and call the callback). We should probably provide nice, clean, and polished documentation examples for *both* ways, since I'm sure there are others will have the same reaction as we did.

But I'd also like to learn about the benefits of this style. References anyone?

libopae++/samples/hello_fpga-4.cpp
```diff
@@ -0,0 +1,143 @@
...
+  // look for accelerator with NLB0_AFUID
+  properties filter;
+  filter.guid.parse(NLB0_AFUID);
+  // Call the accelerator::run static function with the following args:
+  // 1. The filter used to locate the accelerator
+  // 2. The flags used when opening the accelerator
+  // 3. The size of the initial buffer - more can be requested from accel
+  // 4. The callback called once an accelerator is found and opened
+  //    This callback takes as arguments the accelerator and the buffer allocated
+  return accelerator::run<nlb0>(filter,
```
luebbers: I'm a bit lost on intuitive semantics here - so `nlb0` is a `component`, which is used as the template class for an `accelerator` class, which has a static method `run()` that creates the accelerator object, enumerates the accelerator based on the `filter`, and calls the provided callback function, right?

That callback then references attributes of the component `nlb0`, which it can because the accelerator was instantiated with `nlb0` as a template. Correct?

libopae++/samples/hello_fpga-4.cpp
michael-adler: Is it theoretically possible to have a default function that is the callback so that users uninterested in callbacks wouldn't have to deal with it? Perhaps that would support either flow. Otherwise, I see where you all are coming from that it's a different style. As Enno points out, you don't have to do anything interesting in the callback. A default where you don't even have to specify an empty lambda would be even better.

r-rojo: I wasn't trying to design a layer with only _one_ way of doing things. I'm thinking of something flexible enough to allow the developer to use the API with his/her own logic for finding/opening/interacting with accelerator resources (including error handling). 
On the other hand, if a developer just wants to plug in his/her algorithm to an accelerator matching certain criteria and not worry about the setup/teardown flow then this second layer could provide it too. I don't want to limit choices but provide well thought out (and documented) choices so developers can design their own workflows according to their needs (simple tool or integration with other SW).

The callback or an OO interface is simply just a contract telling the developer to expect certain conditions when his/her logic is called - _oh, and by the way - you must return something meaningful to the framework_.
A nice side effect of this method is that any exceptions thrown by our API can be caught by the caller (this framework) of the callback/run function and be handled gracefully.

If you look up strategy pattern (which to summarize):
> Strategy lets the algorithm vary independently from clients that use it.

https://en.wikipedia.org/wiki/Strategy_pattern

You can also read about Inversion of Control which states that strategy pattern is one of
> several basic techniques to implement inversion of control

https://en.wikipedia.org/wiki/Inversion_of_control

libopae++/samples/hello_fpga-4.cpp
r-rojo: `accelerator::run` is a static function that:
* uses function arguments to
  * enumerate and open a resource
  * allocate a shared buffer
* uses template argument to
  * create an instance of the class
     * the `component` constructor is called which saves the accelerator and shared buffer
     * the `nlb0` constructor and initialization list is called which
       * initialize CSR members and shared_buffer members to use the underlying accelerator or already allocated buffer


The developer of `nlb0` would only have to specify:
* csr map
* buffers needed
* algorithm to run with the accelerator (either the callback or implementation of `run`)

luebbers: Thanks. How would events work with this pattern? Would we also need to pass events we're interested in through the `run()` function arguments?

michael-adler:
Since these are examples, more comments would be really good. Describe the requirements for the class, answer questions like Enno's, describe the callback and how it's argument is constructed, etc.

luebbers:
@r-rojo Are you going to merge this?

Issue #176, opened 2018-01-19
fpgaTriggerUmsg() not found in libopae-c-ase.so


eyeloveu:
GCC outputs a link error "undefined reference to `fpgaTriggerUmsg'" when I invoke this Umsg function in my host code. 
Finally, I found there is a mismatch between the doc and the c-lib. 
The doc says 'Use fpgaTriggerUmsg() if you need ASE simulation capabilities', while there is no implementation in 'opae-sdk\ase\api\src\umsg.c'. 
Instead, fpgaGetUmsgPtr() is implemented. 
I don't know which is wrong. Doc error or code bug? 
BTW, is there any reference with more detail about the Umsg APIs? 
![1](https://user-images.githubusercontent.com/29499271/35143569-eb53dfe0-fd3c-11e7-823c-dc1372cb191d.png)
![2](https://user-images.githubusercontent.com/29499271/35143570-eb847678-fd3c-11e7-9d17-f20e5d821a0b.png)

luebbers:
Hi @eyeloveu - thank you for the issue report. It seems like the ASE frontend (libopae-c-ase) does not include an implementation of `fpgaTriggerUmsg()`. We'll add it in a subsequent pull request and update the issue - it'd be great if you could check whether our fix works for you.

Section 3.4 of the [CCI-P specification v0.5](https://01.org/sites/default/files/downloads/opae/intel-fpga-ip-cci-p-inter-spec-external-0.5.pdf) has more details on how UMsgs work.

eyeloveu:
@luebbers Thank you!

Can I use fpgaGetUmsgPtr() instead of fpgaTriggerUmsg() to implement the same CPU to FPGA notification function?

luebbers:
Yes - that's what `fpgaTriggerUmsg()` does:

    fpgaGetUmsgPtr(handle, &umsg_ptr);
    *((volatile uint64_t *) (umsg_ptr)) = value;

(+ error handling, of course).

Note that this only triggers the **first** UMsg. To trigger others, you'd need to write to different locations of the mapped buffer behind `umsg_ptr`. We're looking at adding that to `fpgaTriggerUmsg()`.

luebbers:
@eyeloveu We implemented `fpgaTriggerUmsg()` in PR #182 and merged into master. Can you try if that works for you?

eyeloveu:
@luebbers Thanks, It works for me. And I also tried:
[...]

PR #182, opened 2018-01-19
ASE: Add missing implementation for fpgaTriggerUmsg()

luebbers:
fpgaTriggerUmsg() was defined in <opae/umsg.h>, but not implemented in
libopae-c-ase. This commit adds an implementation that uses
fpgaGetUmsgPtr() to trigger an UMsg (similar to the libopae-c
implementation. It also contains an update to the Doxygen of
fpgaGetUmsgPtr() which still said that the call wasn't supported on ASE.

This addresses issue #176.


luebbers:
Can we also add a test for this somewhere?

lzhan55:
We can add a test for it.

-Tracy

rpSTTD:
Tested with ASE environment.  Test PASS.  Will submit the test to fpga-internal repo for now since ASE testing is not supported in this repo at this point.  Will ported over in the future.

PR #183, opened 2018-01-19
tools: Replace string functions with secure safe string

anandaravuri:
Changes:
   - Replace strlen() with strlen_s()
   - Replace snprintf () with snprintf_s_X ()

### DIFF
--- a/tools/fpgaconf/fpgaconf.c
+++ b/tools/fpgaconf/fpgaconf.c
@@ -273,7 +273,7 @@ int parse_args(int argc, char *argv[])
 				break;
 			endptr = NULL;
 			config.target.bus = (int) strtoul(tmp_optarg, &endptr, 0);
-			if (endptr != tmp_optarg + strlen(tmp_optarg)) {
+			if (endptr != tmp_optarg + strnlen_s(tmp_optarg, sizeof(tmp_optarg))) {
 				fprintf(stderr, "invalid bus: %s\n", tmp_optarg);
 				return -1;
 			}
@@ -284,7 +284,7 @@ int parse_args(int argc, char *argv[])
 				break;
 			endptr = NULL;
 			config.target.device = (int) strtoul(tmp_optarg, &endptr, 0);
-			if (endptr != tmp_optarg + strlen(tmp_optarg)) {
+			if (endptr != tmp_optarg + strnlen_s(tmp_optarg, sizeof(tmp_optarg))) {
 				fprintf(stderr, "invalid device: %s\n", tmp_optarg);
 				return -1;
 			}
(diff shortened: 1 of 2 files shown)


tools/fpgaconf/fpgaconf.c
```diff
@@ -273,7 +273,7 @@ int parse_args(int argc, char *argv[])
 				break;
 			endptr = NULL;
 			config.target.bus = (int) strtoul(tmp_optarg, &endptr, 0);
-			if (endptr != tmp_optarg + strlen(tmp_optarg)) {
+			if (endptr != tmp_optarg + strnlen_s(tmp_optarg, sizeof(tmp_optarg))) {
```
luebbers: How about the following faster alternative that doesn't need `strlen()` at all:

    if (*tmp_optarg == '\0' || *endptr != '\0') {
        fprintf(...)
        ...

This would treat the bus argument as invalid if it's either empty (`tmp_optarg` points to '\0') or invalid (`endptr` points to something other that a '\0').

Actually, the existing code would accept an empty argument and set the bus to `0`, so

    if (*endptr != '\0') {
        fprintf(...)
        ...

would be enough. The check for an empty argument doesn't hurt, though.

tools/fpgaconf/fpgaconf.c
```diff
@@ -306,7 +306,7 @@ int parse_args(int argc, char *argv[])
 				break;
 			endptr = NULL;
 			config.target.socket = (int) strtoul(tmp_optarg, &endptr, 0);
-			if (endptr != tmp_optarg + strlen(tmp_optarg)) {
+			if (endptr != tmp_optarg + strnlen_s(tmp_optarg, sizeof(tmp_optarg))) {
```
luebbers: Same as above for all changes.

tools/ras/main.c
```diff
@@ -1067,7 +1067,7 @@ fpga_result mmio_error(struct RASCommandLine *rasCmdLine)
 	if ( rasCmdLine->function >0 )
 		function = rasCmdLine->bus;
 
-	snprintf(sysfs_path, sizeof(sysfs_path),
+	snprintf_s_iiii(sysfs_path, sizeof(sysfs_path),
```
luebbers: Can you explain why is this necessary? Is `snprintf()` with four integer arguments unsafe?

luebbers: I see.

In this case the format string is a constant containing four `%i`, and the arguments are either constants (`0`) or `int` - is `sprintf_iiii()` necessary in this case?

tswhison: I think so, or else we could try to get a per-instance waiver; even then, we would still have to deal with the issue when Klocwork flags it.

anandaravuri:
Not need

PR #184, opened 2018-01-19
tools: MMlink set default IP address to zero (#180)

tswhison:
Changes:
 - Set socket default IP address to 0.0.0.0 instead of loop back IP address 127.0.0.1


tswhison:
This change will make the default behavior when no ip address is given "listen on any", as was the behavior before recent command line updates which changed the default to "listen on local only".

PR #186, opened 2018-01-22
Add support to AFU JSON files to request automatic frequency setting of uClk

michael-adler:
- Update JSON schema to permit "auto" as a user clock frequency.
- Add a validation function to afu.py that runs as a GBS is written, beyond the base schema.
  New code checks that user clock frequency is a number at GBS write time.
- Improve the Python error message when jsonschema isn't present.
- Add Quartus Tcl functions to the platform interface with helper functions for
  managing user clock frequencies. These Tcl functions are pure Tcl (no Quartus),
  so they work as cross-platform library functions.

### DIFF
--- a/platforms/platform_if/par/afu_json.tcl
+++ b/platforms/platform_if/par/afu_json.tcl
@@ -1,0 +1,81 @@
+##
+## Extract information from an AFU's JSON file.
+##
+
+package require json
+
+##
+## Find the JSON file in the project.  The file must have been added with
+## set_global_assignment in the MISC_FILE namespace.
+##
+## The function returns the first JSON file found.
+##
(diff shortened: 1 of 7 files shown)


michael-adler:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2125962 (OpenCL failures are unrelated)

michael-adler:
I have a Quartus phase to add to the build flow derived from the OpenCL uClk setting code. It will be easy to add: a few file updates and no structural scripting changes.

The Platform Interface Manager accepts either "auto" or "auto-<max freq>". The latter imposes a maximum frequency lower than the platform's 600 MHz limit. Ultimately, uClk frequency is set to whatever is actually achieved.

PR #190, opened 2018-01-23
Accept all Platform Interface Manager platform names in ASE configuration

michael-adler:
- Add mapping from physical platform name to ASE platforms in platform db.
- afu_platform_config writes ASE platform name to a file.
- Add the ability to set the default physical platform name in ASE using
  an environment variable (OPAE_ASE_DEFAULT_PLATFORM).

### DIFF
--- a/ase/scripts/generate_ase_environment.py
+++ b/ase/scripts/generate_ase_environment.py
@@ -377,7 +377,7 @@ def auto_find_sources(fd):
     if (len(str)):
         for js in str.split('\n'):
             try:
-                with open(js) as f:
+                with open(js, 'r') as f:
                     db = json.load(f)
                 f.close()
 
@@ -429,6 +429,7 @@ def gen_afu_platform_ifc(json_file):
         cfg.append("--ifc=" + default_ifc)
 
     if (args.platform == 'discrete'):
+        # Compatibility mode for original "discrete" platform argument.
         cfg.append("discrete_pcie3")
     else:
         cfg.append(args.platform)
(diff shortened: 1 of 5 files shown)


michael-adler:
Regressions pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2129694

This change fixes a hole in the modeling of physical platforms. There are multiple physical platforms with small variations (e.g. different number of memory banks) that all map to the same ASE platform name. The change shifts responsibility to the Platform Interface Manager to handle the mapping.

ase/scripts/generate_ase_environment.py
```diff
@@ -429,6 +429,7 @@ def gen_afu_platform_ifc(json_file):
         cfg.append("--ifc=" + default_ifc)
 
     if (args.platform == 'discrete'):
+        # Compatibility mode for original "discrete" platform argument.
         cfg.append("discrete_pcie3")
```
deepakcu84: what is discrete_pcie3?

michael-adler: The name of the RC platform interface in the Platform Interface Manager.

See https://github.com/OPAE/opae-sdk/blob/master/platforms/platform_db/discrete_pcie3.json

ase/scripts/generate_ase_environment.py
```diff
@@ -478,8 +505,9 @@ def gen_afu_json_verilog_macros(json_file):
                             will be parsed by rtl_src_config.""")
 parser.add_argument('-t', '--tool', choices=['VCS', 'QUESTA'], default=None,
                     help='simulator tool to use, default is VCS if present')
-parser.add_argument('-p', '--platform', choices=['intg_xeon', 'discrete'],
-                    default='intg_xeon', help='FPGA Platform to simulate')
+parser.add_argument('-p', '--platform',
+                    default=get_default_platform(),
+                    help='FPGA Platform to simulate')
 parser.add_argument('-x', '--exclude', default=None,
```
deepakcu84: can we add an option to list all supported platforms?

michael-adler: afu_platform_config --help

PR #191, opened 2018-01-23
 API: Add functions to retrieve version information

luebbers:
Adds API calls to retrieve version and build information from the
library:

    fpgaGetOPAECVersion()
    fpgaGetOPAECVersionString()
    fpgaGetOPAECBuildString()

These return information gathered at library compile time (either
libopae-c or libopae-c-ase); the first two provide a major.minor.revision
tuple or a string "major.minor.revision", respectively, while the latter
provides the git hash of the SDK repository at build time.


r-rojo:
Looks OK to me although I do have a question:
Should we make use of the `fpga_version` data structure? I think it's defined in types.h

luebbers:
Good point (again). Looks better and is simpler.

ase/api/src/version.c
```diff
@@ -0,0 +1,76 @@
...
+}
+
+fpga_result __FPGA_API__ fpgaGetOPAECVersionString(char *version_str, size_t len)
+{
+	if (!version_str) {
+		FPGA_MSG("version_str is NULL");
+		return FPGA_INVALID_PARAM;
+	}
+
+	strncpy_s(version_str, len, INTEL_FPGA_API_VERSION,
```
r-rojo: no need to check return value of strncpy_s?

ase/api/src/version.c
```diff
@@ -0,0 +1,76 @@
...
+
+fpga_result __FPGA_API__ fpgaGetOPAECBuildString(char *build_str, size_t len)
+{
+	if (!build_str) {
+		FPGA_MSG("build_str is NULL");
+		return FPGA_INVALID_PARAM;
+	}
+
+	strncpy_s(build_str, len, INTEL_FPGA_API_HASH,
+		  sizeof(INTEL_FPGA_API_HASH));
```
r-rojo: ditto here, no need to check the return value of strncpy_s?

libopae/src/version.c
```diff
@@ -0,0 +1,76 @@
...
+
+fpga_result __FPGA_API__ fpgaGetOPAECVersionString(char *version_str, size_t len)
+{
+	if (!version_str) {
+		FPGA_MSG("version_str is NULL");
+		return FPGA_INVALID_PARAM;
+	}
+
+	strncpy_s(version_str, len, INTEL_FPGA_API_VERSION,
+		  sizeof(INTEL_FPGA_API_VERSION));
```
r-rojo: same here about return value...?

libopae/src/version.c
```diff
@@ -0,0 +1,76 @@
...
+
+fpga_result __FPGA_API__ fpgaGetOPAECBuildString(char *build_str, size_t len)
+{
+	if (!build_str) {
+		FPGA_MSG("build_str is NULL");
+		return FPGA_INVALID_PARAM;
+	}
+
+	strncpy_s(build_str, len, INTEL_FPGA_API_HASH,
+		  sizeof(INTEL_FPGA_API_HASH));
```
r-rojo: And here?

luebbers:
I think it wouldn't hurt to have the library print its version when `LIBOPAE_LOG` is set to >0. You're thinking of using C constructor functions for that?

r-rojo:
Yes, the C constructor similar to what I did for the log.c file:
https://github.com/OPAE/opae-sdk/blob/5fda41a8467083c76dac8a4677f18cd684bebe9b/libopae/src/log.c#L43-L84

luebbers:
Cool - but let's do that in a separate pull request.

luebbers:
Any other suggestions for changes, or can we get this merged?

lzhan55:
Are you going to add fpgaGetOPAECppVersion() in the future?

luebbers:
Good point - that should be part of the C++ binding, I guess. I don't think there's a use case for someone only using libopae-c to query the version of libopae-cpp using an API implemented in libopae-c.

@r-rojo Can you track this for the C++ API? Thanks!

luebbers:
API and ASE tests should be the same; the version API does not interact with any hardware or the driver.

I've added the function tests (which of course invalidated all the other reviews). Please review.

common/include/opae/version.h
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Get version information about the OPAE library as a string
+ *
+ * Retrieve major version, minor version, and revision information about the
+ * OPAE library, encoded in a human-readable string (e.g. "1.0.0").
+ *
+ * @param[out]  version_str  String to copy version information into
+ * @param[in]   len          Length of `version_str`
+ * @returns FPGA_INVALID_PARAM if `version_str` is NULL, FPGA_OK otherwise.
```
r-rojo: What about `FPGA_EXCEPTION`?

common/include/opae/version.h
```diff
@@ -0,0 +1,79 @@
...
+fpga_result fpgaGetOPAECVersion(fpga_version *version);
+
+/**
+ * Get version information about the OPAE library as a string
+ *
+ * Retrieve major version, minor version, and revision information about the
+ * OPAE library, encoded in a human-readable string (e.g. "1.0.0").
+ *
+ * @param[out]  version_str  String to copy version information into
+ * @param[in]   len          Length of `version_str`
```
r-rojo: So it's possible for users to allocate less memory than required?
Have you thought about passing in a `char**` and having this allocate the buffer and passing back the size in an output variable? This would require that the user free this buffer too though so....

common/include/opae/version.h
```diff
@@ -0,0 +1,79 @@
...
+#define FPGA_VERSION_STR_MAX 10
+
+/**
+ * Get build information about the OPAE library as a string
+ *
+ * Retrieve the build identifier of the OPAE library.
+ *
+ * @param[out]  build_str  String to copy build information into
+ * @param[in]   len        Length of `build_str`
+ * @returns FPGA_INVALID_PARAM if `build_str` is NULL, FPGA_OK otherwise.
```
r-rojo: What about `FPGA_EXCEPTION`?

common/include/opae/version.h
```diff
@@ -0,0 +1,79 @@
...
+fpga_result fpgaGetOPAECVersionString(char *version_str, size_t len);
+#define FPGA_VERSION_STR_MAX 10
+
+/**
+ * Get build information about the OPAE library as a string
+ *
+ * Retrieve the build identifier of the OPAE library.
+ *
+ * @param[out]  build_str  String to copy build information into
+ * @param[in]   len        Length of `build_str`
```
r-rojo: Same comment as above about passing in `char**` for the build_str...

tests/function/gtVersion.cpp
```diff
@@ -0,0 +1,145 @@
...
+	char have[80];
+
+	snprintf(want, 80, "%d.%d.%d",
+			INTEL_FPGA_API_VER_MAJOR,
+			INTEL_FPGA_API_VER_MINOR,
+			INTEL_FPGA_API_VER_REV
+		);
+
+	EXPECT_EQ(FPGA_OK, fpgaGetOPAECVersionString(have, 80));
+	EXPECT_STREQ(want, have);
```
r-rojo: Can we replace `want` with `"INTEL_FPGA_API_VER_MAJOR.INTEL_FPGA_API_VER_MINOR.INTEL_FPGA_API_REV"`

tests/function/gtVersion.cpp
```diff
@@ -0,0 +1,145 @@
...
+ *             fpgaGetOPAECBuildString, the returned string equals
+ *             the constant defined in config_int.h (and the
+ *             function returns FPGA_OK)
+ */
+TEST(LibopaecVersionCommonALL, version_08) {
+	char want[] = INTEL_FPGA_API_HASH;
+	char have[80];
+
+	EXPECT_EQ(FPGA_OK, fpgaGetOPAECBuildString(have, 80));
+	EXPECT_STREQ(want, have);
```
r-rojo: can we replace `want` with `"INTEL_FPGA_API_HASH"`?

tests/CMakeLists.txt
```diff
@@ -75,7 +75,8 @@ set(SRC gtmain.cpp
   function/gtEnumerate.cpp
   function/gtHostif.cpp
   function/gtMMIO.cpp
-  function/gtEvent.cpp)
+  function/gtEvent.cpp
+  function/gtVersion.cpp)
```
r-rojo: I see you added tests but they aren't run as part of the tests.

tests/function/gtVersion.cpp
luebbers: That won't work. Preprocessor macros are not expanded within string constants.

If the constants were strings instead of integers, you could do

    INTEL_FPGA_API_VER_MAJOR "." INTEL_FPGA_API_VER_MINOR "." INTEL_FPGA_API_REV

but they aren't, so we can't. ;)

tests/function/gtVersion.cpp
luebbers: Yes. I did this for consistency between `version_08` and `version_05`.

common/include/opae/version.h
luebbers: We should let users have control over memory allocation where we can. There are cases where we can't avoid allocations within API functions, but I think this is not one of those cases.

common/include/opae/version.h
luebbers: Right. I'll add it.

tests/CMakeLists.txt
luebbers: Right - how do I add them there?

tests/function/gtVersion.cpp
r-rojo: aaah... 👍

tests/CMakeLists.txt
r-rojo: So only a few tests are currently being run using a filter:
https://github.com/OPAE/opae-sdk/blob/7d4e3fb03cd78a1e8dc72329551670ab4c974c94/tests/CMakeLists.txt#L108-L118

The test is named: `test_open_common_mock_drv`

I suggest using narrower filters for the tests. For example, in the C++ tests, I added a macro to add tests with filters. Calling this macro looks like this:
https://github.com/OPAE/opae-sdk/blob/ee8a9e184a2053b6660fccf1ff3db04579c338c3/tests/CMakeLists.txt#L171-L177

And in travis it looks like:
image

common/include/opae/version.h
r-rojo: What if we create a `const char*` in the .c file using these macros and return it (instead of the exit code) from these functions?

rpSTTD:
"COMMAND gtapi  ${CMAKE_BINARY_DIR} --gtest_filter="*MOCK*:-*event_drv_05*:-*nodrv*") " in opae-sdk/tests/CMakeLists.txt execute tests.
you can add "COMMAND gtapi  ${CMAKE_BINARY_DIR} --gtest_filter="*CommonALL*")" in CMakeLists.txt.  I will create a macro in the future to simplifiy the syntax for the test filters.

luebbers:
I added a separate test case in tests/CMakeLists.txt so that the tests are run. The existing test case `test_open_common_mock_drv` seems to be broken - probably because the filter expression is encased in double quotes, which ctest doesn't properly escape when running the test. We should fix that separately.

luebbers:
Thanks for the hints on how to add tests!

common/include/opae/version.h
luebbers: I don't really want to break the pattern of "every fpga* function returns an fpga_result".

r-rojo: 👍

luebbers:
Test CMake configuration was refactored, so I had to merge again from master. Please review again.

lzhan55:
At line 61 of ase/api/src/version.c, strncpy_s() should be replaced with ase_strncpy_s() since Michael Adler made some changes a while ago to remove the dependency of safe string library from ASE.

luebbers:
@zhangzhang10 Can I get a review from you on this?

PR #196, opened 2018-01-24
Bist behavior Run only one card

buie1:
For systems with multiple card must provide a bdf using the 'b', 'd' or 'f' flags.
When there is only one card on the system the tool will default to running on the only available fpga.

### DIFF
--- a/tools/fpgabist/fpgabist
+++ b/tools/fpgabist/fpgabist
@@ -34,49 +34,46 @@ import bist_common
 import bist_nlb3
 import bist_dma
 
-FEATURES = ['fme', 'port', 'temp', 'errors all']
+FEATURES = ['fme', 'port', 'temp', 'power', 'errors all']
 
 
 def main(args):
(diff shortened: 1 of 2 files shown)


tools/fpgabist/fpgabist
```diff
@@ -34,49 +34,46 @@ import bist_common
...
-            mode.run(path, bdf['bus'])
+    gbs_paths = args.gbs_paths
+    for path in gbs_paths:
+        mode_name = bist_common.get_mode_from_path(path)
+        if mode_name == 'nlb_mode_3':
+            mode = bist_nlb3.Nlb3Mode()
+        elif mode_name == "dma_afu":
+            mode = bist_dma.DmaMode()
+        else:
+            print "Unknown AFU for available BIST Modes: {}\n".format(path)
```
dasm: L73-74
You could use the same way of exiting function like above:
`sys.exit('Text')`

PR #197, opened 2018-01-25
Add QIP file support to rtl_src_config

michael-adler:


michael-adler:
Evan needs this for HSSI.

Tests pass:  http://sw-pert.altera.com/pert/pert.php?test_run_id=2135918

PR #198, opened 2018-01-25
Tests: Rename test case names

nakulkorde:
Rename test case names based on the test case naming conventions.


r-rojo:
Why are we including `MOCK` in test names? Are these tests only meaningful when preloading libmock.so?
I propose we disable tests that will only work on HW. We can enable them via the command line when running on real hardware like so:
`./gtapi --gtest_also_run_disabled_tests ...`

This way, all non-disabled tests will run as part of the travis build.

rpSTTD:
Test name should be LibopaecBufCommonHWMOCK if a test works on both HW & MOCK. Please follow the test name convention at https://github.intel.com/OPAE/opae-sdk-x/wiki/Types-of-test-cases to make the proper name change.

PR #199, opened 2018-01-25
Update afu_synth_setup to work with discrete build structure

michael-adler:
- Specify hw/lib directory, not release, since there may not be a release yet.
- Remove "_std" in the name and use an environment variable to set the
  hw/lib path.


michael-adler:
Needed by Chip for configuring with Platform Interface Manager in the middle of the build flow.

No point in regression testing -- the script isn't used anywhere yet. Tested by hand.

PR #200, opened 2018-01-29
Use Python to generate zipped Python executables

michael-adler:
- No longer depends on Unix shell commands
- No longer depends on zip program

### DIFF
--- a/cmake/modules/packaging.cmake
+++ b/cmake/modules/packaging.cmake
@@ -79,7 +79,6 @@ endfunction(DEFINE_PKG)
 
 macro(CREATE_PYTHON_EXE EXE_NAME MAIN_MODULE)
 
-    set(PYTHON_SRC "${ARGN}")
     set(PACKAGER_BIN ${PROJECT_BINARY_DIR}/bin/${EXE_NAME})
 
     # Generate a __main__.py that loads the target module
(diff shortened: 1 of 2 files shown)


michael-adler:
Passes DCP regressions: http://sw-pert.altera.com/pert/pert.php?test_run_id=2143502

cmake/modules/packaging.cmake
```diff
@@ -91,16 +90,45 @@ macro(CREATE_PYTHON_EXE EXE_NAME MAIN_MODULE)
...
+        "  f.write(wr_buf.getvalue())\n"
+        "  # Mark the file executable\n"
+        "  mode = os.fstat(f.fileno()).st_mode\n"
+        "  mode |= stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH\n"
+        "  os.fchmod(f.fileno(), stat.S_IMODE(mode))\n"
+        "\n"
+        "f.close()\n")
+
+    # Run Python to generate the zipped file
+    execute_process(COMMAND python "${BUILD_DIR_MAIN}/do_zip.py"
```
r-rojo: Should we call `find_package(PythonInterp)` somewhere and use its cmake vars for pointing to the Python interpreter?
This will allow us to specify versions of Python (if we need to) and handle cases where Python isn't found.

michael-adler: Good plan. There's some refactoring that could be done, too. The macro I changed here should probably be moved to a different file. E.g., FindPythonPackage.cmake in the same directory also invokes Python and may be a better home than packaging for the Python zip script.

r-rojo: Eventually, I think we should use Python's setuptools to make wheel packages.
See: https://packaging.python.org/guides/tool-recommendations/

PR #203, opened 2018-01-30
Platform support changes:

michael-adler:
- Add a new stub for a platform with HSSI
- Teach ASE to find the physical platform using a mechanism identical to
  Quartus builds.  The current FPGA thus matches the simulation environment.
- Update tags on RTL registers used for timing relaxation based on
  experience with OpenCL.


michael-adler:
Regressions: https://sj-arc.altera.com/arc/dashboard/reports/show_job/664221413

A couple ModelSim/Questa jobs look funny, but there's nothing I did that could cause this. None of the changes here are in an active path in these tests.

PR #204, opened 2018-01-31
opaec++: Add prototype of accelerator classs

r-rojo:
This adds/updates:
* first version of accelerator with `enumerate` and `open` functions
* hello_fpga-2.cpp to only use implemented functions in accelerator
* Unit test for accelerator::open
* libopae-cxx.so library


common/include/opae/cxx/accelerator.h
```diff
@@ -0,0 +1,104 @@
...
+   *
+   * @brief     Defines an alias representing a list of accelerators
+   */
+  typedef std::vector<ptr_t> list_t;
+
+  /**
+   * @brief     Copy constructor
+   * @param     other   The other accelerator object.
+   */
+  accelerator(const accelerator& other);
```
tswhison: I'm not sure how an accelerator could be copied or copy-assigned, because the handle_ is not a copy-able object.

common/include/opae/cxx/accelerator.h
```diff
@@ -0,0 +1,104 @@
...
+   */
+  static list_t enumerate(std::vector<properties> filter = { FPGA_ACCELERATOR } );
+
+  void open(int flags);
+
+ private:
+  token::ptr_t token_;
+  properties::ptr_t props_;
+  handle::ptr_t handle_;
+  accelerator() = delete;
```
tswhison: deleted special functions should appear in public scope.

libopae++/src/accelerator.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+namespace opae {
+namespace fpga {
+namespace resource {
+
+accelerator::accelerator(token::ptr_t t)
+    : token_(t), props_(properties::read(t)) {}
+
+accelerator::accelerator(const accelerator& other)
+    : token_(other.token_), props_(other.props_), handle_(other.handle_) {}
```
tswhison: handle_ is not copy-able nor copy-assignable.  This is building for you?  Hmm..

common/include/opae/cxx/accelerator.h
luebbers: Agree. Within a process, a handle is essentially a singleton. Do you want to create proxy objects on copy? AAL did that for the AIA, but I'm not sure if there'd be a big benefit here instead of handing around pointers or references.

PR #205, opened 2018-01-31
ASE: Replacing Linux commands in ASE python script

lzhan55:
Removing the embedded Linux command in ASE generate_ase_environment.py and replaced it with python commands. 

This change was done to make ASE porting to Windows possible.

Tested it with 01_hello_world sample.


michael-adler:
Did you run DCP regressions? They rely heavily on the find path, so would be a good test.

lzhan55:
DCP regression tests: http://sw-pert.altera.com/pert/pert.php?test_run_id=2152408

PR #206, opened 2018-02-01
fpgainfo: wrap stdout with codec StreamWriter 

r-rojo:
This change replaces `sys.stdout` with a StreamWriter object that uses
`UTF-8` encoding. This will allow fpgainfo to print unicode (non-ascii).
Without this, an exception will be raised when printing unicode
characters and redirecting the output stream.

Conflicts:
	tools/fpgainfo/fpgainfo


r-rojo:
Yeah. Python won't "know" encoding information when you redirect stdout/stderr. This is a cherry-pick

PR #208, opened 2018-02-01
Use BBS_LIB_PATH everywhere for finding the release HW library.

michael-adler:


michael-adler:
Tested by starting a build: https://sj-arc.altera.com/arc/dashboard/reports/show_job/665420399

It is already well beyond the point that the script is used in the flow and the script worked.

PR #218, opened 2018-02-08
Update documentation to include Sphinx target (-DBUILD_SPHINX_DOC=ON)

abelardojarab:
This patch updates the documentation to list the `-DBUILD_SPHINX_DOC=ON|OFF` optional flag for cmake.

* Note 1:
This patch is already enabled in cmake (by default it is set to OFF), this patch only updates the documentation.

* Note 2:
Sphinx documentation requires some Python packages, which can be installed with:

```console
$ pip install --user doc/sphinx/requirements.txt
```

Generated documentation documentation will appear under:

`${CMAKE_BINARY_DIR}/sphinx/major.minor.rev/html/index.html`

image


shivarao:
@chipb can you eyeball whether this will effect the build?

doc/src/build_chain/fpga_api/api_build.md
```diff
@@ -63,14 +85,16 @@ Valid «user targets» are:
   make target                Purpose
   -------------------------- --------------------------------------------------------------------------------------
   make                       Compiles the Intel FPGA API libraries, sample applications, utilities
-  make dist                  Creates distributable tarball package intel-fpga\_0.1.0.tar.gz
-  make package               Create Redhat installer package: intel-fpga\_0.1.0\_1.x86\_64.rpm
+  make dist                  Creates distributable tarball package intel-fpga_0.1.0.tar.gz
+  make package               Create Redhat installer package: intel-fpga_0.1.0_1.x86_64.rpm
```
luebbers: Not sure about the package names, how are they updated?
Are the backslashes not required anymore?

abelardojarab: We dont need backslashes anymore, unless we export to LaTeX/PDF.. so i guess we should keep them.

also as you mentioned the file names and version number are outdated. i think we should configure this file with a template such that the version and package names are correct

PR #220, opened 2018-02-09
Preparation for HSSI support in PIM

michael-adler:
- Add the ability to extend the set of ports requested in an AFU JSON file.  Without this, we'd have to explicitly specify all possible port combinations in afu_top_ifc_db.
- Add the ability for JSON files to include other JSON files in order to break some apart. platform_defaults was becoming unreadable.
- Change "module-arguments" to "module-ports" for proper Verilog naming.
- Add a through path for the HSSI RAW interface.


michael-adler:
Once this is pushed HSSI can be pushed through to ccip_std_afu using a standard interface.

Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2173025

platforms/afu_top_ifc_db/README.md
```diff
@@ -96,19 +96,19 @@ Each of the module-arguments is a dictionary with the following keys:
...
 - **vector**: Boolean [optional (defaults to false)]
 
-  When true, the module argument is a vector of multiple instances
-  of the argument.  For example, local-memory banks may be passed
+  When true, the module port is a vector of multiple instances
+  of the port.  For example, local-memory banks may be passed
   as a vector of memory interfaces.  The minimum and maximum number
   of entries is specified using min-entries and max-entries.
 
 - **min-entries**: Integer [optional (defaults to 1), must be > 0]
```
deepakcu84: Does your parser enforce these requirements?

platforms/platform_db/discrete_pcie3_hssi40.json
```diff
@@ -50,6 +51,11 @@
             "interface": "avalon_mm_legacy_wires_2bank",
             "optional": true,
             "define": [ "INCLUDE_DDR4" ]
+         },
+
+         {
+            "class": "hssi",
+            "interface": "raw_pr"
```
deepakcu84: What is pr?

platforms/afu_top_ifc_db/README.md
michael-adler: Yes.

platforms/platform_db/platform_defaults/defaults_hssi.json
```diff
@@ -0,0 +1,33 @@
+{
```
deepakcu84: I believe this file isn't complete?

platforms/platform_db/discrete_pcie3_hssi40.json
michael-adler: Partial reconfiguration. The data structure is pr_hssi_if.

Ultimately, this is an interface class we don't expect many customers to use. We are expecting most will let us instantiate a MAC in the PIM and expose Avalon streams for the data. The "raw_pr" mode lets the current MACs work.

platforms/platform_if/rtl/device_cfg/hssi_cfg_pkg.sv
```diff
@@ -0,0 +1,66 @@
...
+// Platform-specific HSSI configuration.
+//
+// It is assumed that this package will NOT be wildcard imported.  The
+// package name serves as a prefix instead of making all symbols inside
+// the package long.
+//
+
+package hssi_cfg_pkg;
+
+    parameter VERSION_NUMBER = 1;
```
deepakcu84: How is the version number chosen? Does the version number correspond to the interface version or the DB version?

platforms/platform_db/platform_defaults/defaults_hssi.json
michael-adler: In what way? Its HSSI-only, included by platform_defaults.json above. The definitions there are sufficient for the current HSSI MACs, but will have to be elaborated when the streaming interfaces are defined.

platforms/platform_if/rtl/device_cfg/hssi_cfg_pkg.sv
michael-adler: DB. It was chosen when I entered a 1. :-)

platforms/platform_db/platform_defaults/defaults_hssi.json
deepakcu84: The reason I ask is I don't see the default HSSI interface (raw_pr versus streaming) in this file. But, I see a comment which says there are two interfaces. How is the default interface specified?

platforms/platform_if/rtl/platform_shims/platform_shim_ccip_std_afu.sv
```diff
@@ -180,6 +184,10 @@ module platform_shim_ccip_std_afu
         .local_mem(afu_local_mem),
 `endif
 
+`ifdef AFU_TOP_REQUIRES_HSSI_RAW_PR
+        .hssi(hssi),
```
deepakcu84: Neat!

platforms/platform_db/platform_defaults/defaults_hssi.json
michael-adler: Ah, good question.

There is actually no interface defined here. The PIM only knows about classes (CCI-P, local_memory, hssi) and interfaces names within those classes. All it does is match a class/interface pair specified in a platform (e.g. hssi/raw_pr in https://github.com/OPAE/opae-sdk/blob/mcadler/plat_if_hssi/platforms/platform_db/discrete_pcie3_hssi40.json) with a request in an AFU JSON. There is no notion of an actual Verilog port defintion -- just a name.

Yesterday I realized I'd backed myself into exactly the corner I was claiming to solve. It used to be that an AFU could only ask for an interface specified in https://github.com/OPAE/opae-sdk/tree/mcadler/plat_if_hssi/platforms/afu_top_ifc_db. Then I realized I'd have to enumerate all possible combinations of interfaces there. Part of this change eliminates that requirement. An AFU can now start with an interface in afu_top_ifc_db but then add extended port requests. If you look at the intel-fpga-afu pull request you'll see the AFU JSON for the MACs and how this is done.

I haven't added definitions for HSSI flavors other than the raw HSSI to an AFU MAC yet.

PR #222, opened 2018-02-12
Emit Tcl with PIM configuration

michael-adler:
This will make it possible to drive synthesis from the PIM decisions.  E.g., Quartus can detect
that HSSI is in use with:
    [info exists platform_cfg::AFU_TOP_REQUIRES_HSSI_RAW_PR]


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2174609
Also built a few AFUs (outside arc)

PR #223, opened 2018-02-12
Make it possible to request any clock for CCI-P & local memory

michael-adler:
- Clock can be any string that maps to an available clock.
- Export CCI-P clock with `PLATFORM_PARAM_CCI_P_CLOCK.  The previous method
  using a Verilog module wasn't flexible enough to handle all the possible
  clocks.
- Update the macro writing rules in afu_platform_config to deal with more
  possible illegal macro characters.

### DIFF
--- a/platforms/platform_if/rtl/device_if/ccip_if_clock.sv
+++ b/platforms/platform_if/rtl/device_if/ccip_if_clock.sv
@@ -30,11 +30,13 @@
 
 `include "platform_if.vh"
 
+// ************************************************************************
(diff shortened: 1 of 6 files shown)


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2175505
Don't know what went wrong with hello_intr_afu_questa. Ran it again: http://sw-pert.altera.com/pert/pert.php?test_run_id=2175517

PR #228, opened 2018-02-13
Enable Werror

southerngs:
Treat warnings as errors with Werror flag.

Fix some warnings in ASE code.

Fix / mitigate some warnings in DMA code in BIST tool.


lzhan55:
Changes related to ASE are OK. Need reviews from other components.

southerngs:
@luebbers @rahulrs I finally got around to making sure this works with the unit tests and with DCP.  Can someone review to make sure it still looks okay.  Also I guess with the new system one of the maintainers should merge the PR once they agree that it looks good.

tools/fpgabist/dma/fpga_dma.c
```diff
@@ -738,9 +738,15 @@ static fpga_result poll_interrupt(fpga_dma_handle dma_h) {
 		res = FPGA_EXCEPTION;
 	} else {
 		uint64_t count = 0;
-		read(pfd.fd, &count, sizeof(count));
-		debug_print("Poll success. Return = %d, count = %d\n",poll_res, (int)count);
-		res = FPGA_OK;
+		ssize_t bytes_read = read(pfd.fd, &count, sizeof(count));
+		if(bytes_read <= 0) {
+			fprintf( stderr, "Error: %s\n",
+				bytes_read < 0 ? strerror(errno) : "zero bytes read");
```
r-rojo: Should we provide a little more context in the error message? Maybe throw in something about what operation was being performed?

tools/fpgabist/dma/fpga_dma.c
```diff
@@ -1030,11 +1036,17 @@ fpga_result fpgaDmaTransferSync(fpga_dma_handle dma_h, uint64_t dst, uint64_t sr
 return res;
 }
 
+
+// TODO: warnings disabled until function implemented. Remove after
+// implementation complete
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
 fpga_result fpgaDmaTransferAsync(fpga_dma_handle dma, uint64_t dst, uint64_t src, size_t count,
 										  fpga_dma_transfer_t type, fpga_dma_transfer_cb cb, void *context) {
```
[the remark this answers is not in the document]
southerngs: @buie1 thanks for the note.  I didn't actually touch this line of code, I just added the pragmas above it.  As you note there is a lot of coding style stuff to fix in the DMA code, this PR just fixes the warnings so we can enable Werror.

tools/fpgabist/dma/fpga_dma.c
southerngs: @r-rojo that makes sense.  I added error check to deal with the warning triggered by not checking return value of read, but don't really expect it to be triggered.  But if it ever is, whoever if debugging the problem might appreciate having more information.

r-rojo: 👍

buie1:
@southerngs Do you have a solid understanding of these code owners? You have approvals from @OPAE/ci and @OPAE/ase , however it still says you need approval from @OPAE/tools even though this owner wasn't automatically listed as a review, it seems.

southerngs:
@buie1 I think that @r-rojo is in @OPAE/tools.  I know he approved once, but I made changes so there needs to be another approval.  The merging process is still not completely clear to me.  I think the appropriate maintainer is supposed to merge the change when they are happy with it.  I'm not sure how that works when the PR touches code with multiple maintainers.

PR #230, opened 2018-02-14
ASE Makefile generation using cmake

abelardojarab:
* Objective:

a. Improve synchronization between the C and SystemVerilog data structures by utilizing the generated DPI C header file when compiling the ASE server C code (current Makefile ignores it)

e.g. In fact, the file ase_common.h is redundant with several portions with the platform_dpi.h file generated by Verilog compilation in Modelsim. I attach part of code that is generated on platform_dpi.h and which does not need to be redefined in ase_common.h. This is just one of several redefined data structures (since DPI header file is not included in the C code).

```c
#ifndef MTI_INCLUDED_TYPEDEF_ase_cfg_t
#define MTI_INCLUDED_TYPEDEF_ase_cfg_t

typedef struct {
    int ase_mode;
    int ase_timeout;
    int ase_num_tests;
    int enable_reuse_seed;
    int ase_seed;
    int enable_cl_view;
    int usr_tps;
    int phys_memory_available_gb;
}  ase_cfg_t;

#endif
```

b. Enable generation of Makefiles that are compatible not only with GNU make but also with Visual Studio nmake, ninja, etc.
c. Enable Cross platform support (Windows, Linux)
d. Increase C code coverage by enabling change on ASE server C code compiler flags.
e. Leverage cmake to discover location of DPI header files, OPAE, Questa/VCS and Quartus.

* How to use it:

1.	Build OPAE+ASE as usual: $cmake <<path to source code here>> -DBUILD_ASE=ON, and then perform a system wide install (usually under /usr/local/...)

2.	After installing OPAE, create a **new** building directory (e.g nlb_test)
```console
$ mkdir test; cd test
$ cp -r -L /usr/local/share/opae/ase/samples/intg_xeon_nlb/* .
```
3.	Run cmake in this directory
```console
$ cmake /usr/local/share/opae/ase
```

4.	Compile both C code and Verilog files
```console
$ make
```
5.	Run generated ase_server (eventually I will add it as an extra target `make sim`)

```console
$ ase_server
```


abelardojarab:
image

abelardojarab:
image

abelardojarab:
image

lzhan55:
Good job! I will try it out tomorrow on Linux. 

I am also looking forward to a Windows version working soon without OPAE installation.

ase/rtl/CMakeLists.txt
```diff
@@ -0,0 +1,153 @@
...
+## INTERRUPTION)  HOWEVER CAUSED  AND ON ANY THEORY  OF LIABILITY,  WHETHER IN
+## CONTRACT,  STRICT LIABILITY,  OR TORT  (INCLUDING NEGLIGENCE  OR OTHERWISE)
+## ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
+## POSSIBILITY OF SUCH DAMAGE.
+
+cmake_minimum_required(VERSION 2.8.12)
+project("opae-c-ase-platform")
+
+# find afu_sim_setup
+find_program(AFU_PLATFORM_CONFIG
```
michael-adler: What's going on with this find_program? afu_platform_config should be in OPAE SDK bin.

ase/rtl/CMakeLists.txt
```diff
@@ -0,0 +1,153 @@
...
+  /usr/bin
+  /usr/local/bin
+  /opt/local/bin
+  DOC "AFU platform configuration utility")
+
+# Directories
+set(ASE_SERVER_RTL     ${PROJECT_SOURCE_DIR})
+set(ASE_SCRIPTS_IN     ${PROJECT_SOURCE_DIR}/../in)
+
+set(PLATFORM_SHARE_DIR ${CMAKE_SOURCE_DIR}/../platform)
```
michael-adler: If you need any of these directories you're doing something wrong. afu_platform_config should be pointing at them when needed.

ase/rtl/CMakeLists.txt
```diff
@@ -0,0 +1,153 @@
...
+set(PLATFORM_IF_DIR    ${PLATFORM_SHARE_DIR}/platform_if)
+set(PLATFORM_IF_RTL    ${PLATFORM_IF_DIR}/rtl/device_if)
+
+## Verilog compilation
+set(VLOG_SRC_LIST
+  vlog_files.list
+  CACHE STRING "Project file listing all SystemVerilog source code for the AFU")
+
+set(ASE_PKG_FILE ${ASE_SERVER_RTL}/ase_pkg.sv)
+set(PLATFORM_PKG_FILE ${PLATFORM_IF_RTL}/ccip_if_pkg.sv)
```
michael-adler: This should come from a generated include file.

ase/rtl/CMakeLists.txt
```diff
@@ -0,0 +1,153 @@
...
+    CACHE STRING "Modelsim/Questa simulator flags")
+
+  # Define rule to generate platform_afu_top_config.vh
+  set(questa_flags "")
+
+  # afu_platform_config --sim --tgt=rtl --src ccip_std_afu.json  intg_xeon
+  file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/platform_includes)
+  add_custom_command(OUTPUT "platform_includes/platform_afu_top_config.vh"
+    WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
+    COMMAND ${AFU_PLATFORM_CONFIG} --sim --tgt=platform_includes --src ${CMAKE_BINARY_DIR}/ccip_std_afu.json ${ASE_PLATFORM_ABBREV})
```
michael-adler: I assume this reference to CMAKE_BINARY_DIR/ccip_std_afu.json is wrong, but don't really understand the code. The JSON comes from the sources.

ase/samples/intg_xeon_nlb/sources.txt
```diff
@@ -0,0 +1,51 @@
...
+# Platform includes
+platform_includes/platform_afu_top_config.vh
+-F platform_includes/platform_if_addenda.txt
+-F platform_includes/platform_if_includes.txt
+
+# ASE CCIP emulator
+-F @CMAKE_INSTALL_PREFIX@/@ASE_SHARE_DIR@/rtl/sources_abs.txt
+
+# QSYS-IPs
+@CMAKE_INSTALL_PREFIX@/@ASE_SAMPLES@/intg_xeon_nlb/src/lpbk1_RdRspRAM2PORT.v
```
michael-adler: Everything below this line should be generated by rtl_src_config

ase/rtl/CMakeLists.txt
abelardojarab: im using them as base search for other files it is ok

ase/rtl/CMakeLists.txt
abelardojarab: right; but user can find afu_platform_config under /usr/local/bin or somewhere else. so truly you should need a find_program. this is standard practice in cmake, you dont call executables directly with their names, not even gcc/make/cmake/vsim/vlog... etc.

michael-adler: Ah, I see. You might be better removing the additional PATHS and just use the default PATH in order to match the Quartus flow.

A better question is, why are you invoking afu_platform_config at all here? I suspect that most of the other problems in this CMakeLists come from trying to invoke afu_platform_config here instead of letting generate_ase_environment compile the set of sources for you that you will then import into CMake. There are lots of dynamic decisions made in the Python that became static in this example.

lzhan55:
I built OPAE and installed it at opae/myinst folder. I got errors when I tried to build the test sample: 

_[_100%] Linking C shared library libopae-c-ase-server.so
/usr/bin/ld: cannot find -lsafestr
sw/CMakeFiles/opae-c-ase-server.dir/build.make:354: recipe for target 'sw/libopae-c-ase-server.so.0.13.0' failed
make[2]: *** [sw/libopae-c-ase-server.so.0.13.0] Error 1
CMakeFiles/Makefile2:140: recipe for target 'sw/CMakeFiles/opae-c-ase-server.dir/all' failed
make[1]: *** [sw/CMakeFiles/opae-c-ase-server.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2__

michael-adler:
I assume you can drop -lsafestr since ASE doesn't need it.

luebbers:
It's getting pretty hard to review this change request. Could you please resubmit with the following points addressed?

* Reduce the size of changes to review - the current PR is over 7000 lines, partly because it looks that it includes the NLB source code
* See if you can split the changes into independent chunks that can be reviewed one by one (may not be possible)
* Target a feature branch instead of master, since this appears to be still under heavy development and unlikely to be accepted into master as-is.

Thank you!

PR #232, opened 2018-02-15
Handle Qsys IPX file hierarchy

michael-adler:
- Allow IPX files in rtl_src_config.
- Generate a master components.ipx, pointing to the sub-IPX files, in the Quartus build.

### DIFF
--- a/platforms/scripts/afu_synth_setup
+++ b/platforms/scripts/afu_synth_setup
@@ -237,6 +237,37 @@ def setup_build(args):
     cmd.append('--verilog-hdr=../hw/afu_json_info.vh')
     subprocess.check_call(cmd, cwd=build_dir)
 
+    # Handle Qsys IPX file discovery
+    setup_build_ipx(args, sources, build_dir)
+
+
+# Construct the Qsys components.ipx file in the build directory, if needed.
+def setup_build_ipx(args, sources, build_dir):
+    # Get IPX files from source list
+    cmd = ['rtl_src_config']
+    cmd.append('--ipx')
+    if (os.path.isabs(sources)):
+        cmd.append('--abs')
+        cmd.append(sources)
+    else:
+        cmd.append(os.path.relpath(sources, build_dir))
+    ipx = commands_list_getoutput(cmd, cwd=build_dir)
+    ipx = ipx.strip()
+    # Anything to do?
+    if (len(ipx) == 0):
+        return
+    ipx = ipx.split('\n')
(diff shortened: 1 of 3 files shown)


michael-adler:
This finally fixes Qsys components problems. You can now afu_synth_setup and build anywhere, even the dma_afu. Instead of copying a static components.ipx file that depends on building in exactly one place, the new code gets a list of hierarchical ipx files and constructs a top-level file in the build directory.

Tested by building dma_afu successfully. ASE regressions test the changes to rtl_src_config: http://sw-pert.altera.com/pert/pert.php?test_run_id=2184050

lzhan55:
@michael-adler In file platform_shim_ccip_std_afu.sv , you made the changes to: 
           __avalon_mem_if#(.ENABLE_LOG(0))__

michael-adler:
Good catch, but it's intentionally disabled there. Memory is still logged in ase_top. This was a secondary location at a clock crossing boundary. I decided it was both unnecessary for typical users and it was printing the wrong memory bank number due to convoluted SystemVerilog requirements. I have to revisit the bank numbering problem at some point.

platforms/scripts/afu_synth_setup
```diff
@@ -237,6 +237,37 @@ def setup_build(args):
...
+        cmd.append(os.path.relpath(sources, build_dir))
+    ipx = commands_list_getoutput(cmd, cwd=build_dir)
+    ipx = ipx.strip()
+    # Anything to do?
+    if (len(ipx) == 0):
+        return
+    ipx = ipx.split('\n')
+
+    # Emit components.ipx
+    with open(os.path.join(build_dir, 'components.ipx'), 'w') as f:
```
deepakcu84: I guess, this is fine for now. You' are likely run into issues if ACDS team changes the file format in future QII versions. Have you thought about that?

michael-adler: How is that different from the current situation? We'd have to update either the AFU source (where there is a components.idx that gets copied into the build) or we can update this script. This script could respond dynamically to the Qsys version.

What would be best is if they stop requiring the file!  I'd think it could be derived, given that they are given a set of qsys and ip files in the project.

deepakcu84: I think your best bet would be to dynamically generate  .idx based on the Qsys version. You'll have to watch out for changes between releases. In my experience, changes to IP formats are not uncommon.

PR #237, opened 2018-02-17
cmake: install python virtual env. in build dir.

r-rojo:
Check for Python module, `virtualenv`, on system.
If found, call virtualenv to install a virtual Python env. in the build directory.
This is in preparation for packaging python tools as universal wheel
files and installing them in the virtual env. during the development and testing phase.


luebbers:
Can we keep this on a feature branch until we've figured out how that would impact build, install, and packaging flows?

buie1:
@luebbers @r-rojo 
In response to this, is there a place we can track this as a potential feature?  @dasm and I would like to be included in these conversations so we can gauge the impact on the DCP team and how we can accommodate these potential changes in our development environment.

r-rojo:
I've changed this PR to target a feature branch (feature/python-dev). I also created a project to track this and linked it. See: [Python Development](https://github.com/orgs/OPAE/projects/2)

PR #238, opened 2018-02-21
Fix byteenable in the basic memory model

deepakcu84:
Signed-off-by: Deepak Unnikrishnan <deepak.unnikrishnan@intel.com>

### DIFF
--- a/ase/rtl/device_models/dcp_emif_model_basic/emif_ddr4.v
+++ b/ase/rtl/device_models/dcp_emif_model_basic/emif_ddr4.v
@@ -249,15 +249,23 @@ function automatic Response create_response (
 	return rsp;
 endfunction
 
+// Expand data enable to a data mask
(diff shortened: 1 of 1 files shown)


deepakcu84:
DMA test with address span extender (uses byteenables)

http://sj-pert.altera.com/pert/#do-not-bookmark-3

deepakcu84:
ASE tests: http://sj-pert.altera.com/pert/pert.php?test_run_id=2195171

deepakcu84:
OpenCL test failures are unrelated to this change

michael-adler:
It was a bug and not a feature of the memory subsystem?

deepakcu84:
That's correct. We discovered this bug when Chris switched his OpenCL to basic memory model. His test uses the address span expander which in turn uses byteenables. 

We never had a unit test case for byteenables w/ basic memory model. I plan to add a unit test in hello_mem_afu.

PR #239, opened 2018-02-21
FPGABIST: remove dependency on system path

buie1:
Before the bist tool was assuming that all of the opae tools were on the system path after install. This removes that dependency for use in the development and release flows.

### DIFF
--- a/tools/fpgabist/bist_common.py
+++ b/tools/fpgabist/bist_common.py
@@ -78,9 +78,10 @@ def get_mode_from_path(gbs_path):
     return None
 
 
-def load_gbs(gbs_file, bus_num):
+def load_gbs(obj, gbs_file, bus_num):
     print "Attempting Partial Reconfiguration:"
-    cmd = "{} -b 0x{} -v {}".format('fpgaconf', bus_num, gbs_file)
+    conf_cmd = os.path.join(obj.cwd, 'fpgaconf')
+    cmd = "{} -b 0x{} -v {}".format(conf_cmd, bus_num, gbs_file)
     try:
         subprocess.check_call(cmd, shell=True)
     except subprocess.CalledProcessError as e:
@@ -93,6 +94,7 @@ class BistMode(object):
     name = ""
     executables = {}
     dir_path = ""
+    cwd = os.path.dirname(os.path.realpath(__file__))
 
     def run(self, gbs_path, bus_num):
         raise NotImplementedError
--- a/tools/fpgabist/bist_dma.py
+++ b/tools/fpgabist/bist_dma.py
@@ -38,9 +38,10 @@ class DmaMode(bc.BistMode):
         self.executables = {'fpga_dma_test': '0'}
 
     def run(self, gbs_path, bus_num):
-        bc.load_gbs(gbs_path, bus_num)
+        bc.load_gbs(self, gbs_path, bus_num)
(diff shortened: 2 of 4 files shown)


tools/fpgabist/bist_common.py
```diff
@@ -78,9 +78,10 @@ def get_mode_from_path(gbs_path):
     return None
 
 
-def load_gbs(gbs_file, bus_num):
+def load_gbs(obj, gbs_file, bus_num):
```
dasm: Instead of providing `obj` you could just pass `cwd`.

tools/fpgabist/bist_dma.py
```diff
@@ -38,9 +38,10 @@ def __init__(self):
         self.executables = {'fpga_dma_test': '0'}
 
     def run(self, gbs_path, bus_num):
-        bc.load_gbs(gbs_path, bus_num)
+        bc.load_gbs(self, gbs_path, bus_num)
```
dasm: It doesn't need to use `self`. You can just pass `cwd` to function. Or, in your case: `self.cwd`.

dasm:
I understand commit message, but I don't really understand the reason of change.
Could you provide examples why it needs to be changed?

buie1:
@dasm When trying to run the `fpgabist` tool during development some customers are building from source (as opposed to the rpm flow) so if the tools aren't on the system path you wouldn't be able to find other tools like `fpgaconf` or `fpgainfo`, that bist uses.  This change allows you "find" the other necessary tools relative to where the binaries are installed on the system.

tools/fpgabist/bist_common.py
```diff
@@ -93,6 +94,7 @@ class BistMode(object):
     name = ""
     executables = {}
     dir_path = ""
+    cwd = os.path.dirname(os.path.realpath(__file__))
```
southerngs: I think you should rename `cwd`.  It sounds like *current working directory* but in fact it is the path to the script that you are executing, which might not be the actual working directory of the script.  Not sure of the best name, perhaps `opae_tools_path` (although that name is rather long).

tools/fpgabist/bist_dma.py
southerngs: I agree just use `self.cwd`

tools/fpgabist/fpgabist
```diff
@@ -39,6 +39,7 @@ FEATURES = ['fme', 'port', 'temp', 'power', 'errors all']
 
 def main(args):
     bdf = bist_common.get_all_fpga_bdfs()
+    cwd = os.path.dirname(os.path.realpath(__file__))
```
southerngs: again you should renamed `cwd`.  Also it might be better to get this value from bist_common.  That way if it is ever changed you only need to make the update in one place.

buie1:
@dasm @southerngs 
Thanks for your feedback.
Going to close this because relying on the cwd forces the user to ONLY run the fpgabist tool from the desired `bin` directory or somewhere on the system path.  This breaks the workflow of being able to run fpgabist from `opae-sdk/tools/fpgabist` directory.  Instead will create a readme that informs the user to have the `opae_bin_path` as part of their system path.

southerngs:
@buie1 why can't you use fully qualified paths to reference the required files?

PR #242, opened 2018-02-22
Tools: RAS support for discrete FPGA devices.

anandaravuri:
Changes:
 -Add RAS tool to  support discrete FPGA devices


r-rojo:
Is this change the only thing necessary for compatibility with discrete platforms?

tools/ras/main.c
```diff
@@ -1076,7 +1076,8 @@ fpga_result mmio_error(struct RASCommandLine *rasCmdLine)
 		return result;
 	}
 
-	if(value != FPGA_INTEGRATED_DEVICEID) {
+	if( (value != FPGA_INTEGRATED_DEVICEID) ||
```
tswhison: Shouldn't the operator be && ?

r-rojo: Hmm, good catch 👍

PR #243, opened 2018-02-22
Added tools/pylib to provide OPAE version numbers

asgibbons:
Added pylib to hold common python functions for getting CMAKE variables (currently, only version information).


asgibbons:
I don't like what I had to do to get lint to pass.  The packager.py module depends on 'version', which does not exist until cmake has been run.  Anyone have any suggestions?

r-rojo:
Does this do what #234 was set out to do? I get the intent of reuse and putting version in a "library" but I feel like this might be too much trouble for getting the version of the tool in question. I would hold off on making a common library until we finalize our development model for Python code, including 
* How we _build_ our Python tools
* How we distribute our Python code
  * including Python tools and Python bindings around OPAE

In the end, I would like to see Python code to be distributed as standard Python packages (which includes metdata like version and dependency information, ie. _fpga.*_ package depends on _opae_ package).
For now, I see some options for this PR as:
* Change the Python linter script to look at the build directory for generated packages
* Embed the version in the module entry point itself (preferably in the \_\_version\_\_ variable)

asgibbons:
Yes, this is what #234 was trying to do, and it works.  I believe this structure is independent of how the python files are going to be packaged.

michael-adler:
I’m a fan of this library method. Beyond something trivial like versioning, there are common idioms that force us to do CMake variable replacement, such as discovering install paths. There are also modules that ought to be common, like AFU.py. 

I don’t see the point in waiting for the perfect solution when having library infrastructure is a clear step in the right direction. Nothing here will prevent future improvements to release or packaging strategy.

r-rojo:
My concern is not an issue of waiting for the perfect solution but rather a concern about straying too far from industry standards with a complex build structure that may not be necessary. In the meantime, though, why can't we embed the version information in the entry point (similar to #234).

asgibbons:
Can someone approve this and get it merged?

dasm:
I'm with @r-rojo. We shouldn't be distributing our deliverables like this. Python has defined standard of delivering packages: https://packaging.python.org/tutorials/distributing-packages/ and we're diverging from it with every merged commit.
When doing `pip install <package>` wheels allows to introduce dependencies and fetch and compile external repositories, like C/C++: https://github.com/numba/llvmlite/blob/master/setup.py

What we're doing here is only band-aid and not desired solution.

asgibbons:
The whole idea of this was to add a 'version' command to the packager that prints the OPAE API version number, not to re-design the way all python files get built and distributed.  This seems to have ballooned into something it was not intended to solve.

Does this change materially affect any approach as to how packages are to be built?  When wheels are implemented, this pylib merely becomes another package upon which other python files depend.  Or am I missing something?

dasm:
@asgibbons in current shape `version` cmd cannot be used outside of built version of package. We cannot use it, when entering directory and just running `python packager.py`because there won't be `version` package.

It is not ideal, but let's hardcode version number into packager and ship it. I would like to say: "let's merge this, and refactor later" but I know, there is little chance to refactor it, when something is already uploaded.

asgibbons:
Sorry, but I don't follow your logic.  You recommend hardcoding a version number because merging this will require a refactoring later, which you say has a high probability of not happening (the refactoring).  However, if refactoring has a low probability, how will we update the version number in the future?  Hardcode it again?  We've introduced a dependency that can only be satisfied manually by re-hardcoding the version when it changes.

As I tried to say, we can generate a wheel from pylib (in the future) and add it as a dependency to the packager's wheel (and any other python distributable that requires what the library provides).  This package will become a general utility package, and we can add things like afu.py, gbs.py, etc. as needed.  I view this as infinitely better than hardcoding a version number.

dasm:
@asgibbons what I've tried to say:
"We can merge your changes and introduce dependencies on compilations, but it is highly improbable to be refactored and changed in the future, because: `it works, why even change it?`"

Now, when I go to your change, I cannot use packager anymore, because it doesn't allow me to do so. I need to compile it, install it and then find it somewhere else. It breaks current workflow for internal test tools.
```
[dsmigiel@sj-avl-d03-mc packager]$ python packager.py
  File "packager.py", line 35, in <module>
    import version  # pylint: disable=import-error
ImportError: No module named version
```

r-rojo:
As I said earlier I'd eventually like to see structuring our Python code in modular/package formats but that is outside the scope of the pull request. If the purpose of this pull request is to simply version something like packager then I don't see why we can't do it from a template and configure  during configure/build steps (like you had done in #234) instead of adding more complexity to the build. I'm all for code reuse and am well aware of the maintenance problems introduced by code cloning but I would argue that version numbering and reporting is trivial enough that each tool should implement by itself and not have to `import` another module for. That being said, why can't we just add a \_\_version\_\_ attribute to the entry point (even if it is something replaced in a template at configuration/build time)? Also, we might want to include the git hash somewhere in the version number until we make official releases.
We may also want to look at [PEP 386](https://www.python.org/dev/peps/pep-0386/), [PEP 396](https://www.python.org/dev/peps/pep-0396/), and [PEP 440](https://www.python.org/dev/peps/pep-0440/) for guidance. 
PEP 440, for example, spells out how to encode development releases in the version. Perhaps we could use the `dev` separator to add the git hash?

michael-adler:
Would everyone be happy if, for now, we change the current Python zip script to automatically add a __version__ with the OPAE version info to every zipped image?

asgibbons:
Just talked with Shiva.  I'll abandon this pull request and re-submit the first attempt, which adds the version number as part of cmake directly to the packager.py file.  This has much less of an impact on existing process, and we can decide the whole wheel thing later.

PR #244, opened 2018-02-23
remove boost dependency

r-rojo:
* Removes all references to `boost::any` and `boost::lexical_cast` in source
  * Replace `any` with `any_value`
  * Use `std::stringstream` instead of `boost::lexical_cast` where needed
* Remove all references to boost dependencies from CMake and travis


r-rojo:
Good question :) I manually ran the CppOptions unit tests we have. These are also part of the filter that is used when we do `make test` (part of the CI).
image

PR #245, opened 2018-02-23
Add version command by cmake replacement

asgibbons:
The existing packager.py will run as usual.  The only difference will be that when invoked from the source tree (i.e., `python ./tools/packager/packager.py version`) it prints:

`Intel OPAE FPGA Packager: version @INTEL_FPGA_API_VERSION@`

After a cmake, the installed version will print:

`Intel OPAE FPGA Packager: version 0.13.0`

I hope this satisfies all the concerns brought up in the last 2 pull requests.  Once we decide the final solution, we can change this (if necessary).

Thanks.

### DIFF
--- a/tools/packager/packager.py
+++ b/tools/packager/packager.py
@@ -32,9 +32,11 @@ import shutil
 import argparse
 import json
 import utils
+import shlex
 from afu import AFU
 from gbs import GBS
 from metadata import metadata
+from subprocess import Popen, PIPE
 
 PACKAGER_EXEC = "packager"
 DESCRIPTION = 'Intel OPAE FPGA Packager'
@@ -75,7 +77,13 @@ def run_packager():
         print(USAGE)
 
     elif args.cmd == "version":
-        print("{0}: version {1}".format(DESCRIPTION, VERSION))
+        try:
+            cmd = shlex.split("git log -1 --oneline")
+            p = Popen(cmd, stdout=PIPE).communicate()[0]
+            version = p.split(" ")[0] if VERSION.startswith("@") else VERSION
+        except OSError:
+            version = "UNKNOWN"
+        print("{0}: version {1}".format(DESCRIPTION, version))
     elif args.cmd == "create-gbs":
         subparser.usage = "\n" + cmd_description + \
             " --rbf=<RBF_PATH> --afu-json=<AFU_JSON_PATH>"\
--- a/tools/packager/packager.py
+++ b/tools/packager/packager.py
@@ -32,11 +32,9 @@ import shutil
 import argparse
 import json
 import utils
-import shlex
 from afu import AFU
 from gbs import GBS
 from metadata import metadata
-from subprocess import Popen, PIPE
 
 PACKAGER_EXEC = "packager"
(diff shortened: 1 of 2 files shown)


tools/packager/packager.py
```diff
@@ -73,6 +74,8 @@ def run_packager():
     if args.cmd == "help" or not args.cmd:
         print(USAGE)
 
+    elif args.cmd == "version":
+        print("{0}: version {1}".format(DESCRIPTION, VERSION))
```
r-rojo: can we encode the git hash in another attribute and do something like:
```Python
version = GITHASH if VERSION.startswith("@") else VERSION
print("{0}: version {1}".format(DESCRIPTION, version ))
```
This way if you run it from source you'll see the hash instead of the template variable.

tools/packager/packager.py
```diff
@@ -73,6 +76,14 @@ def run_packager():
     if args.cmd == "help" or not args.cmd:
         print(USAGE)
 
+    elif args.cmd == "version":
+        try:
+            cmd = shlex.split("git log -1 --oneline")
+            p = Popen(cmd, stdout=PIPE).communicate()[0]
```
dasm: `subprocess.call` does the same job.

asgibbons: Yes.  subprocess.call is a wrapper around Popen.  Are you requiring this change, or is what I have good enough?

r-rojo: what about something like this:
```Python
subprocess.check_output('git describe --tags', shell=True).split()[0]
``` 
which yields something like:
```
'0.9.0-296-g777469f'
```
Note that `check_output` is a wrapper around Popen too but it's easier if you just want to get the stdout. Also, note that passing in the kwarg `shell=True` removes the requirement of using `shlex`.

dasm: @asgibbons `call` automatically handles kill and wait which gracefully get rid of dead processes. But like @r-rojo pointed out, I've made mistake, it should be `check_output`.

tools/packager/packager.py
```diff
@@ -32,12 +32,15 @@
 import argparse
 import json
 import utils
+import shlex
```
r-rojo: Remove if it's not being used

tools/packager/packager.py
```diff
@@ -32,12 +32,15 @@
...
 from afu import AFU
 from gbs import GBS
 from metadata import metadata
+from subprocess import Popen, PIPE
```
r-rojo: Should probably be `import subprocess` since `check_output` is being called with the `subprocess.` prefix on line 80

tools/packager/packager.py
```diff
@@ -73,6 +76,10 @@ def run_packager():
     if args.cmd == "help" or not args.cmd:
         print(USAGE)
 
+    elif args.cmd == "version":
+        p = subprocess.check_output('git describe --tags', shell=True)
```
r-rojo: What if this code is exported and missing git metadata? Shouldn't we wrap this with a try/except to catch this?

michael-adler:
“git” not “fit”. Autocorrect failure.

PR #247, opened 2018-02-24
add create_python_dist cmake function

r-rojo:
* Move the Python virtualenv setup before processing subdirectories
* Add `create_python_dist` cmake function that can create python
  distribution packages (both source and wheel packages).
  * This function also adds a cmake target to install the package
    into the build tree by either using virtualenv or by installing to
    the user python site-packages (if virtualenv is not found).
    In the end, all entry points should be installed in the build/bin
    directory at build time.
* Port fpgainfo to use create_python_dist

TODO:
* Handle package installation at installation time
* Handle package integration with OS packages (RPM or DEB)


cmake/modules/create_python_dist.cmake
```diff
@@ -0,0 +1,88 @@
...
+## a directory or not. If it is a directory, then it will assume it is a
+## package (and will create an empty __init__.py file in the binary directory
+## if it does not exits). It is up to the developer of the package to ensure
+## that the contents of the directory contain all required modules and the
+## setup.py is correct. It is assumed that the name of the module or package
+## given as the second argument is the name of the package listed in the setup.py
+## file. In either case (if the input is a package or a module), all input files
+## will be processed with 'configure_file' to replace any CMake variables that use
+## the @ symbol
+function(create_python_dist target mod_or_pkg)
```
michael-adler: How about making this function recursive so that it walks a tree instead of just finding Python files in one directory?

r-rojo:
Thanks, @michael-adler, I hadn't thought about how this would look on Windows so I tried some of this out on Windows.
On Windows, I generated a source distribution (using `setup.py sdist`) and then I installed in my local Python 2.7 installation. 
The end result is that it creates an executable (`fpgainfo.exe`) in the `scripts` sub-directory of the Python installation:
image

Now I know that `fpgainfo` is somewhat of a bad example to test on Windows because the current implementation of `fpgainfo` works only on Linux but I'm glad to see that the Python setuptools wraps the entry points in an executable. In this case, it would be up to the user to have the `scripts` directory in the `%PATH%` environment variable.

As for the suggestion for the recursive cmake function, can you give a little more details on what you mean? One way this could be implemented is to walk the directories and process only directories that have a `setup.py` file inside as directories to create Python packages. I think for this to work, however, we would have to not use arguments in the cmake function and instead use Python to process the `setup.py` file to get the package/module name to configure and use that as the target name. Interesting idea.

michael-adler:
Interesting that it generates an executable. Perhaps that's the only way zipped images can work on Windows. Nice that it's automatic.

I haven't really thought through the recursive suggestion. It seemed logical that, given a directory, it could be viewed as a tree instead of just a flat list of contained files. Python supports package hierarchies, after all. I imagine it would just be a recursive implementation of the current function, adding a parameter to indicate the path relative to the root. Right now you look for all files in a directory. If you look for directories within the directory and invoke the function on each subdirectory, you'd have a recursive walk. The same rules of instantiating a setup.py at any level of the hierarchy that has .py files and no setup.py would work.

r-rojo:
Thanks for the input @michael-adler . I have modified it to process a directory recursively to allow for package hierarchies.

cmake/modules/create_python_dist.cmake
```diff
@@ -29,27 +29,29 @@
...
 function(create_python_dist target mod_or_pkg)
     if(IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/${mod_or_pkg})
-        file(GLOB py_src RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}/${mod_or_pkg} ${mod_or_pkg}/*.py)
+        if (NOT EXISTS ${CMAKE_CURRENT_BINARY_DIR}/${mod_or_pkg}/__init__.py)
+            message(SEND_ERROR "${mod_or_pkg} is a directory but missing __init__.py")
+        endif (NOT EXISTS ${CMAKE_CURRENT_BINARY_DIR}/${mod_or_pkg}/__init__.py)
+        file(GLOB_RECURSE py_src
+            RELATIVE
+            ${CMAKE_CURRENT_SOURCE_DIR}/${mod_or_pkg}
+            ${mod_or_pkg}/*.py)
```
michael-adler: I didn't believe the regular expression could actually work (${mod_or_pkg}/*.py). It looked as though it ought to match just top level, but it is what CMake expects. Odd choice on their part!

Can you imagine a case where we want to embed non-Python files from a tree? Perhaps some JSON data file, or other configuration? Should we just include everything in the tree and not limit it to Python files? (Perhaps exclude binaries like .pyc.)

tools/fpgainfo/fpgainfo/__init__.py
```diff
@@ -0,0 +1,25 @@
+# Copyright(c) 2017, Intel Corporation
```
michael-adler: Do we need copyrights on empty files?

buie1:
@r-rojo Are the changes you've made here similar to the changes that all the python changes should implement to work with this new work flow? I want to get address this in the fpgabist for the feature branch this week

tools/fpgainfo/fpgainfo/__init__.py
r-rojo: 🤷‍♂️

cmake/modules/create_python_dist.cmake
r-rojo: I have thought about this some and go back and forth. Let me know what you think of these options:
* Current implementation: specify module (`file.py`) or package (`dir/__init__.py`) and have cmake process the directory and call configure_file on each file in the tree. I did this in case any cmake variables need to be replaced. The reason I went this direction is because each file should have a \_\_version\_\_ attribute that is independent of the package version.
Take a look at [PEP 396](https://www.python.org/dev/peps/pep-0396/). 
Specifically, points 3 and 4 of the specification. Now it doesn't require all modules have a version but that modules inside a namespace package should and that they should be in the \_\_version\_\_ attribute. 
* Copy the package directory as is. No cmake variable replacement will be done with the contents. The only file that we call `configure_file` on is `setup.py` as that file should contain the version number
* Use explicit filenames in CMakeLists.txt. I tend to prefer explicit but in this case I think we should treat a package directory as the input leave it up to the Python package developer to ensure contents are valid.
* Process setup.py to get a list of collateral (like data files). We may have to use the setuptools api for this but this may require more work.

r-rojo:
@buie1 The main change here is the `create_python_dist` cmake function. The intent here is that other Python tools can use it to create a Python distribution as an output of the build. Right now, it creates two distributions using Python setuptools:
`sdist` : for a source distribution. This is really just a tar.gz file containing all the source and package metadata that pip can use to install
`bdist_wheel` : for a universal binary distribution. I believe the format of this file is a zip format and includes the metadata. 

This `create_python_dist` function sets up the cmake target to create both formats and then will either:
1. Install the `sdist` package into the virtual Python environment created with `virtualenv` which happens to be the root of the cmake binary tree. This is done if the cmake option to use this is ON and `virtualenv` is found in the Python installation. The nice thing about this is that this virtual environment can be thrown away with no harm to the rest of the Python installation.
2. Install the `sdist` package into the user site packages (`~/.local/lib/python(2.7|3.?)/site-packages`) but place entry points in the executable directory (`build/bin`). This is done if `virtualenv` isn't found or being used. The package modules will end up in the site packages area so it is up to the user to uninstall the package if needed.

The net effect is that all entry points listed in the setup.py file will have a wrapper script (or exe in Windows) generated in `build/bin` after doing a `make`.

Also, in both cases, the logic is setup so that the package is always installed, even if pip detects it already being installed.

PR #248, opened 2018-02-26
modify FindPythonPackage to use Python found

r-rojo:
Add find_package(PythonInter...) to the beginning of the file
(outside of the macro) so that cmake finds the Python interpreter
and uses it when checking for existence of a module.

### DIFF
--- a/cmake/modules/FindPythonPackage.cmake
+++ b/cmake/modules/FindPythonPackage.cmake
@@ -1,13 +1,13 @@
 # - Find python package installation status
 #
-# ERR_LEVEL: 
-#     SEND_ERROR     = CMake Error, continue processing 
+# ERR_LEVEL:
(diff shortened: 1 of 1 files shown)


dasm:
@r-rojo nice change, although I wanted to ask about something.
According to my knowledge, it doesn't change anything. Is it required for virtualenv to work? Would it be worth to add check for Python2.7? At least one of tools is pinpointed to this version: https://github.com/OPAE/opae-sdk/blob/master/tools/packager/packager.py#L43

(In the future, I would like to port it over to py3)

r-rojo:
@dasm virtualenv should work with both Python 2.7 and 3 although it might be a good idea to require 2.7 in our build system since I'm not aware of any requirements within our codebase for Python 3.

PR #249, opened 2018-02-26
CI: Automate release-tagged documentation upload process

nakulkorde:
Enable automatic uploading of latest built documentation whenever there is a new release tag. This change builds the documentation and checks if the latest commit is associated with a tag; if true uploads the documentation to a new folder in opae.github.io so that it is accessible as : opae.github.io/NEW_TAG_NAME.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,10 @@ matrix:
         - doxygen
         - python-jsonschema
         - pandoc
+    before install:
+      - wget https://github.com/jgm/pandoc/releases/download/2.1.1/pandoc-2.1.1-1-amd64.deb
+      - sudo dpkg -i pandoc-2.1.1-1-amd64.deb
+      
 env:
   global:
     secure: LciVXkQV5xhP1xHa7Z+ZLaAdS164Lf3n5/O4DswLQbpsmE7IdZAxmPGfzUomi4JlPXfrTFdXjo3Z/VH+xwSLiBgfBQES5f6dN/Dkw+gs+10HMc4TZp3w1IfTE6/QADh/EgfoqwIQOSYPv4XRV8z0+cEnbWWefiq3eLRAYTNw9YwORD8jE8gAQ5UfnVLWzA+uHUnZDXCCJO0HU4QKVaDwlcf6T2/GNsbIvQgowyy4XKcV3EWeGBDAoVY3hZ3l0UMs+RBg3cps1E3KU+4Uf9UokdZDWy5QZWHL4BYM5KCN9PXeVcliYjpBXgfCI7cAcju7xxJ4TJYOe+syYEsDFNrPOV+gFpCqDPwRtC1kvT8hslabP+WrBgDSuR/zhfypMJ5/5/Ltaj0GcdeB25PvZTtG0rrt8KwGA9pUJxo43ucWtfThb7Og5dI2ChbllifwnSiRPnE7Oi9RQq4vKiDZfph2tJbXjFL+XbBh8TxoVwYbnLbLWw6ruRxqSw4cbKs2f4F0THRxu8THzZ2x1pQI0x1k6ERYns1eVlg/9PiiUlZ4n1XZx9i98urAnc+TVVGskpYbIs5azhsb0r4n1bfIrUhbsFHiJfH2rI2Ocs96KUPH60dFI+Z8DQFZDohLi2trNAALvqtxnOAMSScvFS4YpqAjWTTlLjswQZQddOan8Z5B7YE=
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,3 +29,14 @@ matrix:
             - uuid-dev
             - doxygen
             - python-jsonschema
+    - script: scripts/build-documentation.sh
+      cache: ccache
+      addons:
+        apt:
+          packages:
+            - libboost-all-dev
+            - libjson-c-dev
+            - uuid-dev
+            - doxygen
+            - python-jsonschema
+            - pandoc
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,41 +2,44 @@ language: c
 compiler: gcc
 matrix:
   include:
-    - script: scripts/test-codingstyle.sh
-    - script: scripts/test-codingstyle-py.sh
-      cache: pip
-      install:
-        - pip install --user pylint
-        - pip install --user pycodestyle
-        - pip install --user jsonschema
-    - script: scripts/test-build.sh
-      cache: ccache
-      addons:
-        apt:
-          packages:
-            - libboost-all-dev
-            - libjson-c-dev
-            - uuid-dev
-            - doxygen
-            - python-jsonschema
-    - script: scripts/test-gtapi-mock-drv.sh
-      cache: ccache
-      addons:
-        apt:
-          packages:
-            - libboost-all-dev
-            - libjson-c-dev
-            - uuid-dev
-            - doxygen
-            - python-jsonschema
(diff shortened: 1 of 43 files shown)


scripts/build-documentation.sh
```diff
@@ -12,3 +13,14 @@ make
...
+
+COMMIT_ID="$(git rev-parse HEAD)"
+TEMP=$((git describe --exact-match $COMMIT_ID) 2>&1)
+
+if [[ $TEMP != *"fatal"* ]]; then
+    echo "New tag detected; Uploading the documentation to opae.github.io"
+    /bin/bash scripts/push-documentation.sh 
+else
+    echo "No new tag detected; Documentation will not be uploaded to opae.github.io"
+fi
```
r-rojo: I think travis has an environment variables for this type of information.
https://docs.travis-ci.com/user/environment-variables/
image
In particular, can you use the `TRAVIS_TAG` environment variable to determine if this is a tag?

scripts/push-documentation.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
+setup_git() {
+  git config --global user.email "nakul.korde@intel.com"
+  git config --global user.name "Nakul Korde"
+}
```
r-rojo: Should we set up a `travis` user? Just a thought...

scripts/push-documentation.sh
```diff
@@ -0,0 +1,27 @@
...
+
+commit_website_files() {
+  mkdir upload_docs
+  cd upload_docs
+  git clone https://github.com/OPAE/opae.github.io.git
+  cd doc-testing
+  cp -r ../../mybuild_docs/sphinx/html/* .
+  git add -A
+  git commit --message "Travis build: Update Documentation with traceback: second test commit $TRAVIS_BUILD_NUMBER"
```
r-rojo: traceback in some languages refers to a stack trace. This message might confuse people (like me).

scripts/build-documentation.sh
nakulkorde: Needs some investigation to be done in order to figure out the usage. Can use git commands for now.

scripts/push-documentation.sh
nakulkorde: That is good idea although the credentials are encrypted, might be better to have a dedicated travis account for future use; will be beneficial for management.

scripts/push-documentation.sh
nakulkorde: Thanks for pointing out; rectified in new commit.

scripts/build-documentation.sh
luebbers: Have you tested this? How?

.travis.yml
```diff
@@ -2,28 +2,44 @@ language: c
...
-      cache: ccache
-      addons:
-        apt:
-          packages:
-            - libjson-c-dev
-            - uuid-dev
-            - doxygen
-            - python-jsonschema
+  - script: scripts/test-codingstyle.sh
+  - script: scripts/test-codingstyle-py.sh
```
luebbers: Is there a difference to grouping the `script` items under `include:` vs. a level higher? Not sure what `include:` does.

After this change, do we still nee the `include:` line (4)?

scripts/push-documentation.sh
```diff
@@ -0,0 +1,27 @@
...
+setup_git() {
+  git config --global user.email "nakul.korde@intel.com"
+  git config --global user.name "Nakul Korde"
+}
+
+commit_website_files() {
+  mkdir upload_docs
+  cd upload_docs
+  git clone https://github.com/OPAE/opae.github.io.git
+  cd doc-testing
```
luebbers: Not sure how this works - doesn't the `git clone` in line 11 create a directory called "opae.github.io"? Where does `doc-testing` come from?

scripts/push-documentation.sh
luebbers: Will you change the commit message later (this looks like a test)?

scripts/push-documentation.sh
luebbers: How and where is the directory that's named after the tag created?

.travis.yml
r-rojo: LInes 5 is one matrix/job configuration, line 6 to 11 is another.
The real change for the matrix configurations is the addition of the documentation job (lines 32-42).
I added to `include` key originally to explicitly group job configurations (instead of mixing/combining different matrix job settings).

.travis.yml
```diff
@@ -2,28 +2,44 @@ language: c
...
+        packages:
+        - libboost-all-dev
+        - libjson-c-dev
+        - uuid-dev
+        - doxygen
+        - python-jsonschema
+        - pandoc
+env:
+  global:
+    secure: LciVXkQV5xhP1xHa7Z+ZLaAdS164Lf3n5/O4DswLQbpsmE7IdZAxmPGfzUomi4JlPXfrTFdXjo3Z/VH+xwSLiBgfBQES5f6dN/Dkw+gs+10HMc4TZp3w1IfTE6/QADh/EgfoqwIQOSYPv4XRV8z0+cEnbWWefiq3eLRAYTNw9YwORD8jE8gAQ5UfnVLWzA+uHUnZDXCCJO0HU4QKVaDwlcf6T2/GNsbIvQgowyy4XKcV3EWeGBDAoVY3hZ3l0UMs+RBg3cps1E3KU+4Uf9UokdZDWy5QZWHL4BYM5KCN9PXeVcliYjpBXgfCI7cAcju7xxJ4TJYOe+syYEsDFNrPOV+gFpCqDPwRtC1kvT8hslabP+WrBgDSuR/zhfypMJ5/5/Ltaj0GcdeB25PvZTtG0rrt8KwGA9pUJxo43ucWtfThb7Og5dI2ChbllifwnSiRPnE7Oi9RQq4vKiDZfph2tJbXjFL+XbBh8TxoVwYbnLbLWw6ruRxqSw4cbKs2f4F0THRxu8THzZ2x1pQI0x1k6ERYns1eVlg/9PiiUlZ4n1XZx9i98urAnc+TVVGskpYbIs5azhsb0r4n1bfIrUhbsFHiJfH2rI2Ocs96KUPH60dFI+Z8DQFZDohLi2trNAALvqtxnOAMSScvFS4YpqAjWTTlLjswQZQddOan8Z5B7YE=
```
abelardojarab: travis allows you to define private variables. this value should not be placed here, it should not be printed too.

.travis.yml
nakulkorde: @luebbers I tested it manually by running with tagged commits as well as untagged commits in order to verify whether the proper conditions are checked.

scripts/push-documentation.sh
nakulkorde: Yes you are right. It should create a directory opae.github.io and not doc-testing. It was a part of the testing I was doing and pushed the wrong version of the script. I have corrected the error now.

When the documentation is built, it would create a directory for the html files named with the current release in the location as can be seen from here: mybuild_docs/sphinx/html/

https://github.com/OPAE/opae-sdk/blob/55d987f5863ceb3e7f781ddb306409fa190ffe48/doc/sphinx/CMakeLists.txt#L42-L43

I then copy that directory to the where I have cloned the opae.github.io repo and then push it to opae.github.io.

.travis.yml
nakulkorde: @r-rojo you are right, these variables can only be decrypted inside the Travis environment when a build is triggered. I will be using bot credentials in the future developments with Travis.

PR #240, opened 2018-02-22
Adding bist afu support to the fpgabist tool

buie1:
The fpgabist tool will now be able to use the bist_afu as one of the modes.  

There will be additional development on the application to remove warnings.

PR #250, opened 2018-02-28
Avoid clobbering CMAKE_C_FLAGS for bist tool

chipb:


chipb:
I'm not very familiar with CMake, but we almost certainly don't want to completely override compile flags here.

Compare compile command before:
```sh
/tools/gcc/4.8.5/1/linux64/bin/gcc -std=c99 -O2 -Wall -Wextra -Werror -I/build2/qshell_adapt/git/opae_sdk_x/common/include -I/build2/qshell_adapt/git/opae_sdk_x/libopae/src -I/build2/qshell_adapt/work/sw/opae_sdk_x/libopae/include -o CMakeFiles/bist_app.dir/bist_afu.c.o -c /build2/qshell_adapt/git/opae_sdk_x/tools/fpgabist/bist/bist_afu.c
```
...to after:
```sh
/tools/gcc/4.8.5/1/linux64/bin/gcc -Wformat -Wformat-security -fPIC -D_FORTIFY_SOURCE=2 -fstack-protector-all -std=c99 -O2 -Wall -Wextra -Werror -I/build2/qshell_adapt/git/opae_sdk_x/common/include -I/build2/qshell_adapt/git/opae_sdk_x/libopae/src -I/build2/qshell_adapt/work/sw/opae_sdk_x/libopae/include -o CMakeFiles/bist_app.dir/bist_afu.c.o -c /build2/qshell_adapt/git/opae_sdk_x/tools/fpgabist/bist/bist_afu.c
```

r-rojo:
Is this before/after comparison of flags that this pull request implements?
Or is it a comparison of flags before and after #240?

chipb:
It was a comparison before/after this proposed #250 in my groups nightly integration build environment. I would not be surprised if our special snowflake environment is the only place to easily reproduce the issue, but I would think it just comes down to whether or not the CC defaults to enabling `-fPIC`.

I did a build check and inspected the different output for the ${CMAKE_BINARY_DIR}/compile_commands.json file. Only that specific compilation unit (i.e. `tools/fpgabist/bist/bist_afu.c`) was affected by the change as the override was local to `tools/fpgabist/bist/CMakeLists.txt`.

For completeness, this is the error we saw in our nightly build:
```
/tools/gcc/4.8.5/1/linux64/lib/gcc/x86_64-unknown-linux-gnu/4.8.5/../../../../x86_64-unknown-linux-gnu/bin/ld: CMakeFiles/bist_app.dir/bist_afu.c.o: relocation R_X86_64_32 against `.[ 77%] rodata.str1.1' can not be used when making
 a shared object; recompile with -fPIC
CMakeFiles/bist_app.dir/bist_afu.c.o: could not read symbols: Bad value
```
...which makes sense as the override had the unintended effect of dropping `-Wformat -Wformat-security -fPIC -D_FORTIFY_SOURCE=2 -fstack-protector-all` in this environment.

chipb:
I'm open to other solutions, as well.

Perhaps we should be setting the default C standard globally for the project? I see at least `tools/ras/CMakeLists.txt` and `samples/CMakeLists.txt` use `-std=gnu99` I think would be sufficient for @buie1's purposes.

r-rojo:
The root CMakeLists.txt includes comiler_config.cmake:
https://github.com/OPAE/opae-sdk/blob/ef3d15d9572c3e2e1ee7adadff1424c1999bbd42/CMakeLists.txt#L61-L63

which sets the default flags for the project:
https://github.com/OPAE/opae-sdk/blob/ef3d15d9572c3e2e1ee7adadff1424c1999bbd42/cmake/modules/compiler_config.cmake#L54-L64

I think setting compiler flags in one directory only applies to all targets in that directory (and shouldn't affect the rest of the build).

PR #254, opened 2018-02-28
fpgainfo: allow executing namespace package

r-rojo:
* rename `fpgainfo.py` to `__main__.py` to make the namespace package
  executable
* import main into root `__init__.py` file
* change stup.py to reference `fpgainfo:main` for the entry point


tools/fpgainfo/setup.py
```diff
@@ -5,7 +5,7 @@
     packages=find_packages(),
     entry_points={
         'console_scripts': [
-            'fpgainfo = fpgainfo.fpgainfo:main',
+            'fpgainfo = fpgainfo:main',
```
dasm: 👍

PR #259, opened 2018-03-02
OPAE: fix gcc 7.3.1 build warnings

tswhison:
Fix a handful of warnings that were exposed by -Werror -Wall when using
gcc 7.3.1 in Fedora 27. Add safestring dependency to hello_events and
scrub snprintf.


libopae/src/enum.c
```diff
@@ -338,10 +338,10 @@ bool del_dev(struct dev_list *pdev, struct dev_list *parent)
 	return true;
 }
 
-static const fpga_guid FPGA_FME_GUID = {
-	0xbf, 0xaf, 0x2a, 0xe9, 0x4a, 0x52, 0x46, 0xe3,
-	0x82, 0xfe, 0x38, 0xf0, 0xf9, 0xe1, 0x77, 0x64
-};
+//static const fpga_guid FPGA_FME_GUID = {
```
southerngs: What is the warning that commenting out this code fixes?  Why not just delete this code?  If there is a reason to keep it then please add a comment explaining why.

tswhison: @southerngs see line 381 in the same file.

safe_string/CMakeLists.txt
```diff
@@ -113,6 +113,9 @@ set_property(TARGET safestr PROPERTY POSITION_INDEPENDENT_CODE ON)
 target_compile_definitions(safestr PRIVATE PIC=1)
 target_compile_options(safestr PRIVATE -O3)
 
+# mem_primitives_lib.c is optimized through the use of fallthrough switch statements.
+set_source_files_properties(mem_primitives_lib.c PROPERTIES COMPILE_FLAGS -Wno-implicit-fallthrough)
```
luebbers: Nice and elegant.

libopae/src/event.c
```diff
@@ -352,6 +352,8 @@ static fpga_result driver_register_event(fpga_handle handle,
 		} else if (objtype == FPGA_ACCELERATOR) {
 			return send_port_event_request(handle, event_handle, FPGA_IRQ_ASSIGN);
 		}
+		FPGA_ERR("Invalid objtype: %d", objtype);
+		return FPGA_EXCEPTION;
```
lzhan55: @tswhison should we try to avoid multiple return statements in a function by setting the "res = FPGA_EXCEPTION in default case and returning the 'res' variable at the end of this function?

tswhison: Not sure that would address the warning here.  The warning was that the switch case was implicitly fall-through.  So we needed either a break or return to satisfy the compiler.

lzhan55: @tswhison This is not related to the warning here. Let's discuss the coding style in the future.

PR #261, opened 2018-03-07
c++core: rename read/write to read_csr/write_csr

r-rojo:
* Rename csr read/write functions in `handle` class to read_csr<T> and
write_csr<T> to be more explicit in the operation and specify that the
functions are operating on CSRs. This is a template function where the
template argument specifies the register size (uint32_t or uint64_t for
32-bit or 64-bit respectively).
* Update samples to use updated API functions.

### DIFF
--- a/common/include/opae/cxx/core/handle.h
+++ b/common/include/opae/cxx/core/handle.h
@@ -60,33 +60,44 @@ class handle {
    */
   operator fpga_handle() const { return handle_; }
 
-  /** Write 32-bit value to MMIO.
-   * @param[in] offset The byte offset from MMIO base to write.
-   * @param[in] value  The value to be written.
-   * @return Whether the write was successful.
+  /**
+   * @brief Read CSR from resource associated with the handle
+   *
+   * @tparam T The type of value to return.
+   *
+   * @note Only uint32_t or uint64_t are allowed.
+   *
+   * @param offset The register offset
+   * @param csr_space The CSR space to read from. Default is 0.
+   *
+   * @return The value of type T read from the CSR
    */
-  virtual bool write(uint64_t offset, uint32_t value);
+  template<typename T>
+  T read_csr(uint64_t offset, uint32_t csr_space = 0) const {
+    (void)offset;
+    (void)csr_space;
+    throw except(OPAECXX_HERE);
+  }
 
-  /** Write 64-bit value to MMIO.
-   * @param[in] offset The byte offset from MMIO base to write.
-   * @param[in] value  The value to be written.
-   * @return Whether the write was successful.
+  /**
+   * @brief Write value to CSR from resource associated with the handle
+   *
+   * @tparam T The type of value to write.
+   *
+   * @note Only uint32_t or uint64_t are allowed.
+   *
+   * @param offset The register offset.
+   * @param value The value to write to the register.
+   * @param csr_space The CSR space to read from. Default is 0.
+   *
    */
-  virtual bool write(uint64_t offset, uint64_t value);
-
-  /** Read 32-bit value from MMIO.
-   * @param[in]  offset The byte offset from MMIO base to read.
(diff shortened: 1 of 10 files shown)


common/include/opae/cxx/core/handle.h
```diff
@@ -60,33 +60,44 @@ class handle {
...
+   *
+   * @return The value of type T read from the CSR
    */
-  virtual bool read(uint64_t offset, uint32_t &value) const;
-
-  /** Read 64-bit value from MMIO.
-   * @param[in]  offset The byte offset from MMIO base to read.
-   * @param[out] value  Receives the value read.
-   * @return Whether the read was successful.
+  template<typename T>
```
tswhison: I follow that we want the generic case to be prevented, but why don't we use = delete?
Also, don't we need the signatures of the allowed fn's (uint32_t, uint64_t) to be declared here so that clients can see them?  I might be missing something here..

r-rojo:
I'm actually a big fan of generic programming using templates and have had over ten years experience developing software with C++. Now in this case, the use of templates doesn't really accomplish any sort of generic programming (as you can see with the default implementation of the functions). 
When @luebbers, @tswhison, and I discussed this, we considered first using function names like you have suggested. In the end, however, we made the decision to use template arguments to encode the type of the variable  (which infers the register size).

common/include/opae/cxx/core/handle.h
r-rojo: As a templated fn we can't delete it otherwise it won't compile. The signatures for uint32_t/uint64_t are template specializations and can go in the cpp. The only reason I put them in the cpp is because the mmio.h header is included in .cpp file and not the .h. I could've just as easily included mmio.h in this header and implemented the specialization there too.
You did make me think of something though. The way it's currently implemented, throwing an exception is a runtime error and not a compile time error. I think we would actually prefer to have a compile time error here which would require something like throwing in a static_assert on the template argument. Given this, and the points raised by @michael-adler, I'm starting to wonder if it's more trouble than what it's worth...

tswhison: I thought you would be able to explicitly instantiate the two forms you needed and then = delete the rest.  No matter if we don't want templates after all.

michael-adler: To me it seems like more trouble than it's worth. Having the template will annoy the hardware people (who have to type more C++ syntax to specify what they want) and give the software people a false sense of security that the system will infer correctly what it should be doing when generating MMIO traffic. Using explicit names would keep the hardware people happy and keep the software folks from getting into trouble they don't understand.

r-rojo: Yeah, I'll probably change it to read_csr(32|64) and write_csr(32|64).  As I said before, we're not really accomplishing any sort of generic programming. Thanks for your input 😄

luebbers:
I tend to agree with Michael - templates are useful constructs for places where we expect new (and possibly yet unknown types) to be used; I understand that they would be used to work with *any* type supporting operations that the templated functions perform.

Here, we only care about different *sizes* of the argument (we don't really care about or interpret their *type* at all), so it appears that templates don't add any real value, especially if the underlying hardware limits the selection of valid sizes (32 bit and 64 bit for now).

libopae++/src/handle.cpp
```diff
@@ -125,6 +109,33 @@ void handle::reset() {
   }
 }
 
+uint32_t handle::read_csr32(uint64_t offset, uint32_t csr_space) const {
+  uint32_t value = 0;
+  if (FPGA_OK == fpgaReadMMIO32(handle_, csr_space, offset, &value)){
```
tswhison: <nit>Do you need a space between )) and {? (applies to the four new fn's)</nit>

r-rojo: Probably, but I plan to do a clang-format on this soon so this will get fixed then.

PR #262, opened 2018-03-08
Move pClk frequency setting to the clocks class.

michael-adler:
It was configured in the wrong place (the CCI-P interface class).


michael-adler:
DCP tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2237248

PR #263, opened 2018-03-08
opaecxx: refactor MMIO tracking in handle

tswhison:
Remove the csr_space parameter from the handle factory method and from
handle constructors. Track each mapped MMIO space internally as it is
mapped either by the CSR accessors or by mmio_ptr().

### DIFF
--- a/common/include/opae/cxx/core/handle.h
+++ b/common/include/opae/cxx/core/handle.h
@@ -24,7 +24,7 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 #pragma once
-#include <vector>
+#include <map>
 #include <memory>
 
 #include <opae/types.h>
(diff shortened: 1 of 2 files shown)


michael-adler:
What's the value of calling track_mmio_space() on every CSR read/write call? Isn't that just extra overhead? If the user really needs the mapped pointer, can't they get it with an explicit call to track_mmio_space()?

Given that, should it really be called track_mmio_space()? Isn't it more like a map or get? The fact that you're tracking it is an implementation/optimization detail, right?

tswhison:
The thing is that each of the MMIO accessors results in a lazy mapping of the MMIO space, but fpgaClose doesn't unmap them.  So, we have to track internally what gets mapped so that we don't leak the regions in the d'tor.

michael-adler:
Should we fix fpgaClose instead? That seems like a bug, especially since the mapping is a side effect in C mode, too.

tswhison:
@luebbers could you comment on why we don't automatically munmap() the mapped MMIO regions in wsid_cleanup()?  That's how it reads to me, but maybe I've missed something.

luebbers:
That looks like an oversight. When people had to explicitly MapMMIO() their MMIO regions, it made sense to rely on them also UnmapMMIO() them. We changed that to allow lazy/automatic mapping when MMIO was accessed, so we should add the unmapping to fpgaClose().

tswhison:
Need to add MMIO region freeing code to fpgaClose, then re-design this change.

PR #266, opened 2018-03-09
libopae: free MMIO mappings on close

tswhison:
Parameterize wsid_cleanup() with an optional callback so that work can
be done on each mapping when the list is being destroyed. Add call to
destroy handle::mmio_root in fpgaClose().

### DIFF
--- a/libopae/src/wsid_list.c
+++ b/libopae/src/wsid_list.c
@@ -116,7 +116,7 @@ bool wsid_del(struct wsid_map **root, uint64_t wsid)
  *
  * @param root
  */
-void wsid_cleanup(struct wsid_map **root)
+void wsid_cleanup(struct wsid_map **root, void (*clean)(struct wsid_map * ))
 {
 	if (!*root)
 		return;
@@ -124,9 +124,13 @@ void wsid_cleanup(struct wsid_map **root)
 	while ((*root)->next) {
 		struct wsid_map *tmp = *root;
 		*root = (*root)->next;
+		if (clean)
+			clean(tmp);
 		free(tmp);
 	}
 
+	if (clean)
+		clean(*root);
 	free(*root);
 	*root = NULL;
 }
--- a/libopae/src/wsid_list.c
+++ b/libopae/src/wsid_list.c
@@ -116,7 +116,7 @@ bool wsid_del(struct wsid_map **root, uint64_t wsid)
  *
  * @param root
  */
-void wsid_cleanup(struct wsid_map **root, void (*clean)(struct wsid_map * ))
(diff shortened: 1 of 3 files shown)


tswhison:
@shivarao please check to see if you guys want this in 0.14.

shivarao:
@tswhison we're not putting further changes on 0.14 except if they help with some critical issues we have.

Is this a critical bug fix? What's the impact of not having this? Memory leak?

libopae/src/wsid_list.c
```diff
@@ -116,17 +116,21 @@ bool wsid_del(struct wsid_map **root, uint64_t wsid)
...
-void wsid_cleanup(struct wsid_map **root)
+void wsid_cleanup(struct wsid_map **root, void (*clean)(struct wsid_map * ))
 {
 	if (!*root)
 		return;
 
 	while ((*root)->next) {
 		struct wsid_map *tmp = *root;
 		*root = (*root)->next;
+		if (clean)
```
michael-adler: Aren't you asking for trouble using multi-line without brackets?

tswhison:
@shivarao Yes, leaked resources and DOS for subsequent processes attempting to obtain the same resource.

libopae/src/wsid_list.c
tswhison: I guess I'm not seeing it.  Isn't clean(tmp); a single statement?

shivarao:
@tswhison  this sounds like a good bug fix but also like an OK bug for our 1.1 Alpha - especially considering no one has reported an issue here as far as we know.

I say leave it off release/0.14 and we'll pick it up when we branch for 1.1 Beta. Let me know if you think my assessment is incorrect

michael-adler:
I think that's a fine choice. It's only a meaningful leak if SW closes an FPGA connection without exiting, which is relatively rare.

rpSTTD:
Is there any existing test test this new logic?

tswhison:
@rpSTTD So, to test this, I commented out the call to fpgaUnmapMMIO() at the end of hello_fpga and stepped through the fpgaClose() call in the debugger, verifying that the callback was invoked to unmap the MMIO region.
As far as introducing an automated test, no.  I can't imagine what that might look like without muddying up the code with some validation hooks.  Do you have any ideas how to test it sans hooks?

tswhison:
@luebbers what do you think about adding a debug assert that handle::mmio_root is NULL at the end of fpgaClose()?  Also, should we apply the corresponding buffer unmap/free logic to a non-empty handle::wsid_root in fpgaClose()?

rpSTTD:
The manual test works for checkin the patch.   https://jira01.devtools.intel.com/browse/OPAE-691 is opened to track automated test development for this patch.

PR #270, opened 2018-03-13
Enable additional common tests

rpSTTD:
Enable additional common tests.


rpSTTD:
PR272 is duplicate with this PR.

PR #272, opened 2018-03-13
Fixed test_open_common_mock_drv test target.

ibarraza:
The previous gtest filter defined in tests/CMakeLists.txt caused the filter to be empty so none of the tests would actually be run for the target test_open_common_mock_drv. When they were actually running, four of the tests would fail for that target. This fixes the filter, disables reconfiguration so LibopaecOpenFCommonMOCK.02 can pass, and disables the three tests (LibopaecOpenFCommonMOCK.03, LibopaecOpenFCommonMOCKHW.06, LibopaecOpenFCommonMOCK.open_drv_09) that do not work with the mock driver.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -70,7 +70,7 @@ endif()
 ############################################################################
 add_test(NAME test_open_common_mock_drv
   WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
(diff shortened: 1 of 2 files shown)


tswhison:
Per 'ctest -V', there are roughly 12 failed test cases.  Could you have a look?

ibarraza:
Now that we have test output on failure, it looks like the failing test is LibopaecUmsgCommonMOCK.Umsg_drv_05, which is returning FPGA_NO_MEMORY. I guess hugepages are required.

rpSTTD:
Please rename this test to LibopaecUmsgCommonHW.Umsg_drv_05.

PR #274, opened 2018-03-15
fpgadiag: fix "freq" option for bw calc.

r-rojo:
Set the `frequency_` member variable for nlb0, nlb3, and nlb7 to the
value in the `freq` option. Without this, the frequency used for
bandwitdth calculations was always 400 MHz (ignoring user input).

### DIFF
--- a/tools/fpgadiag/nlb0.cpp
+++ b/tools/fpgadiag/nlb0.cpp
@@ -300,6 +300,8 @@ bool nlb0::setup()
             cont_timeout_ += hours(timeout_hour);
         }
     }
+
+    options_.get_value<uint32_t>("freq", frequency_);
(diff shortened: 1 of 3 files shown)


southerngs:
Seems like a reasonable bugfix.  

Is there a reason we specify the frequency in hertz in the command line option.  Could we make it MHz instead?

Also we probably should still find a way to automatically detect the frequency in the long term rather than making the user specify it.  But the long term fix shouldn't block this from going in.

r-rojo:
I honestly don't know (or remember) why the frequency is in Hz rather than MHz. I agree, it should probably be in MHz but auto-detecting the frequency would be better. We will definitely keep this when we update the tools.

PR #276, opened 2018-03-16
Generate CCI-P almost full signals on PCIe-only systems.

michael-adler:
- Always use outoforder_wrf_channel.
- Tweak buffer sizes to reduce the number of buffered requests, more like HW.
- Add support for interrupt requests through outoforder_wrf_channel.
- Remove Verilog preprocessor macros for read vs. write in outoforder_wrf_channel.
  They were always defined, so weren't controlling anything.

### DIFF
--- a/ase/rtl/platform.vh
+++ b/ase/rtl/platform.vh
@@ -124,34 +124,32 @@
 `define LAT_UNDEFINED              300
 
 `define RDWR_VL_LATRANGE           20,118
-`define RDWR_VH_LATRANGE           240,270
+`define RDWR_VH_LATRANGE           140,180
 
 `define ASE_MAX_LATENCY            300
 
 
 /*
- * Disable Transaction shuffling
- * - In integrated mode, OutOfOrder channel control is used
- * - In discrete mode, InOrder channel control is used
+ * Select transaction shuffling mode.
  *
  * Infinite bandwidth test mode (independent of platform selection)
  * `define INFINITE_BANDWIDTH_MODE
- *
- * Specific platform features like interrupts and UMsgs are explicitly
+ */
+`ifdef INFINITE_BANDWIDTH_MODE
+ `define FORWARDING_CHANNEL            inorder_wrf_channel
+`else
+ `define FORWARDING_CHANNEL            outoforder_wrf_channel
+`endif
+
+/* Specific platform features like interrupts and UMsgs are explicitly
  * selected based on platform type
  */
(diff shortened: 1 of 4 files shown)


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2258819

ase/rtl/platform.vh
```diff
@@ -124,34 +124,32 @@
 `define LAT_UNDEFINED              300
 
 `define RDWR_VL_LATRANGE           20,118
-`define RDWR_VH_LATRANGE           240,270
+`define RDWR_VH_LATRANGE           140,180
```
deepakcu84: how are these values chosen?

ase/rtl/platform.vh
```diff
@@ -124,34 +124,32 @@
...
- * - In discrete mode, InOrder channel control is used
+ * Select transaction shuffling mode.
  *
  * Infinite bandwidth test mode (independent of platform selection)
  * `define INFINITE_BANDWIDTH_MODE
- *
- * Specific platform features like interrupts and UMsgs are explicitly
+ */
+`ifdef INFINITE_BANDWIDTH_MODE
+ `define FORWARDING_CHANNEL            inorder_wrf_channel
```
deepakcu84: Is there a reason to keep inorder_wrf_channel and INFINITE_BANDWIDTH mode?

ase/rtl/platform.vh
michael-adler: A balance of accuracy vs. simulator speed. The right latency relative to the size of the simulated pipeline gives a reasonable number of requests in flight.

ase/rtl/platform.vh
michael-adler: Yes! If you are debugging looking for pipeline hazards in an AFU it is sometimes valuable to have the simulated FIU offer infinite bandwidth, leaving any hazards in the AFU exposed. All pipeline bubbles in infinite bandwidth mode are caused by the AFU.

michael-adler:
nlb triggers almost full in its current state.

michael-adler:
@abelardojarab or @lzhan55 -- waiting on approval for merge.

lzhan55:
Why don't use inorder_wrf_channel for DCP? Is there hardware change for it?

michael-adler:
Because it doesn't have a latency model and gives a false sense of security that all transactions will be ordered, which we don't want to guarantee for future physical channels.

PR #277, opened 2018-03-16
opaecxx: add exception for all the result codes

r-rojo:
* Create an exception class for each result code (other than `FPGA_OK`) in
  OPAE C API
* Create an errors.h file with
  * Macros to simplify asserting result codes (specifically `ASSERT_FPGA_OK`)
  * A table of result codes to expressions that throw an exception if
    the result code is not `FPGA_OK`
* Update one file (`handle.cpp`) to use this macro as an example of how it's used


michael-adler:
Can you think of a way to avoid the duplication in the various errors/*.h files?

r-rojo:
Hmm... Other one way could be a multi-line macro. Another could be a template class which would look like:

```C++
template<fpga_result R>
class opae_exception
{
public:
// blah blah



};
```
Then, the client code would have to do something like:
```
try {
// some operation(s)
} catch(const opae_exception<FPGA_BUSY> & e) {
// handle exception...
}
```

Personally, I'd rather keep them in separate classes/files but I'm open to suggestions.

michael-adler:
The problem I have with repetition like that is it becomes a barrier to changing the overall structure, since it is difficult to edit every single file.

tswhison:
Why not inherit from a "except" base class which implements the common methods? "except" could implement what() and operator fpga_result().  Then, each fpga_result-specific class just needs a c'tor.

r-rojo:
@tswhison , that could be worth a shot...
Let me try it out

tswhison:
Whoops, CI failed.  I like the refactor.  Can we compress each of the exception sub-classes into except.h?

r-rojo:
Yeah... I think I may have forgotten to initialize msg_ in one of the c’tors. 
I’ll look at that later and roll up all the exceptions into one file while I’m at it.

michael-adler:
That's better!

Now that you've refactored, the boilerplate (copyright, namespace, etc.) is larger than the content. Is there a good reason each of these exception classes needs to have its own file? Maybe just put all of the exception classes in one file?

common/include/opae/cxx/core/except.h
```diff
@@ -99,10 +111,259 @@ class except : public std::exception {
...
+}
+
+}  // end of namespace detail
+
+/// Macro to check of result is FPGA_OK
+/// If not, throw exception that corresponds
+/// to the result code
+#define ASSERT_FPGA_OK(r) \
+  opae::fpga::types::detail::assert_fpga_ok(r, \
+      opae::fpga::types::src_location(__FILE__, __func__, __LINE__));
```
tswhison: Looks like you got rid of the code that shortens __FILE__ to the filename.  This will display whatever relative path the compiler happens to use.  Not sure we wan't that.

tswhison: Nvm, I see it now.

PR #279, opened 2018-03-19
Add note about unplugging VFs from a guest

luebbers:
Adds a note in the driver architecture guide about issues with
unplugging a VF while a guest app is still accessing it.


doc/src/drv_arch/drv_arch.md
```diff
@@ -82,6 +82,15 @@ user-space before calling the reconfiguration IOCTL.
 
 1. Unload the driver from the guest
 2. Unplug the VF from the guest
+
+.. note::
+
+```
+    NOTE: Unplugging the VF from the guest while an application on the guest is
+    still accessing its resources may lead to VM instabilities. We recommend
+    cloing all applications accessing the VF in the guest before unplugging the
```
r-rojo: typo in `cloing`?
Also, what about rephrasing to:
```
It is recommended that all applications using a resource be closed prior to unplugging the VF.
```

doc/src/drv_arch/drv_arch.md
```diff
@@ -82,6 +82,15 @@ user-space before calling the reconfiguration IOCTL.
 
 1. Unload the driver from the guest
 2. Unplug the VF from the guest
+
+.. note::
+
+```
+    NOTE: Unplugging the VF from the guest while an application on the guest is
```
r-rojo: Is it necessary to include `NOTE:` even though it' under the `.. note::` markup? I don't know how it will look once rendered.

luebbers: I don't know - I'm playing it safe and am sticking with the markup that's already there.

Looking at the current output at https://opae.github.io/0.13.0/docs/drv_arch/drv_arch.html, it doesn't seem to make a difference, though.

doc/src/drv_arch/drv_arch.md
luebbers: Cloing definitely sounds wrong. I'll fix it.

PR #280, opened 2018-03-20
Add definition of a board with 27 address bits per bank.

michael-adler:


michael-adler:
Test pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2267079
(ModelSim license problem, but even that compiled and is just waiting to run.)

michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2267599

PR #282, opened 2018-03-21
Add checks to ensure commands in system path

buie1:
Adding checks to ensure that the required commands for fpgabist to execute are included in the users system PATH.


tools/fpgabist/bist_common.py
```diff
@@ -78,6 +80,13 @@ def get_mode_from_path(gbs_path):
     return None
 
 
+def is_command_in_path(cmd):
+    for dir in os.getenv("PATH").split(':'):
+        if (os.path.exists(os.path.join(dir, cmd))):
```
r-rojo: `dir` is a keyword in Python. You might want to pick a different variable name.

tools/fpgabist/bist_common.py
```diff
@@ -78,6 +80,13 @@ def get_mode_from_path(gbs_path):
     return None
 
 
+def is_command_in_path(cmd):
```
r-rojo: I don't know if it makes sense to use it but `distutils.spawn`package has a function called `find_executable` that you may be able to use. Taken from the Python help - `help('distutils.spawn')`:
```
    find_executable(executable, path=None)
        Tries to find 'executable' in the directories listed in 'path'.
        
        A string listing directories separated by 'os.pathsep'; defaults to
        os.environ['PATH'].  Returns the complete filename or None if not found.
```

tools/fpgabist/fpgabist
```diff
@@ -84,4 +84,15 @@ if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     bist_common.global_arguments(parser)
     args = parser.parse_args()
+    reqs = []
+    for cmd in bist_common.REQ_CMDS:
+        if(not bist_common.is_command_in_path(cmd)):
```
dasm: Superfluous brackets:

```
if not bist.common.is_command_in_path(cmd):
```
is sufficient.

tools/fpgabist/fpgabist
```diff
@@ -84,4 +84,15 @@ if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     bist_common.global_arguments(parser)
     args = parser.parse_args()
+    reqs = []
```
dasm: I would suggest to create separate function for this to not pollute `__main__`.

tools/fpgabist/fpgabist
```diff
@@ -84,4 +84,15 @@ if __name__ == "__main__":
...
+    for cmd in bist_common.REQ_CMDS:
+        if(not bist_common.is_command_in_path(cmd)):
+            reqs.append(cmd)
+    if reqs:
+        req = ', '.join([r for r in reqs])
```
dasm: This is superfluous. `join` operates on iterables.
```
>>> list_ = ['1', '2', '3']
>>> req1 = ', '.join([el for el in list_])
>>> req2 = ', '.join(list_)
>>> req1 == req2
True
>>> req1
'1, 2, 3'
>>> req2
'1, 2, 3'
>>>
```

tools/fpgabist/fpgabist
```diff
@@ -84,4 +84,15 @@ if __name__ == "__main__":
...
+        print "Unable to find {} in system PATH.".format(req)
```
dasm: You can move `join` over here, and use it directly.

tools/fpgabist/fpgabist
```diff
@@ -84,4 +84,15 @@ if __name__ == "__main__":
...
+        print "Please see README for required commands"
+        print "Terminating BIST."
+        sys.exit(-1)
```
dasm: FYI. It doesn't need to be changed.
* Default return code for "No such file or directory" is `2`.
* You can pass any text to `exit()` and print it on exit, signaling `1` as exit code.
```
Help on built-in function exit in module sys:

exit(...)
    exit([status])

    Exit the interpreter by raising SystemExit(status).
    If the status is omitted or None, it defaults to zero (i.e., success).
    If the status is an integer, it will be used as the system exit status.
    If it is another kind of object, it will be printed and the system
    exit status will be one (i.e., failure).
```

tools/fpgabist/bist_common.py
```diff
@@ -78,6 +80,13 @@ def get_mode_from_path(gbs_path):
...
+    for dir in os.getenv("PATH").split(':'):
+        if (os.path.exists(os.path.join(dir, cmd))):
+            return True
+    return False
```
dasm: If you'll change your function to `find_executable` this won't be needed, although instead of having two `return`s you could do:
```
checked_paths = [os.path.exists(os.path.join(directory, cmd)) for directory in os.getenv("PATH").split(":")]
return any(checked_paths)
```
```
Help on built-in function any in module __builtin__:

any(...)
    any(iterable) -> bool

    Return True if bool(x) is True for any x in the iterable.
    If the iterable is empty, return False.
```

buie1:
I have an update for this PR. Want to test before pushing my next commit.

tools/fpgabist/bist_common.py
```diff
@@ -32,9 +33,21 @@
 
 # TODO: Use AFU IDs vs. names of AFUs
 BIST_MODES = ['bist_afu', 'dma_afu', 'nlb_mode_3']
+REQ_CMDS = ['lspci', 'fpgainfo', 'fpgaconf', 'fpgadiag', 'fpga_dma_test',
+            'bist_app']
 
 
 # Return a list of all available bus numbers
+def check_commands():
```
dasm: You could use better name.

tools/fpgabist/bist_common.py
```diff
@@ -32,9 +33,21 @@
...
+    terminate = False
+    for cmd in REQ_CMDS:
+        executable = find_executable(cmd, path=os.environ['PATH'])
```
dasm: By default, if `path` is not provided, `find_executable` function uses `os.environ['PATH']`

tools/fpgabist/bist_common.py
```diff
@@ -32,9 +33,21 @@
...
+        if executable == cmd:
```
dasm: From offline discussion I know what you're doing here.
Instead of using such thing, you could do:
```
path = os.environ['PATH'].split(os.pathsep)
# Remove current directory from PATH
del path[-1]
executable = find_executable(cmd, path=path)
```

tools/fpgabist/bist_common.py
```diff
@@ -32,9 +33,21 @@
...
+            terminate = True
+            print "Could not find {} in $PATH".format(cmd)
+    return terminate
```
dasm:
```
return all([find_executable(cmd) for cmd in REQ])
```

tools/fpgabist/fpgabist
```diff
@@ -84,4 +84,7 @@ if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     bist_common.global_arguments(parser)
     args = parser.parse_args()
+    if bist_common.check_commands():
+        sys.exit("Terminating BIST.")
```
dasm: This should be moved to function. It is nice to have all features enclosed in function.

tools/fpgabist/bist_common.py
buie1: This doesn't seem to remove the cwd from the PATH. When I made this fix and tested I'm still running into the same issue. going to forego using the find_executable method because its making my life much harder than it needs to be IMO

buie1:
@OPAE/tools There was one approval can we follow-up on this PR?

buie1:
@dasm Looks like using the `# pylint disable=W605` didnt work for the lines when addressing the issues.

tools/extra/fpgabist/bist_common.py
```diff
@@ -24,6 +24,7 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+from distutils.spawn import find_executable
```
dasm: You still need to update your PR, so you can remove this import. You're not using it anymore.

luebbers:
Hey! What's the status on this one?

buie1:
@dasm @r-rojo Can we get this approved and merged.
All changes have been addressed.

PR #284, opened 2018-03-21
fpgadiag: get pclock freq from nlb register

r-rojo:
The pClock frequency is static per device id but the getting the device
id from the properties is not implemented right now.
As a temporary workaround, the frequency will be encoded in status2
register (0x0170). Read the frequency from here and set it frequency_
member variable to that if it is non-zero.


r-rojo:
Thanks for testing it out, @michael-adler

PR #286, opened 2018-03-21
Enable DEB and RPM package generation

nakulkorde:
**RPM packaging:**
Addresses issues with RPM package generation in order to make it easier to be accepted in Fedora and RHEL community namely:
1) A single spec file generates multiple packages
2) libraries are installed in lib or lib64 directories based on the architecture

****Steps to generate RPM packages****
`$ cmake .. -DBUILD_ASE=1 -DCPACK_GENERATOR=RPM  -DCMAKE_INSTALL_PREFIX=/usr
 $ make package_rpm`

****DEB packaging:****
Enables generation of debian packages
**One thing that is missing is that the debian packages do not have a dependency field since that feature is not present in cmake 2.8. For eg: opae-devel can be installed without installing opae-libs(which, ideally should not be allowed)** ===> This has been resolved with my latest commit. Documentation will reflect that to build DEB packages on Ubuntu 14 and lower, one needs to update the cmake version to 3.x. Ubuntu 16 and up have the newer version of cmake. 
@abelardojarab will help with updating the documentation.

****Steps to generate DEB packages****
`$cmake .. -DBUILD_ASE=1 -DCPACK_GENERATOR=DEB  -DCMAKE_INSTALL_PREFIX=/usr
$ make package_deb`

**Providing CMAKE_INSTALL_PREFIX is optional, by default the install prefix will be /usr/local**


CMakeLists.txt
```diff
@@ -168,8 +168,24 @@ endif()
...
+      set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+endif() 
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+############################################################################
+## libs install location ###################################################
+############################################################################
+
+get_property(LIB64 GLOBAL PROPERTY FIND_LIBRARY_USE_LIB64_PATHS) 
+if ("${LIB64}" STREQUAL "TRUE")
+    set(LIB_INSTALL_DIR "lib64")
```
r-rojo: can this be moved up to line 174?

CMakeLists.txt
```diff
@@ -168,8 +168,24 @@ endif()
...
+else()
+    set(LIB_INSTALL_DIR "lib")
```
r-rojo: can this be moved up to line 176?

ase/api/CMakeLists.txt
```diff
@@ -63,6 +63,13 @@ set_target_properties(opae-c-ase PROPERTIES
   VERSION ${INTEL_FPGA_API_VERSION}
   SOVERSION ${INTEL_FPGA_API_VER_MAJOR})
 
-install(TARGETS opae-c-ase
-  LIBRARY DESTINATION lib
-  COMPONENT opaecase)
+get_property(LIB64 GLOBAL PROPERTY FIND_LIBRARY_USE_LIB64_PATHS)
```
r-rojo: Can this be refactored to be a CACHE variable setup in the root CMakeLists.txt?

libopae/CMakeLists.txt
```diff
@@ -106,10 +106,20 @@ if(CMAKE_BUILD_TYPE STREQUAL "Release")
   set_property(TARGET opae-c PROPERTY C_VISIBILITY_PRESET hidden)
 endif(CMAKE_BUILD_TYPE STREQUAL "Release")
 
-# Binary install locations
-install(TARGETS opae-c
-  LIBRARY DESTINATION lib
-  COMPONENT opaeclib)
+#install(TARGETS opae-c
+#  LIBRARY DESTINATION ${LIB_INSTALL_DIR}
```
r-rojo: Why is this commented out? Can it be refactored to use a CACHE variable so we don't have to call "get_property" in line 113?

nakulkorde: Yes that can be done. I will store it in a cached variable and use that variable throughout.

PR #287, opened 2018-03-22
libopae++: add MMIO tests

tswhison:
Add tests for class handle's MMIO API.

### DIFF
--- a/tests/unit/gtCxxMMIO.cpp
+++ b/tests/unit/gtCxxMMIO.cpp
@@ -83,6 +83,7 @@ static const uint64_t nlb0_afu_id_h = 0xd8424dc4a4a3c413ULL;
 /**
  * @test mmio_05
  * Given an open accelerator handle object<br>
+ * for the NLB0 accelerator,<br>
  * When I call handle::read_csr64() with csr_space 0<br>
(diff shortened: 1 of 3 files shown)


tests/unit/gtCxxMMIO.cpp
```diff
@@ -0,0 +1,118 @@
...
+  EXPECT_THROW(accel_->write_csr64(0, 0, 10), no_access);
+}
+
+static const uint64_t nlb0_dfh      = 0x1000000000001070ULL;
+static const uint64_t nlb0_afu_id_l = 0xf89e433683f9040bULL;
+static const uint64_t nlb0_afu_id_h = 0xd8424dc4a4a3c413ULL;
+
+/**
+ * @test mmio_05
+ * Given an open accelerator handle object<br>
```
r-rojo: Maybe put something in the preconditions that also says the accelerator handle represents nlb0... ??
```
Given an open accelerator handle<br>
And the handle represents NLB0 accelerator<br>
```

r-rojo:
There was a merge error when I merged master. It is fixed in feature/c++core so you should probably pull that into this branch.

PR #288, opened 2018-03-22
make use of opae c++ exceptions

r-rojo:
Update the following classes to use the `ASSERT_FPGA_OK` macro (which will throw exception when fpga_result code is not `FPGA_OK`):
* pvalue
* properties
* token
* dma_buffer
* events

### DIFF
--- a/common/include/opae/cxx/core/pvalue.h
+++ b/common/include/opae/cxx/core/pvalue.h
@@ -47,20 +47,19 @@ struct guid_t {
 
   /** Update the local cached copy of the guid.
    */
-  fpga_result update() {
+  void update() {
     fpga_result res = fpgaPropertiesGetGUID(*props_,
             reinterpret_cast<fpga_guid *>(data_.data()));
-    is_set_ = (FPGA_OK == res);
-    return res;
+    ASSERT_FPGA_OK(res);
+    is_set_ = true;
   }
 
   /** Return a raw pointer to the guid.
    * @retval nullptr if the guid could not be queried.
    */
   operator uint8_t *() {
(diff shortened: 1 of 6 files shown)


michael-adler:
Does it make sense to have a global mode in which the callee is responsible for checking the return value and raising an exception instead of the caller? That would leave the user code pretty clean, without having to explicitly wrap each call in the ASSERT_FPGA_OK macro.

r-rojo:
@michael-adler , what do you mean by callee/caller? Is callee the OPAE C++ API implementation or is it the OPAE C API implementation?  Is the caller the client code of the OPAE C++ API or is it the OPAE C++ implementation?

michael-adler:
E.g. in update() in pvalue.h, update() is the caller and fpgaPropertiesGetGUID() is the callee. If the OPAE methods check and raise the exception then the user code doesn't have to be riddled with ASSERT_FPGA_OK wrapping everything. You'd have to make the raising of exceptions in OPAE methods conditional, based on some global property.

I don't know that this is technically possible -- just that it would be nice not to have to check the return value on every call if an exception can be raised automatically as a side effect.

r-rojo:
I don't think it's technically possible to throw exceptions from C code but this C++ code isn't meant to be user code. It's the C++ implementation of the OPAE C++ core API. Users of this code should really be concerned with handling the exceptions thrown.
We did discuss to have some sort of global flag that controls the behavior of this implementation with respect to exceptions (always throw vs always log) and we ended up deciding that throwing exceptions for most non-FPGA_OK codes was the thing to do.

PR #289, opened 2018-03-22
Change AFU interface type's key from "name" to "class".

michael-adler:
Using "name" was a source of confusion every time I explained the PIM's AFU
JSON.  They would see "afu-top-interface:name" and think that's the module
name, which was a reasonable assumption.  It isn't.  Instead, it is the class
of the interface, defining the ports and not the module name.

For compatibility, the legacy key "name" is accepted as a synonym.

### DIFF
--- a/platforms/afu_top_ifc_db/ccip_std_afu.sv.template
+++ b/platforms/afu_top_ifc_db/ccip_std_afu.sv.template
@@ -28,7 +28,7 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-import ccip_if_pkg::*;
+`include "platform_if.vh"
 
 module ccip_std_afu
    (
--- a/platforms/afu_top_ifc_db/ccip_std_afu_avalon_mm.sv.template
+++ b/platforms/afu_top_ifc_db/ccip_std_afu_avalon_mm.sv.template
@@ -28,7 +28,7 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-`include "device_if.vh"
+`include "platform_if.vh"
 
 module ccip_std_afu
   #(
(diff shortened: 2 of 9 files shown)


michael-adler:
Tested both old JSON (using "name") and new JSON (using "class"). Both work:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2273220
http://sw-pert.altera.com/pert/pert.php?test_run_id=2273301

platforms/afu_top_ifc_db/ccip_std_afu.sv.template
```diff
@@ -28,7 +28,7 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-import ccip_if_pkg::*;
```
r-rojo: Is this change related to the `name` => `class` change?

platforms/afu_top_ifc_db/ccip_std_afu_avalon_mm.sv.template
```diff
@@ -28,7 +28,7 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-`include "device_if.vh"
```
r-rojo: Is this change related to the name => class change?

platforms/afu_top_ifc_db/ccip_std_afu_avalon_mm_legacy_wires.sv.template
```diff
@@ -28,7 +28,7 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-import ccip_if_pkg::*;
```
r-rojo: Is this change related to the name => class change?

platforms/afu_top_ifc_db/ccip_std_afu.sv.template
michael-adler: It's related cleanup. The AFU needs platform_if.vh to function properly.

platforms/afu_top_ifc_db/ccip_std_afu_avalon_mm.sv.template
michael-adler: Same here.

platforms/afu_top_ifc_db/ccip_std_afu_avalon_mm_legacy_wires.sv.template
michael-adler: And here.

PR #290, opened 2018-03-26
The PIM no longer offers auto mode for extra-timing-reg-stages.

michael-adler:
It was too confusing and, we have decided, of little value.  Instead, interface
classes will have consistent register requirements across platforms.  Any extra
buffering required at the PR boundary will be handled by supplied libraries.

Renamed "add-extra-timing-reg-stages" to simply "add-timing-reg-stages", since
that's now exactly what the code does.


michael-adler:
AFUs have already been updated. Several tests run:

New AFUs pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2279158
Old AFUs that have the deprecated field pass (default was 0 for auto, so their buffering is ok): http://sw-pert.altera.com/pert/pert.php?test_run_id=2279165

Synthesized and ran a few AFUs outside the test environment.

PR #292, opened 2018-03-26
reorg tools directories

r-rojo:
Reorganize tools directories to bucket the tools into two categories (base, extra):
The goal is to organize tools into categories that share common dependencies in addition to categories that can be used when making packages (RPM or DEB).

The two categories are described below:
* Base tools should be the group with the fewest dependencies and are independent of platform or bitstream. 
* Extra are exactly that - not basic - 😁 and are made up of tools that support testing, validation, and platform development.

Also modify CMakeLists.txt as necessary to account for the change in directories.

The new layout is descried below:
* base
  * fpgaconf
  * fpgad
  * fpgainfo
  * fpgaport

* extra
   * coreidle
   * fpgaflash
   * hssi
   * packager
   * userclk
   * c++utils
   * libopae++
   * mmlink
   * ras
   * fpgadiag
   * fpgabist


michael-adler:
My first reaction was yea - organization is good. Then I started reading the list of tools and the categories. Some are quite hard. For example, what makes fpgaconf intrinsic and userclk platform. What makes fpgainfo platform and not intrinsic?

The general question is, does hierarchy make it easier or harder to find a tool's source? I assume you're still planning to install them all in a shared bin directory. I don't have strong feelings either way, especially if everyone else thinks it is better.

Have you thought about how this will affect the merge of the opae-sdk-x/release/0.14 branch to the external repository?

r-rojo:
@michael-adler, I've updated the description to include more details for this change.

luebbers:
The merge will not be affected - there'll just be a "release/0.14" branch, but we're not merging that back into master. It may affect cherry-picking of tools changes, though I'd actually prefer if the release branches cherry-picked this change into their branches and then made changes to the tools.

I agree it's not perfect, but then again I don't think we can ever come up with a categorization that doesn't have any corner cases. My thoughts the ones you pointed out are:

- userclk is not a base tool because you don't *need* it for operating the device (user clocks should be set during PR from the metadata).
- fpgainfo is base, not platform (did that get changed between Michael's comment and mine?)

"support" was originally meant to be called "test" or "testing", since they're all tools for performance characterization or debug. But "test" and "testing" were deemed to confusing with the top-level "test" directory we already have. Any suggestions to improve on "support"?

luebbers:
Maybe some additional context: the background for this is that we want to create a tools package with only the absolute essentials that has a minimum of dependencies to accommodate customers who had trouble satisfying e.g. packager dependencies although they only needed fpgaconf and fpgainfo. To that end, a "base" and a "everything-else" categorization would also work, but that seemed a bit lazy. ;)

buie1:
I see no issue with this reorganization of these tool.  To me is seems like this is step towards a larger goal. I know there's been discussion on also having the tools be written in the same languages and have a deeper level of uniformity.  Care to make any comments on the bigger picture here?

luebbers:
One consequence of having a smaller subset of "base" tools is that if we can limit that subset to a single language and a reduced number of library dependencies, the resulting "opae-tools-base" package would only carry that reduced set of dependencies, which would open a wider range of target systems where it could be deployed.

PR #293, opened 2018-03-29
fpgainfo: Add first version of fpgainfo in C

r-rojo:
This adds a new cmake target, fpgainfo-c, which is compiled from C source.
When this is ready to replace the Python fpgainfo, we can rename the target/directory to fpgainfo.

This first version contains a main which parses the first level of arguments which include
the arguments to locate an accelerator or device (BDF and socket-id).

These arguments are used to populate a query filter (`fpga_properties`) which will later be used
when calling `fpgaEnumerate`

This main function also defines an array of `struct command_handler` objects where each item in
this array defines:
* command name - `{errors, port, fme, power, temp}`
* filter function - function called to add additional filters based on the sub-arguments
* run function - the function that performs the subcommand

When main identifies the `struct command_hander` object that pertains to the subcommand, it will
* Call the flter function to possibly add more filtering criteria
* Call `fpgaEnumerate` twice
  * Once to get the number of tokens for the filter, then allocate enough memory for tokens
  * Again to get the actual tokens
* Call the `run` function in the command handler (passing in the tokens and sub-arguments).

Finally, this adds to tests for `fpgainfo fme` and `fpgainfo port` commands


tools/base/fpgainfo-c/portinfo.c
```diff
@@ -0,0 +1,123 @@
...
+
+struct port_info {
+	fpga_guid guid;
+	uint64_t object_id;
+	uint8_t bus;
+	uint8_t device;
+	uint8_t function;
+	uint8_t socket_id;
+	uint16_t device_id;
+	char model[MODEL_SIZE];
```
buie1: I believe this struct should include `afu_id` as one of the things reported by port_info.

tools/base/fpgainfo-c/portinfo.c
```diff
@@ -0,0 +1,123 @@
...
+};
+
+fpga_result get_port_info(fpga_token tok, struct port_info *finfo)
+{
+	fpga_result res = FPGA_OK;
+	fpga_properties props;
+	res = fpgaGetProperties(tok, &props);
```
buie1: I've noticed that `fpgaGetProperties` and the other functions in the `properties.h` file don't include things like `AFU_ID` and `TEMPERATURE` is there a way to get these included? I think they are valuable things that we would want to include. I noticed afu_id is included in `sysfs.c` but I feel can be apart of the common properties.

r-rojo: @buie1 , 
AFU_ID is the GUID property for a resource of type `FPGA_ACCELERATOR`

`TEMPERATURE` is part of the metrics API which hasn't been implemented yet.

cmake/modules/test_config.cmake
```diff
@@ -155,7 +173,10 @@ function(Build_Test_Target Target_Name Target_LIB)
                                $<BUILD_INTERFACE:${LIB_SRC_PATH}>)
                       
     target_link_libraries(${Target_Name} commonlib safestr ${Target_LIB} ${libjson-c_LIBRARIES} 
-                              uuid ${GTEST_BOTH_LIBRARIES} dl opae-c++-utils opae-c++ opae-cxx-core)
+                              uuid ${GTEST_BOTH_LIBRARIES} dl
+                              argsfilter
+                              opae-c++-utils
+                              opae-c++ opae-cxx-core)
```
asgibbons: Do we still need the C++ libraries?

PR #294, opened 2018-03-29
Fixes to automatically update documentation on opae.github.io

nakulkorde:
- Add a directory 'latest' that stores the latest documentation that is generated with every Travis build on a commit that is not tagged for release
- Generate and update a landing page to link to the respective document versions: Currently just a basic page with links and will go through improvements in visualization after the release.


nakulkorde:
There seems to be an issue with the document generation, the documentation is generated and uploaded to opae.github.io correctly but the webpages don't seem to be loading with the proper theme.
@luebbers @abelardojarab @r-rojo can you think of a reason why there could be a problem?

abelardojarab:
@nakulkorde it seems it is not finding the .css style file

nakulkorde:
I fixed it, just had to add the .nojekyll file to the root directory so that github pages would not try to use it's default parsing.

scripts/build-documentation.sh
```diff
@@ -18,8 +18,9 @@ COMMIT_ID="$(git rev-parse HEAD)"
 TEMP=$((git describe --exact-match $COMMIT_ID) 2>&1)
 
 if [[ $TEMP != *"fatal"* ]]; then
-    echo "New tag detected; Uploading the documentation to opae.github.io"
-    /bin/bash scripts/push-documentation.sh 
+    echo "New tag detected; Uploading documentation under new tag on opae.github.io"
+    /bin/bash ../scripts/push-documentation.sh new_tag
 else
-    echo "No new tag detected; Documentation will not be uploaded to opae.github.io"
+    echo "No new tag detected; Latest documentation will be uploaded to opae.github.io"
```
luebbers: Is this right? E.g. what happens on pull request builds (without tags) - they shouldn't update "latest", should they?

nakulkorde: This is because we don't have a daily or a weekly build on travis. Should this be added in some sort of scheduled build?

nakulkorde: The script now uploads the documents from master branch instead of the PR branch if no tag is detected. That way, "latest" will always have the docs version that is present on master.

doc/doxygen/CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ endif(DOXYGEN_DOT_FOUND)
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Doxyfile.in
                ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile @ONLY)
-add_custom_target(doxygen
+add_custom_target(doxygen ALL
```
r-rojo: so we are building the documentation by default then?

scripts/index_generator.py
```diff
@@ -0,0 +1,41 @@
...
+{% endfor %}
+</ul>
+</body>
+</html>"""
+
+
+def main():
+    versions = []
+    for x in os.listdir('.'):
+        if os.path.isdir(x):
```
r-rojo: a regex might be better...
```Python
import re
...
for x in os.listdir(somedir):
    if re.match('\d+\.\d+\.\d+', x):
        versions.append(x)
```

This way, we only look at directories that match a version format. This example only works if the versions are numeric and ignores `latest` directory. Maybe something we can do later?

luebbers: As long as someone writes it down for later. :)

PR #296, opened 2018-03-30
Add error API

luebbers:
Implements OPAE error API.

Exposes a property on the number of supported error registers.
Users can get information about the errors using fpgaGetErrorInfo(),
read errors using fpgaReadError() and clear errors using
fpgaClearError() and fpgaClearAllErrors().

Error sysfs nodes are dynamically parsed from the respective
sysfs directories (some are excluded based on a blacklist of
names). If there is a "clear" file, the "errors" file in sysfs is
assumed do be "clearable" (so far only true for PORT errors).


libopae/src/error.c
```diff
@@ -0,0 +1,286 @@
...
+			n--;
+			break;
+		}
+		strncpy_s(new_entry->info.name, FPGA_ERROR_NAME_MAX, de->d_name, FILENAME_MAX);
+		strncpy_s(new_entry->error_file, SYSFS_PATH_MAX, basedir, FILENAME_MAX);
+		new_entry->next = NULL;
+		// see if error can be cleared (currently only errors called "errors" corresponding
+		// to a "clear" file can)
+		new_entry->info.can_clear = false;
+		if (strcmp(de->d_name, "errors") == 0) {    // FIXME: SAFE
```
r-rojo: Will this `FIXME` be addressed in another pull request?

libopae/src/error.c
```diff
@@ -0,0 +1,286 @@
...
+		if (!new_entry) {
+			FPGA_MSG("can't allocate memory");
+			n--;
+			break;
+		}
+		strncpy_s(new_entry->info.name, FPGA_ERROR_NAME_MAX, de->d_name, FILENAME_MAX);
+		strncpy_s(new_entry->error_file, SYSFS_PATH_MAX, basedir, FILENAME_MAX);
+		new_entry->next = NULL;
+		// see if error can be cleared (currently only errors called "errors" corresponding
+		// to a "clear" file can)
```
r-rojo: So clearing errors will only be supported for errors that are cleared by writing to the `clear` file?
Will we eventually support clearing other types of errors?

libopae/src/error.c
```diff
@@ -0,0 +1,286 @@
...
+}
+
+/* files and directories to ignore when looking for errors */
+const char *errors_exclude[] = {
+	"revision",
+	"uevent",
+	"power",
+	"clear"
+};
+#define NUM_ERRORS_EXCLUDE (sizeof(errors_exclude) / sizeof(char *))
```
r-rojo: Wouldn't it be better to more explicit on what nodes we operate on and whitelist the files? 
This would make it easier to determine how to clear errors too (without relying on the existence of a "clear" file).

libopae/src/error.c
luebbers: No - I'll address that in this PR; it'll get shot down by static analysis otherwise.

libopae/src/error.c
luebbers: We should. This follows what the mock driver models, but that should be expanded.

libopae/src/error.c
luebbers: I was looking for something that minimizes maintenance if people start adding additional errors. Perhaps being more explicit about the naming of clearable errors would help (like "errors" and "errors_clear")?

luebbers:
There are a number of limitations with the current approach in terms of performance and thread safety. I'll close this PR and open a new one.

libopae/src/token_list.c
```diff
@@ -85,6 +86,11 @@ struct _fpga_token *token_add(const char *sysfspath, const char *devpath)
 		return NULL;
 	}
 
+	/* populate error list */
+	char errpath[SYSFS_PATH_MAX];
+	snprintf_s_s(errpath, SYSFS_PATH_MAX, "%s/errors", sysfspath);
```
r-rojo: what about fme errors? those are found in the subdirectory `errors/fme-errors/errors` with their clear file being `errors/fme-errors/clear`

r-rojo: Nevermind, I forgot this function recurses subdirectories

lzhan55:
Just have a general question about error registers: are they part of the FME error registers located in the error DFH of FME? Or, each AFU has its own error registers in AFU's MMIO? Is there any error register related to AFUs?

luebbers:
Added support for clearing FME errors and rebased on top of master. Please review again. Thanks! :)

libopae/src/error.c
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	while ((de = readdir(dir))) {
+		// skip hidden ('.*') files (includes "." and "..")
+		if (de->d_name[0] == '.')
+			continue;
+
+		// skip names on blacklist
+		for (i = 0; i < NUM_ERRORS_EXCLUDE; i++) {
+			if (strcmp(de->d_name, errors_exclude[i]) == 0) {  // FIXME: SAFE?
```
r-rojo: You should probably use the safestring version of this

libopae/src/error.c
```diff
@@ -0,0 +1,328 @@
...
+				break;
+			}
+		}
+		if (i < NUM_ERRORS_EXCLUDE)
+			continue;
+
+		// build absolute path
+		strncpy(basedir + len, de->d_name, FILENAME_MAX - len);
```
r-rojo: You should probably use one of the safestring variants of this.

libopae/src/error.c
```diff
@@ -0,0 +1,328 @@
...
+
+		// append error info to list
+		struct error_list *new_entry = malloc(sizeof(struct error_list));
+		if (!new_entry) {
+			FPGA_MSG("can't allocate memory");
+			n--;
+			break;
+		}
+		strncpy_s(new_entry->info.name, FPGA_ERROR_NAME_MAX, de->d_name, FILENAME_MAX);
+		strncpy_s(new_entry->error_file, SYSFS_PATH_MAX, basedir, FILENAME_MAX);
```
r-rojo: Should we check for errors (by looking at the return value of `strncpy_s`)?

libopae/src/error.c
```diff
@@ -0,0 +1,328 @@
...
+		new_entry->next = NULL;
+		// Errors can be cleared:
+		//   * if the name is "errors" and there is a file called "clear" (generic case), OR
+		//   * if the name is in the "errors_clearable" table
+		new_entry->info.can_clear = false;
+		if (strcmp(de->d_name, "errors") == 0) {
```
r-rojo: safestring version of strcmp?

libopae/src/error.c
```diff
@@ -0,0 +1,328 @@
...
+			strncpy_s(basedir + len, FILENAME_MAX - len, "clear", sizeof("clear"));
+			// try accessing clear file
+			if (lstat(basedir, &st) != -1) {
+				new_entry->info.can_clear = true;
+				strncpy_s(new_entry->clear_file, SYSFS_PATH_MAX, basedir, FILENAME_MAX);
+			}
+		} else {
+			for (i = 0; i < NUM_ERRORS_CLEARABLE; i++) {
+				if (strcmp(de->d_name, errors_clearable[i]) == 0) {
```
r-rojo: safestring version of strcmp?

libopae/src/error.c
```diff
@@ -0,0 +1,328 @@
...
+		// Errors can be cleared:
+		//   * if the name is "errors" and there is a file called "clear" (generic case), OR
+		//   * if the name is in the "errors_clearable" table
+		new_entry->info.can_clear = false;
+		if (strcmp(de->d_name, "errors") == 0) {
+			strncpy_s(basedir + len, FILENAME_MAX - len, "clear", sizeof("clear"));
+			// try accessing clear file
+			if (lstat(basedir, &st) != -1) {
+				new_entry->info.can_clear = true;
+				strncpy_s(new_entry->clear_file, SYSFS_PATH_MAX, basedir, FILENAME_MAX);
```
r-rojo: Check return value of `strncpy_s`?

libopae/src/error.c
```diff
@@ -0,0 +1,328 @@
...
+			}
+		} else {
+			for (i = 0; i < NUM_ERRORS_CLEARABLE; i++) {
+				if (strcmp(de->d_name, errors_clearable[i]) == 0) {
+					strncpy_s(basedir + len, FILENAME_MAX - len, de->d_name, FILENAME_MAX);
```
r-rojo: Check return value of `strncpy_s`?

libopae/src/error.c
```diff
@@ -0,0 +1,328 @@
...
+					// try accessing clear file
+					if (lstat(basedir, &st) != -1) {
+						new_entry->info.can_clear = true;
+						strncpy_s(new_entry->clear_file, SYSFS_PATH_MAX, basedir, FILENAME_MAX);
```
r-rojo: Check return value of `strncpy_s`?

libopae/src/error.c
```diff
@@ -0,0 +1,328 @@
...
+					}
+				}
+			}
+		}
+
+		if (!new_entry->info.can_clear) {
+			memset_s(new_entry->clear_file, sizeof(new_entry->clear_file), 0);
```
r-rojo: Check return value of `memset_s`?

libopae/src/error.c
```diff
@@ -0,0 +1,328 @@
...
+	uint32_t n = 0;
+	unsigned int i;
+	struct error_list **el = list;
+
+	if (len+1 > FILENAME_MAX) {
+		FPGA_MSG("path too long");
+		return 0;
+	}
+
+	strncpy(basedir, path, FILENAME_MAX-1);
```
r-rojo: safestring version?

libopae/src/error.c
```diff
@@ -0,0 +1,328 @@
...
+#define NUM_ERRORS_CLEARABLE (sizeof(errors_clearable) / sizeof(char *))
+
+/* Walks the given directory and adds error entries to `list`.
+ * This function is called during enumeration when adding tokens to
+ * the global tokens list. When tokens are cloned, their error
+ * lists are only shallowly copied (which works because errors of
+ * a token never change).
+ * returns the number of error entries added to `list` */
+uint32_t __FIXME_MAKE_VISIBLE__
+build_error_list(const char *path, struct error_list **list)
```
r-rojo: Just out of curiosity, what happens if we call this twice with the same list? Will it duplicate the list?

libopae/src/error.c
```diff
@@ -0,0 +1,328 @@
...
+			n--;
+			break;
+		}
+		strncpy_s(new_entry->info.name, FPGA_ERROR_NAME_MAX, de->d_name, FILENAME_MAX);
+		strncpy_s(new_entry->error_file, SYSFS_PATH_MAX, basedir, FILENAME_MAX);
+		new_entry->next = NULL;
+		// Errors can be cleared:
+		//   * if the name is "errors" and there is a file called "clear" (generic case), OR
+		//   * if the name is in the "errors_clearable" table
+		new_entry->info.can_clear = false;
```
r-rojo: Should the routine to check if an error file is clearable be refactored into its own function?

libopae/src/error_int.h
```diff
@@ -0,0 +1,52 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct error_list {
+	struct fpga_error_info info;
+	struct error_list *next;
+	char error_file[SYSFS_PATH_MAX];
+	char clear_file[SYSFS_PATH_MAX];
+};
```
r-rojo: Have you thought about wrapping this in an outer struct with metadata (like the error count)?
It could look something like (and maybe more interesting metadata):
```C
struct error_info {
        struct error_list *errors;
        size_t count;
};
```

libopae/src/error.c
luebbers: strcmp() is not a critical/banned function. But I should remove the comment.

libopae/src/error.c
luebbers: Yes, it would append all errors it finds in `path` to the list `list`. It actually calls itself recursively. I'll add a comment.

libopae/src/error.c
luebbers: Yes, we should. I'll add it.

libopae/src/error.c
luebbers: Good idea. I'll look into it.

libopae/src/error.c
luebbers: `strcmp()` is not a critical/banned function.

libopae/src/error.c
luebbers: Yep, see above.

libopae/src/error.c
luebbers: See above (not critical).

libopae/src/error.c
luebbers: See above.

libopae/src/error.c
luebbers: ...

libopae/src/error.c
luebbers: Good idea, will do.

libopae/src/error_int.h
luebbers: `error_info` is the structure handled (and accessed directly) by users of the API. I'd rather not expose implementation-specific information (like sysfs paths) to them, since they'll start relying on that and it will break as soon as (for example) there's not sysfs entry for an error.

libopae/src/error.c
luebbers: done

libopae/src/error.c
luebbers: done

libopae/src/error.c
luebbers: done

libopae/src/error.c
r-rojo: Ok, it can be a TODO...

PR #297, opened 2018-03-31
Add argsfilter static library

r-rojo:
Add a static library, argsfilter, that can be used to parse common
command line arguments (bus, device, function, socket-id).

The only function this contains is set_properties_from_args which
can modify an fpga_properties filter by setting properties
when their corresponding arguments are found in the command line
vector.

NOTE:
* This will modify the argc/argv variables passed in so those options are "removed" from the command line vector.
* Currently, it is up to the client to display correct help/usage information that includes these common options

This is built as a static library with PIC turned on so it can be
linked into shared libraries.


tests/unit/gtArgsFilter.cpp
```diff
@@ -0,0 +1,392 @@
...
+  int exit_code = fpgaPropertiesFromArgs(m_Properties, &res, &argc,
+                                         const_cast<char **>(argv));
+  EXPECT_EQ(exit_code, 0);
+  EXPECT_EQ(res, FPGA_EXCEPTION);
+  EXPECT_EQ(5, argc);
+  EXPECT_STREQ(argv[0], "prog");
+  EXPECT_STREQ(argv[1], "--opt1");
+  EXPECT_STREQ(argv[2], "val1");
+  EXPECT_STREQ(argv[3], "word1");
+  EXPECT_STREQ(argv[4], "word2");
```
tswhison: Not sure that these checks on argc and argv are adding value?

tests/unit/gtArgsFilter.cpp
```diff
@@ -0,0 +1,392 @@
...
+  EXPECT_EQ(FPGA_OK, fpgaPropertiesGetBus(m_Properties, &value));
+  EXPECT_EQ(0x5e, value);
+  EXPECT_EQ(exit_code, 0);
+  EXPECT_EQ(res, FPGA_OK);
+  EXPECT_EQ(5, argc);
+  EXPECT_STREQ(argv[0], "prog");
+  EXPECT_STREQ(argv[1], "--opt1");
+  EXPECT_STREQ(argv[2], "val1");
+  EXPECT_STREQ(argv[3], "word1");
+  EXPECT_STREQ(argv[4], "word2");
```
tswhison: I don't follow what this is doing.  Does fpgaPropertiesFromArgs() mutate the argv parameter?

r-rojo: Yes, as mentioned in the note in the description of this PR. 😄

tswhison: Doh.. (slurps coffee to improve focus)

tests/unit/gtArgsFilter.cpp
r-rojo: This just verifies that calling `fpgaPropertiesFromArgs` (without the relevant arguments) doesn't mutate the argument vector.

tools/base/argsfilter/argsfilter.c
```diff
@@ -0,0 +1,183 @@
...
+	int bus;
+	int device;
+	int function;
+	int socket_id;
+} config = {.bus = -1, .device = -1, .function = -1, .socket_id = -1};
+
+#define SUPPORTED_OPTIONS 4
+
+int fpgaPropertiesFromArgs(fpga_properties filter, fpga_result *result,
+			   int *argc, char *argv[])
```
luebbers: The `fpga` prefix is usually reserved for public OPAE FPGA APIs - can we call this something else?

PR #300, opened 2018-04-03
Add platforms directory

r-rojo:
platforms directory was inadvertanly removed from cmake processing.
This adds it back.


southerngs:
@r-rojo why didn't Travis flag removing this directory as an error?  Can we add something to Travis to check to make sure the tools builds correctly?  This is especially important since these changes get automatically applied in sdk-x.

michael-adler:
ASE isn't tested yet, but we can figure out some invocation of afu_platform_config to add to Travis that will fail if the platforms tree is completely broken.

PR #306, opened 2018-04-05
Simplify cmake build by providing hints on how to install dependencies

nakulkorde:
Check whether all the necessary dependencies are installed, if not notify the user accordingly.
Also, display how to install the respective packages in DEB/RPM environments.

The absolutely required dependencies libuuid and libjsonc if not found will throw a fatal error in the end with details on how to install them.
Other dependencies like Python which are only required for certain tools will give out messages that they are not present and the respective components will not function.
Basic dependencies like C/C++ compiler will throw out messages on how to install them and CMake will leave the configuration incomplete automatically.


nakulkorde:
I tried to remove compiler_config.cmake from the commits but the issues is I had edited the file in a commit and then removed the edited code in a later commit. Hence the file shows up.

cmake/modules/dependency_notifier.cmake
```diff
@@ -0,0 +1,101 @@
...
+   please specify the location by defining the variable LIBJSON-C_ROOT in 
+   your cmake command as follows: cmake <path to clone dir> -DLIBJSON-C_ROOT=<path to install location>")
+endif()
+
+if(NOT libuuid_FOUND)
+    message("-- uuid not found. Please install uuid package for your respective distribution:
+    DEB: apt-get install uuid-dev
+    RPM: yum install libuuid
+   If you have already installed this package in a nonstandard location 
+   please specify the location by defining the variable LIBJSON-C_ROOT in
```
r-rojo: copy/paste error? this is referring to` LIBJSON-C_ROOT`, not `libuuid_ROOT`

nakulkorde: Fixed it. Thanks for pointing out.

cmake/modules/dependency_notifier.cmake
```diff
@@ -0,0 +1,101 @@
...
+   please install python-sphinx:
+    DEB: apt-get install python-sphinx
+    RPM: Please follow the official documentation to install Sphinx: http://www.sphinx-doc.org/en/stable/install.html
+   If you have already installed this package in a nonstandard location 
+   please specify the location by defining the variable SPHINX_ROOT in 
+   your cmake command as follows: cmake <path to clone dir> -DSPHINX_ROOT=<path to install location>")
+endif()
+
+
+if((NOT libjson-c_FOUND) AND (NOT libuuid_FOUND))
```
r-rojo: Can this be simplified? Maybe have to cmake vars, REQUIRED_DEPENDENCIES and OPTIONAL_DEPENDENCIES. Then, above, for example, when UUID is not found you can do:
```cmake
set(REQUIRED_DEPENDENCIES "libuuid ${REQUIRED_DEPENDENCIES}")
```
Then, instead of checking the different combinations, simple state:
```cmake
message(FATAL_ERROR,
  "The following dependencies are required and libopae-c will not be built without them. ${REQUIRED_DEPENDENCIES}")
```

Just a suggestion...

cmake/modules/dependency_notifier.cmake
```diff
@@ -50,7 +50,7 @@ if(NOT libuuid_FOUND)
     DEB: apt-get install uuid-dev
     RPM: yum install libuuid
    If you have already installed this package in a nonstandard location 
-   please specify the location by defining the variable LIBJSON-C_ROOT in 
+   please specify the location by defining the variable UUID_ROOT in
```
r-rojo: We have the finder looking for at `LIBUUID_ROOT`
https://github.com/OPAE/opae-sdk/blob/de5d378bd534311b393470771d7ad499a29dd1b4/cmake/modules/FindUUID.cmake#L17

r-rojo:
So some files show up in this PR but there seem to be no real diffs.  Is it because these were opened from Windows? Are the deltas carriage returns?

nakulkorde:
@r-rojo yes, that is the case for two of the files. Do you know how I can fix them? I thought of rebasing but it will not solve this issue. 
When I checked out the files, they got converted to CRLF on windows and that is the reason they were overwritten by a merge from master.

r-rojo:
At the very least, you can remove the files that are really unmodified from the change. About the other files, you can probably run dos2unix on them to remove the line endings. You may also want to change your editor preferences on Windows to use unix style line endings so this doesn't happen again in the future.

cmake/modules/dependency_notifier.cmake
```diff
@@ -0,0 +1,86 @@
...
+    message("-- No C++ compiler was found. Please install the g++ package for your distribution:
+    DEB: apt-get install g++
+    RPM: yum install gcc-c++") 
+endif()
+
+if(NOT libjson-c_FOUND)
+    message("-- json-c not found. Please install json-c package for you respective distribution:
+    DEB: apt-get install libjson0-dev
+    RPM: yum install json-c-devel
+   If you have already installed this package in a nonstandard location
```
r-rojo: should be `non-standard`

cmake/modules/dependency_notifier.cmake
```diff
@@ -0,0 +1,86 @@
...
+   please specify the location by defining the variable LIBJSON-C_ROOT in 
+   your cmake command as follows: cmake <path to clone dir> -DLIBJSON-C_ROOT=<path to install location>")
+   set(REQUIRED_DEPENDENCIES "libjson-c ${REQUIRED_DEPENDENCIES}")
+endif()
+
+if(NOT libuuid_FOUND)
+    message("-- uuid not found. Please install uuid package for your respective distribution:
+    DEB: apt-get install uuid-dev
+    RPM: yum install libuuid-devel
+   If you have already installed this package in a nonstandard location
```
r-rojo: ditto here - `non-standard`

cmake/modules/dependency_notifier.cmake
```diff
@@ -0,0 +1,86 @@
...
+    RPM: yum install gcc-c++") 
+endif()
+
+if(NOT libjson-c_FOUND)
+    message("-- json-c not found. Please install json-c package for you respective distribution:
+    DEB: apt-get install libjson0-dev
+    RPM: yum install json-c-devel
+   If you have already installed this package in a nonstandard location 
+   please specify the location by defining the variable LIBJSON-C_ROOT in 
+   your cmake command as follows: cmake <path to clone dir> -DLIBJSON-C_ROOT=<path to install location>")
```
r-rojo: maybe `<path to json-c install location>`

cmake/modules/dependency_notifier.cmake
```diff
@@ -0,0 +1,86 @@
...
+   set(REQUIRED_DEPENDENCIES "libjson-c ${REQUIRED_DEPENDENCIES}")
+endif()
+
+if(NOT libuuid_FOUND)
+    message("-- uuid not found. Please install uuid package for your respective distribution:
+    DEB: apt-get install uuid-dev
+    RPM: yum install libuuid-devel
+   If you have already installed this package in a nonstandard location 
+   please specify the location by defining the variable LIBUUID_ROOT in 
+   your cmake command as follows: cmake <path to clone dir> -DLIBUUID_ROOT=<path to install location>")
```
r-rojo: maybe `<path to uuid install location>`

cmake/modules/dependency_notifier.cmake
```diff
@@ -0,0 +1,86 @@
...
+    message("-- No suitable Python interpreter found. Some Python based tools will not function. Please install Python >= 2.7.")
+endif()
+
+if(NOT DOXYGEN_FOUND)
+    message("-- Doxygen not found. Documentation will not be built. If you want documentation to be built, please install doxygen:
+    DEB: apt-get install doxygen
+    RPM: yum install doxygen
+   If you have already installed this package in a nonstandard location 
+   please specify the location by defining the variable DOXYGEN_ROOT in 
+   your cmake command as follows: cmake <path to clone dir> -DDOXYGEN_ROOT=<path to install location>")
```
r-rojo: maybe `<path to doxygen install location>`

cmake/modules/dependency_notifier.cmake
```diff
@@ -0,0 +1,86 @@
...
+    RPM: yum install libuuid-devel
+   If you have already installed this package in a non-standard location 
+   please specify the location by defining the variable LIBUUID_ROOT in 
+   your cmake command as follows: cmake <path to clone dir> -DLIBUUID_ROOT=<path to uuid install location>")
+   set(REQUIRED_DEPENDENCIES "libuuid ${REQUIRED_DEPENDENCIES}")
+endif()
+
+find_package(PythonInterp 2.7)
+if(NOT PYTHONINTERP_FOUND)
+    message("-- No suitable Python interpreter found. Some Python based tools will not function. Please install Python >= 2.7.")
```
luebbers: Is any version of Python 2.7 good, or do we require a specific one (like 2.7.13 or higher)?

rpSTTD: RHEL 7.3 has python 2.7.5 as default.

PR #308, opened 2018-04-06
Correct MMIO read timeout value.

michael-adler:


michael-adler:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2304296

lzhan55:
@michael-adler How was this timeout value set? Based on ASE simulation tests?

michael-adler:
It was set to match the HW.

PR #309, opened 2018-04-08
Update clock crossing FIFOs and Avalon buffering components

michael-adler:
- New script scrapes Quartus internal modules automatically to avoid having to run Qsys to
  update.  Tested with Quartus 16 through 18.
- Enable synchronous reset trees.


michael-adler:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2306315

The sources here are copies of IP shipped with Quartus and made unique by changing the name. The only change is a recent version of Quartus IP was used, which includes SDC updates and support for synchronous reset.

Tested on HW and works with Quartus all the way back to version 16, which is used by integrated Xeon+FPGA. Ran MPF torture tests on all versions. Also tested hello_mem_afu on discrete HW, which is the only main user of this code.

michael-adler:
Full DCP 1.1 platform & AFU build had no timing violations.

PR #310, opened 2018-04-09
Clear Port errors and report errors after PR

buie1:
Adding a critical fix for DCP. Right now during reconfiguration we may need to need to clear port errors for a maximum of 64 transactions to ensure there all port errors at clear.
Slight refactoring. Thought it would be a good idea to report if there were any port errors on the system after the PR was completed.


libopae/src/reconf.c
```diff
@@ -133,6 +135,17 @@ static fpga_result validate_bitstream(fpga_handle handle,
 	}
 }
 
+static fpga_result read_port_error(char *errpath, char *path, uint64_t *error)
```
luebbers: I'm not certain that we need to introduce this function here.

How much does this buy us in terms of re-use? I only see this function called twice. Can we just use `sysfs_read_u64()` there?

libopae/src/reconf.c
```diff
@@ -148,17 +161,19 @@ static fpga_result clear_port_errors(fpga_handle handle)
 		return result;
 	}
 
-	snprintf_s_ss(syfs_errpath, sizeof(syfs_errpath), "%s/%s", syfs_path, PORT_SYSFS_ERRORS);
 	// Read port error.
-	result = sysfs_read_u64(syfs_errpath, &error);
+	result = read_port_error(syfs_errpath, syfs_path, &error);
```
luebbers: `read_port_error()` modifies `sysfs_errpath`, but it's never used afterwards. From a "what does this function do" perspective, it's not clear that `read_port_error()` needs the user to pass in a buffer that then is used to construct a path that's not used again.

If you want to keep the function, you could allocate a new buffer for the resulting path inside. That way you could also be sure about its size. But as noted above, you could also just put the `snprintf_s_ss()` call here and use `sysfs_read_u64()`.

libopae/src/reconf.c
```diff
@@ -133,6 +135,17 @@ static fpga_result validate_bitstream(fpga_handle handle,
 	}
 }
 
+static fpga_result read_port_error(char *errpath, char *path, uint64_t *error)
+{
+	snprintf_s_ss(errpath, sizeof(errpath), "%s/%s", path, PORT_SYSFS_ERRORS);
+	return sysfs_read_u64(errpath, error);
+}
+
+static fpga_result clear_error(char *errpath, char *path, uint64_t error)
```
luebbers: See above (`read_port_error()`), but this is used only once.

libopae/src/reconf.c
```diff
@@ -133,6 +135,17 @@ static fpga_result validate_bitstream(fpga_handle handle,
 	}
 }
 
+static fpga_result read_port_error(char *errpath, char *path, uint64_t *error)
+{
+	snprintf_s_ss(errpath, sizeof(errpath), "%s/%s", path, PORT_SYSFS_ERRORS);
```
luebbers: Does this work? `sizeof(errpath)` is 8 (`errpath` is a pointer).

libopae/src/reconf.c
```diff
@@ -148,17 +161,19 @@ static fpga_result clear_port_errors(fpga_handle handle)
...
 	}
+	// Check if errors exsist
+	if (error != 0) {
+		FPGA_ERR("Port Errors found");
+	}
 
-	snprintf_s_ss(syfs_errpath, sizeof(syfs_errpath), "%s/%s", syfs_path, PORT_SYSFS_ERR_CLEAR);
 	// Clear port error.
-	result = sysfs_write_u64(syfs_errpath, error);
+	result = clear_error(syfs_errpath, syfs_path, error);
```
luebbers: See above.

libopae/src/reconf.c
```diff
@@ -408,6 +431,16 @@ fpga_result __FPGA_API__ fpgaReconfigureSlot(fpga_handle fpga,
 		result = FPGA_RECONF_ERROR;
 	}
 
+	//Check for Port Errors after PR complete
+	result = read_port_error(syfs_errpath, syfs_path, &porterr);
```
luebbers: As above.

libopae/src/reconf.c
```diff
@@ -408,6 +431,16 @@ fpga_result __FPGA_API__ fpgaReconfigureSlot(fpga_handle fpga,
...
+	if (result != FPGA_OK) {
+		FPGA_ERR("Failed to get port errors");
+	}
+	printf("Port Error: %" PRId64, porterr);
+	if (porterr != 0) {
+		FPGA_ERR("Errors exsist after Attempted PR.");
```
luebbers: Capitalization and spelling ("... exist after attempted PR.").

libopae/src/reconf.c
```diff
@@ -289,11 +307,16 @@ fpga_result __FPGA_API__ fpgaReconfigureSlot(fpga_handle fpga,
...
+	// Clear port errors for enough time for a max of 64 trasactions
+	int i;
+	for (i = 0; i < 2*MAX_TRANSACTIONS; ++i) {
+		result = clear_port_errors(fpga);
+		if (result != FPGA_OK) {
+			FPGA_ERR("Failed to clear port errors.");
+			goto out_unlock; // Why don't we exit here?
+		}
 	}
+	printf("Finished Clearing Errors");
```
luebbers: Please avoid using `printf()` in library functions. Users may not want to get the message in their output. Is this debug output (`FPGA_DBG()`), or is this something you want to relay as an intermediate state to the user?

libopae/src/reconf.c
```diff
@@ -408,6 +431,16 @@ fpga_result __FPGA_API__ fpgaReconfigureSlot(fpga_handle fpga,
 		result = FPGA_RECONF_ERROR;
 	}
 
+	//Check for Port Errors after PR complete
+	result = read_port_error(syfs_errpath, syfs_path, &porterr);
+	if (result != FPGA_OK) {
+		FPGA_ERR("Failed to get port errors");
+	}
+	printf("Port Error: %" PRId64, porterr);
```
luebbers: See above.

shivarao:
@luebbers this is a change that is needed due to a bug fix we made in the FIM for DCP. We had a bug where doing a PR while an application is running could crash the system.

The change on the FIM was to make it so that you can't clear the port errors while PR is running. Due to this change, you may need to clear the port errors up to 64 times to actually clear them. Lee Ping Chua has more details about the change and why they had to do it that way.

libopae/src/reconf.c
buie1: For reading the port error I've abstracted it out so that PR can report if an error exists after a PR is completed/ before unlocking the fpga

libopae/src/reconf.c
buie1: fixed.

libopae/src/reconf.c
buie1: thanks for this suggestion

buie1:
@luebbers Thanks for the feedback. From the behavior I've seen I believe the `ioctl` command works as expected.  The reason for iterating through attempting to clear comes from the fact that when trying to PR while an application is running for a different AFU we have a state where the FPGA is busy.  I didnt want to create a loop such that: `while result == FPGA_BUSY: \n {do stuff} ` because if the application is continuing to run this loop will be running for an undefined amount of time.
@anandaravuri @r-rojo Can you guys weigh in on this PR. It's a critical issue for our release.

luebbers:
We're trying to address this in two ways:

* Acquire an AFU resource before PRing it (makes sure noone else can talk to it); return BUSY if it's already acquired
* Add a "--force" flag to fpgainfo to clear errors up to 64 times

There'll be separate pull requests for this.

buie1:
Getting addressed in https://github.com/OPAE/opae-sdk/pull/329

PR #311, opened 2018-04-09
make numa support in cmdq3 optional in cmake

r-rojo:
The CMakeLists.txt for fpgadiag (and cmdq3) currently doesn't use a CMake
package finder to find numa and instead only looks for the library.
If it finds the library but the development files (headers) aren't present
on the system, the build can run into issues trying to include numa.h.
The PR makes it a CMake option to turn on numa support in cmdq3.

_TODO: Use (and maybe develop) a numa package finder to check for numa.h_


buie1:
bitmoji

PR #312, opened 2018-04-10
Automatically add platform releases to the platform DB search path

michael-adler:
- Look in ${OPAE_PLATFORM_ROOT}/hw/lib/platform for JSON databases before searching
  the global OPAE SDK database.  This allows a release to define the properties of
  its platform.
- Generate Verilog PLATFORM_CLASS_NAME macros to enable platform-specific compilation.
- Add platform-level syntax to define Verilog macros.  This will be used to define
  global properties, such as the FPGA family.


michael-adler:
This does not affect ASE. It adds the ability for a platform release to define the JSON platform database as part of the release. It also adds the ability to define some Verilog preprocessor macros to describe the platform. It allows us and others to define new systems without having to store every platform db in OPAE SDK.

ASE tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2310033

PR #314, opened 2018-04-11
Update platform db search path when --lib is passed to afu_synth_setup

michael-adler:
Now that platform_db libraries may be in release hw/lib trees, afu_platform_config
must be able to find the library.  Normally it uses $OPAE_PLATFORM_ROOT.  When
the root is not defined but --lib is passed to afu_synth_setup the db search
path must be updated.

### DIFF
--- a/platforms/scripts/afu_synth_setup
+++ b/platforms/scripts/afu_synth_setup
@@ -63,7 +63,7 @@ def errorExit(msg):
 #        starting with the discrete platform's 1.1 release.  The
 #        hw/lib directory is ${OPAE_PLATFORM_ROOT}/hw/lib.
 #
-def getHWLibPath(args):
+def getHWLibPath():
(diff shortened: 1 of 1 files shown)


lzhan55:
Does that mean that the 'platform' folder could be installed in a folder of OPAE such as /usr/lib/platform or /var/lib/platform ?

michael-adler:
The main one already is installed with OPAE in share/opae/platform. These commits add automatic searches for databases stored with board releases, in addition to the search of the OPAE platform tree.

lzhan55:
The board release doesn't have OPAE in it, right?

michael-adler:
I'm not sure how to answer that. The board release ships with OPAE RPMs, but OPAE can be installed from source too. The board release depends on some installation of OPAE. The two are separate.

PR #315, opened 2018-04-11
Database files moved to HW releases.  No longer needed here.

michael-adler:


lzhan55:
Are there any ASE DCP tests using those files?

michael-adler:
They won't be soon. I'll actually commit as soon as I'm sure of that. This is in preparation.

What's going on with Travis complaining about files not touched in this commit?

chipb:
Not really sure on Travis complaining about unrelated files though...

PR #316, opened 2018-04-11
Use flat Doxygen builds

michael-adler:
Building 4K subdirectories per category is very slow.


michael-adler:
How do you feel about this? The 4K subdirectory mechanism seems to be overkill for the amount of documentation we are generating. This change makes builds much, much faster.

asgibbons:
I wondered what happened - my build times doubled. :-(

PR #324, opened 2018-04-13
Remove PLATFORM_NAME macro creation in PIM's Quartus script

michael-adler:
Replaced with macros defined in platform_db entries.


michael-adler:
No impact on ASE. MPF is the only current consumer of these macros and I updated it. The macros were a hack, using Quartus versions to figure out release names. The new method adds explicit names from platform_db entries tied to releases -- already present. A side benefit: the new method gives the same result in simulation and synthesis. The old method "worked" only in synthesis.

Tests pass (OpenCL failures are unrelated): http://sw-pert.altera.com/pert/pert.php?test_run_id=2317510

PR #328, opened 2018-04-14
fpgainfo: add --force option

luebbers:
Specifying --force will try clearing errors up to 64
times. This may be necessary on certain error
conditions to flush in-flight transactions.


tools/base/fpgainfo/fpgaerr.py
```diff
@@ -285,6 +286,13 @@ def args(self, parser):
                             default='fme',
                             help='specify what kind of errors to operate on')
 
+        # FIXME: should get a '-f' option later
```
shivarao: Seems like a easy FIXME to knock out now

shivarao:
This seems ok except that we'll need to document the --force flag now.

What's the risk of making the --force behavior the default behavior? It should result in one clear on systems that don't require multiple clears.

tools/base/fpgainfo/fpgaerr.py
luebbers: "-f" is currently used for the PCI function; as part of consolidating all tools, that should be turned into "-F", but that would change the current usage; we'll need to coordinate that carefully.

luebbers:
Having to clear the same error multiple times does not seem to be a feature to me - I'd rather go back and reexamine the clearing flow so that it's not necessary.

luebbers:
If it's any help, the `--force` flag should really rarely be necessary, since we're catching the error condition early now.

I think having the default behavior not clear rare (and avoidable) error conditions helps guide the users towards understanding what's causing those conditions (i.e. uncoordinated PR). If we blanket-hide the symptoms, we end up with a less clean solution.

PR #329, opened 2018-04-14
Add force option for fpgaReconfigureSlot

r-rojo:
This changes behavior of fpgaReconfigureSlot to try to open
an accelerator resource on the given slot and bail out if it can't.
This is to mitigate situations when a slot is reconfigured while
the accelerator is in use by another process.

This behavior can be skipped if the FPGA_RECONF_FORCE is given.
This also adds the corresponding command line option to fpgaconf


shivarao:
What happens when SR-IOV is enabled?

My understanding was that there was no way for the host to detect if the accelerator resource is in use if it's assigned to the VF. We don't currently enable PR in this case but we are looking to do so.

libopae/src/reconf.c
```diff
@@ -295,6 +366,13 @@ fpga_result __FPGA_API__ fpgaReconfigureSlot(fpga_handle fpga,
 		FPGA_ERR("Failed to clear port errors.");
 	}
 
+	if (!(flags & FPGA_RECONF_FORCE)) {
```
buie1: Here you are using the `flags` param but on line 344 its included in the `UNUSED_PARAM` macro. I think we should delete this correct?

libopae/src/reconf.c
```diff
@@ -295,6 +366,13 @@ fpga_result __FPGA_API__ fpgaReconfigureSlot(fpga_handle fpga,
...
+		result = port_busy(fpga);
+		if (result != FPGA_OK) {
+			FPGA_ERR("Port device(s) may be in use.");
```
buie1: If the Port is in use we should not allow reconfiguration and `goto out_unlock`
Also wouldn't it make more since to check if the Port is busy before trying to clear port_errors.

tools/base/fpgaconf/fpgaconf.c
```diff
@@ -237,6 +240,7 @@ int parse_args(int argc, char *argv[])
 		{"device",        required_argument, NULL, 'd'},
 		{"function",      required_argument, NULL, 'f'},
 		{"socket",        required_argument, NULL, 's'},
+		{"force",         no_argument,       NULL, 0xf},
```
buie1: The only concern I have with this is I feel that this should be the default behavior when trying to do a PR.  For fpgainfo i feel there could be an argument made for an additional flag but when doing reconfiguration we should report if the fpga is already in use to notify the user.

libopae/src/reconf.c
r-rojo: Yes. I will make the appropriate changes.

libopae/src/reconf.c
r-rojo: * Yes, we should `goto out_unlock;`
* I was going to do this (check before clearing errors at first) but changed my mind for some reason.
  I can't think of why I changed my mind so I'll put it back in the order you recommend.

tools/base/fpgaconf/fpgaconf.c
r-rojo: If we default to `force` during PR, then we get back to what started this PR (pull request).
Setting the `force` flag will bypass the check for the resource-in-use.

luebbers: The default behavior should be the one that doesn't cause any disruption or error. If an accelerator is in use, PR would disrupt the owning application (and possibly cause a undesired port error condition); those consequences should be accepted explicitly (by providing "-f"), not implicitly.

r-rojo:
So, in essence, the change in behavior this change makes is:
* Default: Check for port busy and error out if it is busy
* --force: Do a PR regardless if the port is busy `[don't check for port being busy and just do the PR]`

Isn't that what we want?

libopae/src/reconf.c
```diff
@@ -134,6 +135,76 @@ static fpga_result validate_bitstream(fpga_handle handle,
...
+	//       We only want to query for accelerators for the
+	//       slot being reconfigured
+	result = fpgaEnumerate(&props, 1, &token, 1, &matches);
+	if (result != FPGA_OK) {
+		FPGA_ERR("Error enumerating while reconfiguring");
+		goto free_props;
+	}
+
+	if (0 == matches) {
+		FPGA_ERR("No accelerators found while reconfiguring.");
```
luebbers: And I think we usually don't use a `.` at the end of `FPGA_ERR()` messages, but I'd need to check other code.

libopae/src/reconf.c
```diff
@@ -134,6 +135,76 @@ static fpga_result validate_bitstream(fpga_handle handle,
...
+		FPGA_ERR("Error closing accelerator.");
+		goto destroy_token;
+	}
+
+
+destroy_token:
+	result = fpgaDestroyToken(&token);
+	if (result != FPGA_OK) {
+		FPGA_ERR("Error destroying a token.");
+	}
```
luebbers: Don't need the `}` (and its sibling `{`, of course).

libopae/src/reconf.c
```diff
@@ -289,6 +358,16 @@ fpga_result __FPGA_API__ fpgaReconfigureSlot(fpga_handle fpga,
 		goto out_unlock;
 	}
 
+	// error out if "force" flag is NOT indicated
+	// and the resource is in use
+	if (!(flags & FPGA_RECONF_FORCE)) {
+		result = port_busy(fpga);
+		if (result != FPGA_OK) {
+			FPGA_ERR("Port device(s) may be in use.");
```
luebbers: Why "may" and "device(s)"? Is there a chance for a false positive?

tools/base/fpgaconf/fpgaconf.c
```diff
@@ -268,6 +272,10 @@ int parse_args(int argc, char *argv[])
 			config.dry_run = true;
 			break;
 
+		case 0xf:    /* force */
```
luebbers: Sneaky. ;)

libopae/src/reconf.c
```diff
@@ -170,33 +169,26 @@ static fpga_result port_busy(fpga_handle handle)
...
 	if (result != FPGA_OK) {
-		FPGA_ERR("Error enumerating while reconfiguring");
+		FPGA_ERR("Error enumerating for accelerator to reconfigure");
 		goto free_props;
 	}
 
-	if (0 == matches) {
-		FPGA_ERR("No accelerators found while reconfiguring.");
+	if (matches == 0) {
+		FPGA_ERR("No accelerator found to reconfigure");
```
luebbers: Just wondering - if we can't find an accelerator whose parent is our FME, doesn't that mean we should go ahead with the reconfiguration? After all, if _we_ can't find the accelerator, probably no one else has it.

r-rojo: Not sure. What happens in the case of virtualization?

luebbers: Good point. If we pass through the VF to a VM, we won't find an accelerator either. Would we want to fail reconfiguration in that case, if "--force" is not given?

r-rojo: Yes. This check is only performed if `--force` (or `FPGA_RECONF_FORCE`) is not in the flags.

buie1:
@r-rojo @shivarao @luebbers Doing some final tests on this PR this afternoon. Will report back soon. So far the changes look good

libopae/src/reconf.c
```diff
@@ -134,6 +135,68 @@ static fpga_result validate_bitstream(fpga_handle handle,
...
+	}
+
+	result = fpgaOpen(token, accel, 0);
+	if (result != FPGA_OK) {
+		FPGA_ERR("Could not open accelerator for given slot");
+		goto destroy_token;
+	}
+
+destroy_token:
+	result = fpgaDestroyToken(&token);
```
luebbers: This overwrites the result of `fpgaOpen()` in line 182, which may mask a failure and return FPGA_OK to the caller.

PR #332, opened 2018-04-16
remove parameter name from property_map::set

r-rojo:
The parameter is unused in the implementation of the function

### DIFF
--- a/tools/extra/libopae++/property_map.h
+++ b/tools/extra/libopae++/property_map.h
@@ -78,7 +78,7 @@ public:
             {
                 value = opt->value<T>();
             }
-            catch(const intel::utils::any_value_cast_error & bac)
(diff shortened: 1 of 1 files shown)


luebbers:
Should we run checkers or compiler flags to highlight this?

r-rojo:
So I tried an experiment where I compiled a test .cpp file using the same compile flags we're using for our build. This is the contents of the test:

```C++
#include "any_value.h"

  int function(int arg1, int arg2)
  {
    try {
      return arg1 * arg2;
    } catch (const intel::utils::any_value_cast_error &bac) {
      return 0;
    }
  }
```

When I add an unused parameter to the function signature:
```C++
int function(int arg1, int arg2, int arg3)
```

It does error out. So it seems like unused parameters in `catch` blocks aren't flagged by the compiler.

PR #335, opened 2018-04-16
Use pre-determined 'cmake' executable path

chipb:


chipb:
Relatively minor fix, but I noticed we have some stray path search leaks. Mechanized with:
```sh
git grep -l 'COMMAND[[:space:]\n]\+cmake' -- '*[Cc][Mm]ake*' | xargs sed -i 's,COMMAND cmake,COMMAND ${CMAKE_COMMAND},g'
```

chipb:
*ahem* I guess I should pay attention to the contribution guide; I just added the S-o-B to the commit.

PR #339, opened 2018-04-18
c++utils: add double/float support in opt parsing

r-rojo:
Add template specializations for options that are or type `float` and `double`
when parsing from command line arguments.


r-rojo:
So this will throw an exception if it can't parse. We have three options:
* do a try/catch and print out a meaningful error message, return 0.0
* do a try/catch and print out a meaningful error message, rethrow
* keep it as is and let the exception be thrown (which may not have a meaningful message)

What do you think we should do?

tswhison:
I like print and re-throw.

PR #351, opened 2018-04-19
drv doc: add note about binutils >= 2.25

tswhison:
as a requirement for 512-byte PR.

### DIFF
--- a/doc/src/drv_arch/drv_arch.md
+++ b/doc/src/drv_arch/drv_arch.md
@@ -71,6 +71,12 @@ comparing the interface ID noted in the AF header against the interface ID
 exposed by the FME through sysfs. This check is usually done by
 user-space before calling the reconfiguration IOCTL.
 
+.. note::
+
+```
+    Platforms that support 512-byte Partial Reconfiguration require
+    binutils >= version 2.25.
+```
 
 .. note::
 


doc/src/drv_arch/drv_arch.md
```diff
@@ -71,6 +71,13 @@ the AF.
 
 In all other cases PR fails and may cause system instability.
 
+.. note::
+
+```
+    Platforms that support 512-byte Partial Reconfiguration require
```
r-rojo: byte or bit? Doesn't PR use 512-bit MMIO (through a special AVX instruction)? Is this referring to that?

tswhison: I remember Enno telling me byte before.. Now that I look at the CSR spreadsheet again, there are eight 64-bit FME_PR_DATA registers (total of 512 bits).  Seems that it should be bit.  @luebbers, am I misremembering this?

r-rojo: I'm pretty sure PR uses 512-bit MMIO. The instructions it uses are:
* vmovdqu64
* vmovntdq

You can find both of these instructions in the AVX-512 wikipedia page:
https://en.wikipedia.org/wiki/AVX-512

tswhison: I'm familiar with the instr's.  I must have been thinking 64-byte vs. 64-bit write.  Will change it.

PR #345, opened 2018-04-19
Add mechanisms for selecting RTL sources based on FPGA family and Quartus versions

michael-adler:
Some sources, such as Qsys-generated RTL, are specific to the FPGA family or major version of Quartus.

- Add an FPGA family property to the platform database.
- Add automatically generated environment variables to rtl_src_config that can be used in source paths.

PR #352, opened 2018-04-19
Add mechanisms for selecting RTL sources based on FPGA family and Quartus versions (#345)

michael-adler:
- The FPGA family tag may be stored in the platform db and extracted with a new script: afu_platform_info.
- rtl_src_config automatically defines the environment variables OPAE_PLATFORM_FPGA_FAMILY and QUARTUS_VERSION_MAJOR when it detects a reference to the tags in the list of sources.


michael-adler:
Having this in the release will make it far easier to write AFUs with Qsys components that target different FPGA families.

PR #353, opened 2018-04-19
opae: Minimum user clock frequency reduced to 25 MHz

anandaravuri:
Changes:
User can set user clock frequency   from 25 MHz to 1200 MHz.


tools/extra/userclk/main.c
```diff
@@ -187,8 +187,7 @@ int main( int argc, char** argv )
 		goto out_destroy_prop;
 	}
 
-	if (userclkCmdLine.freq_high > 0 &&
-		userclkCmdLine.freq_low  > 0) {
```
r-rojo: is there no longer a need to check the low frequency?

anandaravuri: yes
User clock high frequency index used to get values from table

PR #355, opened 2018-04-19
Add fpgaGetPropertiesFromHandle

asgibbons:
Function needed to obtain properties for an open fpga_handle.


common/include/opae/properties.h
```diff
@@ -54,6 +54,26 @@
...
+/**
+ * Create a fpga_properties object
+ *
+ * Initializes the memory pointed at by `prop` to represent a properties
+ * object, and populates it with the properties of the resource referred to by
+ * `handle`. Individual properties can then be queried using fpgaPropertiesGet*()
+ * accessor functions.
+ *
+ * @note fpgaGetPropertiesFromHandle() will allocate memory for the created properties
+ * object returned in `prop`. It is the responsibility of the using application
```
r-rojo: Should `using application` be rephrased to `caller`?

PR #358, opened 2018-04-19
libopae: fix fpgaEnumerate() doxygen bug

tswhison:
Fix the doxygen for parameter 'filters'. The docs previously mentioned
that passing NULL for 'filters' would return all fpga objects. The
correct query-all method is to pass an initialized but empty
fpga_properties object.

### DIFF
--- a/common/include/opae/enum.h
+++ b/common/include/opae/enum.h
@@ -76,7 +76,7 @@ extern "C" {
  *                         without any filter criteria set) or pass a NULL
  *                         filters parameter with num_filters set to 0.
  * @param[in] num_filters  Number of entries in the `filters` array, or 0 to
- *                         match all FPGA resources.
+ *                         match all FPGA resources when `filters` is NULL.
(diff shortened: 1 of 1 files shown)


tswhison:
Many moons ago.  I'm not certain that the API ever worked that way (by passing NULL).

r-rojo:
It does work with NULL filters as long as the number of filters is also 0.
In `fpgaEnumerate`:
https://github.com/OPAE/opae-sdk/blob/4bd2663284ff262f2942d7002f73a2cb7b8fcca9/libopae/src/enum.c#L637-L645

If `filters` is NULL and `num_filters` is > 0, we get past this check and eventually call `matches_filters` which checks `num_filter` being 0:
https://github.com/OPAE/opae-sdk/blob/4bd2663284ff262f2942d7002f73a2cb7b8fcca9/libopae/src/enum.c#L303-L304

tswhison:
Thanks for reminding me.  Do we want to abandon this then?

PR #361, opened 2018-04-20
Small change to PIM's hw/lib directory check makes scripting easier

michael-adler:


michael-adler:
Not a very exciting change, but useful internally for BBS build scripts and no impact on external users.

Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2332454

PR #362, opened 2018-04-23
Pass the local memory bank number as a normal signal.

michael-adler:
Bank number is used only for debugging, but now that it is a wire instead of a parameter
it should be passed from instance to instance using combination assignments.

### DIFF
--- a/ase/rtl/ase_top.sv
+++ b/ase/rtl/ase_top.sv
@@ -66,8 +66,6 @@ module ase_top();
    logic [1:0] pck_cp2af_pwrState;   // CCI-P AFU Power State
    logic       pck_cp2af_error;      // CCI-P Protocol Error Detected
 
-   genvar b;
-
 `ifdef PLATFORM_PROVIDES_LOCAL_MEMORY
  // Now many memory banks?
  `ifdef AFU_TOP_REQUIRES_LOCAL_MEMORY_AVALON_MM_LEGACY_WIRES_2BANK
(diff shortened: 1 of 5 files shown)


michael-adler:
Tests pass (OpenCL broken for unrelated reasons): http://sw-pert.altera.com/pert/pert.php?test_run_id=2334413

ase/rtl/ase_top.sv
```diff
@@ -237,6 +228,10 @@ module ase_top();
             .ddr4b_avmm_byteenable                 (ddr4[b+1].byteenable),
             .ddr4b_avmm_clk_clk                    (ddr4_avmm_clk[b+1])
          );
+
+         // Mostly used for debugging
+         assign ddr4[b].bank_number = b;
```
lzhan55: @michael-adler You moved the "assign ddr4[b].bank_number = b" inside the PLATFORM_PROVIDES_LOCAL_MEMORY. Does that mean that AFU_TOP_REQUIRES_LOCAL_MEMORY_AVALON_MM_LEGACY_WIRES_2BANK platform doesn't need this assignment?

platforms/platform_if/rtl/platform_shims/platform_shim_ccip_std_afu.sv
```diff
@@ -127,18 +127,9 @@ module platform_shim_ccip_std_afu
     logic afu_local_mem_clk[NUM_LOCAL_MEM_BANKS];
     logic afu_local_mem_reset[NUM_LOCAL_MEM_BANKS];
 
-    avalon_mem_if#(.ENABLE_LOG(0), .NUM_BANKS(NUM_LOCAL_MEM_BANKS))
+    avalon_mem_if#(.ENABLE_LOG(1), .NUM_BANKS(NUM_LOCAL_MEM_BANKS))
```
lzhan55: @michael-adler Why did you remove below code from platforms/platform_if/rtl/platform_shims/platform_shim_ccip_std_afu.sv ? Was ddr4[].bank_number signal connected to this afu_loca_mem[].bank_number somewhere? 

133   -    genvar b;    
134   -    generate    
135   -        for (b = 0; b < NUM_LOCAL_MEM_BANKS; b = b + 1)  
        ....

michael-adler: The loop isn't needed anymore because the bank_number signal is connected properly in platform_shim_avalon_mem_if().

ase/rtl/ase_top.sv
michael-adler: Correct.

PR #363, opened 2018-04-23
Add NUMA awareness to OPAE

asgibbons:
Add NUMA awareness to OPAE, and extract and maintain the PCIe segment as part of the PCIe address.  NUMA will not be enabled by default.

This change merely adds the indication that NUMA is available to the properties struct.  Using this feature requires package numactl-devel be installed in order to find numa.h and libnuma.so.


common/include/opae/properties.h
```diff
@@ -495,7 +495,7 @@ fpga_result fpgaPropertiesSetModel(fpga_properties prop,
  * @note This API is not currently supported.
  */
 fpga_result fpgaPropertiesGetLocalMemorySize(const fpga_properties prop,
-					     uint64_t *lms);
+		uint64_t *lms);
```
luebbers: Unintentional change of indentation? (more below)

libopae/CMakeLists.txt
```diff
@@ -69,6 +72,13 @@ target_link_libraries(opae-c
   ${libuuid_LIBRARIES}
   ${CMAKE_THREAD_LIBS_INIT})
 
+if (USE_NUMA)
+    find_package(NUMA REQUIRED)
```
luebbers: If numactl is not present, will this fail the entire build?

libopae/src/sysfs.c
```diff
@@ -632,18 +626,22 @@ fpga_result sysfs_bdf_from_path(const char *sysfspath, int *b, int *d, int *f)
...
+	*f = (int)strtoul(p + 6, NULL, 16);
 	*(p + 5) = 0;
 
-	*d = (int) strtoul(p+3, NULL, 16);
+	*d = (int)strtoul(p + 3, NULL, 16);
 	*(p + 2) = 0;
 
-	*b = (int) strtoul(p, NULL, 16);
+	*b = (int)strtoul(p, NULL, 16);
+	*(p - 1) = 0;
```
luebbers: Can we add a check that `p` actually points far enough into the buffer that we don't walk into previous values on the stack (same for line 638 below)?

libopae/CMakeLists.txt
```diff
@@ -69,6 +72,13 @@ target_link_libraries(opae-c
...
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -lnuma -I${NUMA_INCLUDE_DIR}")
```
r-rojo: Can't we use `include_directories(${NUMA_INCLUDE_DIR})` instead?
Also, line 79 is redundant with `-lnuma` in this line

r-rojo:
Does this change set the stage for supporting numa properties?

Also, there seem to be a lot of unrelated changes (indentation or spacing) that add noise to the PR and may decrease the level of attention given by someone reviewing the PR.

libopae/CMakeLists.txt
asgibbons: Yes.  If you have set the option to use NUMA and it's not present, the build will fail.  Is this an issue?

libopae/CMakeLists.txt
asgibbons: Yes.  Done.

common/include/opae/properties.h
asgibbons: Not unintentional.  I'm using AStyle to format my source changes, which will automate formatting and make things consistent.  We should discuss at some time, because manually "fixing" all the formatting cruft is extremely time-consuming.

libopae/src/sysfs.c
asgibbons: I *can* add a check, but the "p += 6" above pretty much ensures it'll be OK, since the strrchr() found a '/' at 'p'.

asgibbons:
@r-rojo Yes, this sets the stage for NUMA support.

Perhaps we should publish editor settings or AStyle / indent settings that we could use to ensure proper formatting?  I'm using AStyle to filter with the following command-line:

`astyle --style=linux --indent=force-tab=8 --align-pointer=name --max-code-length=90 --break-after-logical -p -H -U`

libopae/CMakeLists.txt
luebbers: Ah, no. If USE_NUMA is not set by default, and someone sets it, no NUMA package should fail the build.

buie1:
@r-rojo @asgibbons 
Can we revisit this PR?

r-rojo:
Overall, I think I like what this brings to the OPAE API.
I have a few comments/questions.
* What is the expected use of this fpgaAdviseDmaBuffer API?
  * Is this expected to be used with fpgaPrepareBuffer using the PREALLOCATED flag or instead of the fpgaPrepareBuffer API?
* Can the reformatting of non-relevant code be done  separately?

deepakcu84:
I agree. Please use a separate commit or pull request for formatting changes. 

Does rest of the OPAE source (i.e untouched by this PR) use astyle formatting? I'd be concerned if only formatting changes apply only to files in this PR. If we intend to reformat the entire source to conform to astyle, we should do that in a separate PR.

common/include/opae/buffer.h
```diff
@@ -129,6 +129,32 @@ fpga_result fpgaReleaseBuffer(fpga_handle handle, uint64_t wsid);
...
+ * @param[in]  buf           Buffer intended to be used for DMA transactions
+ * @param[in]  len           The length of the buffer to prepare
+ * @param[in]  bind_thread   If true, the calling thread will have its CPU affinity
+ *                           set to the processors local to the FPGA's NUMA node.
+ *                           If false, only the buffer is moved.
+ * @returns FPGA_OK on success. FPGA_INVALID_PARAM if invalid parameters were
+ * provided, or if the parameter combination is not valid. FPGA_EXCEPTION if an
+ * internal exception occurred while trying to access the handle.
+ */
+fpga_result fpgaAdviseDmaBuffer(fpga_handle handle, void *buf, uint64_t len,
```
deepakcu84: I don't understand this API. Does this API allocate the buffer or simply relocate it after allocation?

Do you expect the application to first allocate the buffer using fpgaPrepareBuffer and then bind it to NUMA node using fpgaAdviseDmaBuffer? If it is the latter, why do you need length for binding? 

Please clarify how this API is expected to be used in the description.

libopae/src/numa.c
```diff
@@ -0,0 +1,212 @@
...
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/ioctl.h>
+#include <sys/mman.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <unistd.h>
+
+fpga_result save_and_bind(struct _fpga_handle *_handle, bool bind, bool save_state)
```
deepakcu84: Will be useful to add some comments here. Example: Saves the thread state and binds it to NUMA node"

libopae/src/numa.c
```diff
@@ -0,0 +1,212 @@
...
+		_handle->numa->saved.membind_mask = numa_get_membind();
+		_handle->numa->saved.runnode_mask = numa_get_run_node_mask();
+	}
+
+	numa_bind(_handle->numa->fpgaNodeMask);
+#endif
+	return res;
+}
+
+fpga_result restore_and_unbind(struct _fpga_handle *_handle, bool bind, bool restore_state)
```
deepakcu84: I think save_and_bind, restore_and_bind and move_memory_to_noad should be declared static unless you expect users to call them directly.

libopae/src/properties.c
```diff
@@ -492,7 +521,53 @@ fpgaPropertiesSetObjectType(fpga_properties prop, fpga_objtype objtype)
...
+		result = FPGA_NOT_FOUND;
+	}
+
+	err = pthread_mutex_unlock(&_prop->lock);
+	if (err)
+		FPGA_ERR("pthread_mutex_unlock() failed: %s", strerror(err));
+	return result;
+}
+
+fpga_result __FPGA_API__ fpgaPropertiesSetSegment(fpga_properties prop,
```
deepakcu84: where is the declaration for fpgaPropertiesSetSegment and fpgaPropertiesGetSegment

libopae/src/properties.c
```diff
@@ -1309,3 +1379,47 @@ fpgaPropertiesSetObjectID(fpga_properties prop, uint64_t object_id)
...
+		return result;
+
+	_prop->numa_node = numa;
+	SET_FIELD_VALID(_prop, FPGA_PROPERTY_NUMANODE);
+
+	err = pthread_mutex_unlock(&_prop->lock);
+	if (err)
+		FPGA_ERR("pthread_mutex_unlock() failed: %s", strerror(err));
+	return result;
+}
```
deepakcu84: Have you verified these APIs on a NUMA system?

asgibbons:
Will open another pull request for this once cleaned up.

PR #368, opened 2018-04-24
reconf: Fix closing of resource (if opened)

r-rojo:
* Set accell to NULL
* Move closing the accel handle to after `out_unlock` label
* Condition closing the accel handle only if it is non-NULL


buie1:
Did we get this merged into master as well?
@r-rojo @luebbers

PR #370, opened 2018-04-25
c++core: Make handle::close a public function

r-rojo:
Also cleanup some doxygen markup comments

### DIFF
--- a/common/include/opae/cxx/core/handle.h
+++ b/common/include/opae/cxx/core/handle.h
@@ -117,6 +117,9 @@ class handle {
    *
    * @param[in] flags The flags parameter to fpgaOpen().
    *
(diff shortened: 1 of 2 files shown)


r-rojo:
@michael-adler, can you look over this?

michael-adler:
Should you note in the documentation for close() that the destructor also calls close() or do you want to push people toward explicitly closing?

r-rojo:
@michael-adler , thanks for the suggestion.
I don't want to push people toward explicitly closing but I would like this API to be public to allow users of the API to explicitly close without having to rely on the desctructor being called (especially since instances of handle will almost always be shared pointers).
I will update the documentation.

PR #373, opened 2018-04-26
Suppress full power output for DCP

buie1:
We want to handle the power command better for DCP specific use case.
This adds `device_id` as an property in sysfsinfo as well as makes the `DCP_ID` a part of fpga_common.py


buie1:
Bump @r-rojo @OPAE/tools

PR #380, opened 2018-04-27
Prefault page tables for a mapping.

asgibbons:
Add the MAP_POPULATE flag to the mmap calls.  This will pre-fault the page tables for the mapping, giving the proper return value for an out-of-memory condition.  HSD: https://hsdes.intel.com/appstore/article/#/2202798208

### DIFF
--- a/libopae/src/buffer.c
+++ b/libopae/src/buffer.c
@@ -85,11 +85,11 @@ static fpga_result buffer_allocate(void **addr, uint64_t len, int flags)
 	   For buffer > 2M, use 1G-hugepage to ensure pages are
 	   contiguous */
 	if (len > 2 * MB)
-		addr_local = mmap(ADDR, len, PROTECTION, FLAGS_1G, 0, 0);
+		addr_local = mmap(ADDR, len, PROTECTION, FLAGS_1G | MAP_POPULATE, 0, 0);
 	else if (len > 4 * KB)
-		addr_local = mmap(ADDR, len, PROTECTION, FLAGS_2M, 0, 0);
+		addr_local = mmap(ADDR, len, PROTECTION, FLAGS_2M | MAP_POPULATE, 0, 0);
 	else
-		addr_local = mmap(ADDR, len, PROTECTION, FLAGS_4K, 0, 0);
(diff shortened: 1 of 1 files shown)


r-rojo:
Just as an FYI, I added the `MAP_POPULATE` flag to the mmap call like you had previously recommended and the mmap call didn't fail as expected. The mmap call still returned an address that was used in the ioctl which resulted in the `Bad Address` error.
As part of my experiments, I added a memset after the mmap succeeded. The call to memset with this address triggered a SIGBUS signal to be raised.

asgibbons:
@r-rojo The behavior of MAP_POPULATE for huge pages is different based on kernel version.  I have an older kernel and it does give me the error on mmap().

Can you try an experiment by adding a call to madvise(void*,len,MADV_WILLNEED) after the mmap() and see whether it gives you EIO or ENOMEM?  If so, we could use that to trigger the out-of-memory condition.

r-rojo:
I did what you suggested and `madvise` didn't error out. It always returned 0.
So for this test, I copied the relevant code into a standalone test (no OPAE involved, no ioctl). My test is as follows:
* call buffer_allocate to get a buffer address
* call memset on the buffer
* call buffer_free on the address.

This test was conducted with the following system/test configuration:
* two socket (integrated xeon)
* redhat, x86_64, 4.7 kernel
* 0 2MB hugepages on node 0
* 1 2MB hugepages on node 1
* Use MAP_POPULATE in the mmap flags
 

These are the results:

Requested Buffer Size | Numa Bind| madvise result | test pass
-------------| ------------|------------------|----------
4097           |  node 0     | 0 (no error)         | FAIL - Bus Error on calling `memset`
4097           |  node 1     | 0 (no error)         | PASS
4097           | none         | 0 (no error)         | PASS

tswhison:
@r-rojo Am I reading your test results correctly, that even with MAP_POPULATE, we don't get the expected result in all cases?

r-rojo:
@tswhison , Yes, even with the MAP_POPULATE flag we don't see expected results. I don't believe this change will much more to address the issue in question

asgibbons:
Closing as it's almost a year old.

PR #382, opened 2018-04-27
Add gtapi_sys for testing REL version OPAE

rpSTTD:
Add  gtapi_sys test suite to test  RELEASE version OPAE SDK.


tests/sys/.cproject
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
ibarraza: Are these supposed to have been added?
tests/sys/.cproject
tests/sys/.project
tests/sys/.settings.language.settings.xml

r-rojo:
Doing a quick cursory glance, I notice that the copyright notice is inconsistent in several files. Also, new files have 2017 for the year. Shouldn't it be 2018 for new files?

tests/sys/.cproject
rpSTTD: No. I will remove it.

PR #387, opened 2018-05-01
Compile Qsys when configuring ASE simulation.

michael-adler:
This eliminates the last significant difference between the ASE configuration
provided in OPAE and the ASE configuration shipped with discrete cards.  It was
a source of confusion in documentation, since we generally advised the use of
afu_sim_setup, which invokes generate_ase_environment.py.  For AFUs with Qsys,
we suggested a different script.  All shipped AFUs can now use the same
afu_sim_setup for configuration.

### DIFF
--- a/ase/scripts/generate_ase_environment.py
+++ b/ase/scripts/generate_ase_environment.py
@@ -59,6 +59,7 @@ from collections import defaultdict
 import fnmatch
 import json
 import subprocess
+from sets import Set
 
 reload(sys)
 sys.setdefaultencoding('utf8')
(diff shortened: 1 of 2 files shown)


ase/scripts/generate_ase_environment.py
```diff
@@ -212,14 +213,14 @@ def config_sources(fd, filelist):
...
-                s = spl[0] + ' ' + '\ '.join(spl[1:])
+                s = spl[0] + ' ' + '\\ '.join(spl[1:])
             vlog_srcs.append(s)
             vlog_found = True
         else:
             # Convert extensions to lower case for comparison
             sl = s.lower()
             # Escape spaces in pathnames
-            s = s.replace(' ', '\ ')
+            s = s.replace(' ', '\\ ')
```
abelardojarab: what is the purpose of this?

does Python provide built-in functions to manipulate pathnames instead of manipulating filenames as strings?

michael-adler: Some in os.path, but not to do either of the operations here.

PR #388, opened 2018-05-02
 Compile Qsys when configuring ASE simulation.

michael-adler:
This eliminates the last significant difference between the ASE configuration
provided in OPAE and the ASE configuration shipped with discrete cards.  It was
a source of confusion in documentation, since we generally advised the use of
afu_sim_setup, which invokes generate_ase_environment.py.  For AFUs with Qsys,
we suggested a different script.  All shipped AFUs can now use the same
afu_sim_setup for configuration.

### DIFF
--- a/ase/scripts/generate_ase_environment.py
+++ b/ase/scripts/generate_ase_environment.py
@@ -59,6 +59,7 @@ from collections import defaultdict
 import fnmatch
 import json
 import subprocess
+from sets import Set
 
 reload(sys)
 sys.setdefaultencoding('utf8')
@@ -238,12 +239,16 @@ def config_sources(fd, filelist):
             if (sl[-5:] == '.json'):
                 json_srcs.append(s)
 
+    qsys_sim_files = config_qsys_sources(filelist, vlog_srcs)
+
     # List Verilog & SystemVerilog sources in a file
-    if (vlog_found):
+    if (vlog_found or qsys_sim_files):
         fd.write("DUT_VLOG_SRC_LIST = " + VLOG_FILE_LIST + " \n\n")
         with open(VLOG_FILE_LIST, "w") as f:
             for s in vlog_srcs:
                 f.write(s + "\n")
+            if (qsys_sim_files):
+                f.write("-F " + qsys_sim_files + "\n")
 
     # List VHDL sources in a file
     if (vhdl_found):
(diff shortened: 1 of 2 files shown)


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2351443

ase/scripts/generate_ase_environment.py
```diff
@@ -238,12 +240,16 @@ def config_sources(fd, filelist):
             if (sl[-5:] == '.json'):
                 json_srcs.append(s)
 
+    qsys_sim_files = config_qsys_sources(filelist, vlog_srcs)
```
lzhan55: Was Qsys from Quartus installation? Did it use the old version files in ASE simulation previously?

michael-adler: Qsys is part of Quartus.

I don't understand the 2nd question.

lzhan55: I just wonder how it worked before this change. ASE simulation did not compile Qsys before?

michael-adler: There was code in setup scripts just for Qsys in the internal intel-fpga-afu repository. Everything else has been moved out of that script and into afu_sim_setup/generate_ase_environment. This is the last piece.

There's another pull request on that repository, removing the code there. The tests were done with the two changes combined.

ase/scripts/generate_ase_environment.py
```diff
@@ -238,12 +240,16 @@ def config_sources(fd, filelist):
...
+    if (vlog_found or qsys_sim_files):
         fd.write("DUT_VLOG_SRC_LIST = " + VLOG_FILE_LIST + " \n\n")
         with open(VLOG_FILE_LIST, "w") as f:
             for s in vlog_srcs:
                 f.write(s + "\n")
+            if (qsys_sim_files):
+                f.write("-F " + qsys_sim_files + "\n")
 
     # List VHDL sources in a file
     if (vhdl_found):
```
deepakcu84: What about Qsys flow for VHDL?

ase/scripts/generate_ase_environment.py
```diff
@@ -282,6 +288,130 @@ def config_sources(fd, filelist):
...
+    qsys_srcs = []
+    ip_dirs = []
+    srcs = srcs.split('\n')
+    for s in srcs:
+        if (s):
+            # Record all build target directories
+            ip_dirs.append(os.path.splitext(s)[0])
+
+            # Collect all qsys files
+            if (s.lower()[-5:] == '.qsys'):
```
deepakcu84: what is -5? If you intended to use the length of string ".qsys", can you use the python equivalent of strlen

ase/scripts/generate_ase_environment.py
```diff
@@ -282,6 +288,130 @@ def config_sources(fd, filelist):
...
+                qsys_srcs.append(s)
+
+    # Any Qsys files found?
+    if (not qsys_srcs):
+        return None
+
+    # First step: copy the trees holding Qsys sources to a temporary tree
```
deepakcu84: Does the temporary tree get deleted after simulation? Also, make sure the temporary tree is cleaned up on error.

ase/scripts/generate_ase_environment.py
```diff
@@ -282,6 +288,130 @@ def config_sources(fd, filelist):
...
+        sim_files_found.add(os.path.basename(s))
+
+    # Find all generated Verilog/SystemVerilog sources
+    qsys_sim_files = 'qsys_sim_files.list'
+    with open(qsys_sim_files, "w") as f:
+        for d in ip_dirs_copy:
+            for dir, subdirs, files in os.walk(d):
+                for fn in files:
+                    if ((os.path.basename(dir) == 'synth') and
+                            (fn[-2:] == '.v' or fn[-3:] == '.sv')):
```
deepakcu84: Avoid magic numbers -2, -3

ase/scripts/generate_ase_environment.py
lzhan55: Are Qsys files needed for ASE simulation? Or it is only used for RTL sysnthesis?

michael-adler: Simulation.

ase/scripts/generate_ase_environment.py
michael-adler: Qsys is told to generate Verilog here, so the generated files get added to the Verilog file list. VHDL can still instantiate it, right? This is how the old script worked, too.

ase/scripts/generate_ase_environment.py
lzhan55: -5 is used to get the last five characters in string 's' in python.

ase/scripts/generate_ase_environment.py
michael-adler: The temporary tree is inside the directory created for simulation, just like the old version. We don't currently clean up the created simulation tree on any error. I suppose we could, but that could be done at a higher level than this change since this source of errors is only one of many and the action would be the same.

ase/scripts/generate_ase_environment.py
michael-adler: I just discovered Python's str.endswidth(). I'll switch to that.

ase/scripts/generate_ase_environment.py
deepakcu84: My understanding is --synthesis=VHDL must be passed to generate the right file sets for VHDL. At-least the generated top-level wrapper must be VHDL and .vhd files must be copied over. Unfortunately, we don't have a Qsys VHDL test. Let us revisit it later.

ase/scripts/generate_ase_environment.py
deepakcu84: Ok.

PR #389, opened 2018-05-02
Improve error reporting of sim/synth configuration scripts.

michael-adler:


michael-adler:
@abelardojarab or @lzhan55 it is safe to approve this despite the failed check. The Python check is fixed by https://github.com/OPAE/opae-sdk/pull/388. I will merge that first. I want to use that order because this pull will be cherry picked for 1.1.

Regression tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2352191

PR #390, opened 2018-05-02
Merge feature/c++core into master

r-rojo:
This merges feature/c++core into master.
All commits (excluding some commits where master was merged in) into this feature branch have all gone through the same pull request process with code owner(s) approval.


michael-adler:
This is really great and represents a lot of work!

What is the value of libopae++/samples/simple_example.cpp? I learned a lot more reading the hello example. Is there something simple_example.cpp teaches that isn't also learned easily in the more complete hello example?

There is a name clash between your dma class and the dma AFUs shipped with PACs. In your case, DMA just sets up a buffer in preparation for DMA. The AFU actually manages traffic. I hate to bring this up so late, but didn't see the class earlier. Does it make sense to rename your class to something more about buffer management?

Does the hello example work with ASE? I suspect the answer is yes -- just wanted to be sure.

r-rojo:
@michael-adler 
The simple_example started out before the API definitions as a way of how the API could be used. We can remove it if it's not useful.

For the dma_buffer class, I've wondered that myself (about the name clash). Should we rename it to something like `shared_buffer`?

PR #399, opened 2018-05-07
Update codeowners to include C++ API

r-rojo:


.github/CODEOWNERS
```diff
@@ -7,6 +7,7 @@
 # API specification, implementation, and samples
 /common/include/OPAE/          @OPAE/api
 /libOPAE/                      @OPAE/api
+/libOPAE++/                    @OPAE/api
```
tswhison: Are these not case-sensitive?  It's actually libopae and libopae++..

r-rojo: I wondered that myself. I used all caps for OPAE to copy look/feel of previous but I suppose it would be better to be explicit. Will change both.

PR #400, opened 2018-05-08
hssi: add delay in mdio

r-rojo:
This is a workaround to account for lack of synchronization in mdio controller.


r-rojo:
I certainly hope so

PR #401, opened 2018-05-08
libopae: update drv/user shared file intel-fpga.h

tswhison:


r-rojo:
Any reason why the `#define` preprocessor macros moved inline with struct definitions?

tswhison:
That's just the way the driver team decided to do things..

PR #406, opened 2018-05-09
Update version number to latest release

nakulkorde:


nakulkorde:
@r-rojo I think everything has been cherry-picked. But I will verify and if there are any, I will add them in a separate PR.

PR #413, opened 2018-05-11
Initial version of OPAE Python API

r-rojo:
# PYOPAE - Python bindings for OPAE
This review introduces the Python bindings for the OPAE SW stack. The Python classes introduced here are modeled after the opae-cxx-core classes (which are actually wrapped by the binding code).

# Namespace packages
This change sets up the Python installation convention for OPAE Python packages.
This follows `pkgutil-style namespace packages` as referenced in the [packaging](https://packaging.python.org/guides/packaging-namespace-packages) guide for Python.
This is the recommended format for compatibility with Python 2.3+ and Python 3.
The root namespace for this is `opae` with this API implementation being installed in the `opae.api` package. Python packages that are part of the OPAE SDK but not part of the API should be installed in the `opae` namespace so long as they do not conflict with the `opae.api` package.

# Python API classes and methods
The following table describes the Python classes and methods that are introduced by this pull request.

 Class Name | Type | Description                                                                                                                    
--------------|-------|---------------------------------------------------------------------------------------
version         | Method | Reports the OPAE API version as a three element tuple
properties    | Class      | Container class which can be used to enumerate or to describe an existing resource 
token           | Class      | Used to represent an accelerator resource (from enumeration)
handle         | Class      | Represents a resource that has been opened 
shared_buffer | Class | Represents a buffer that can be shared with an accelerator resource
event         | Class | Represents an event object that has been registered for notification
fpga_objtype | Enum | Used to describe an object type (fpga or accelerator)
fpga_opae_flags | Enum | Used to describe flags used when opening a resource
fpga_status | Enum | Used to describe if a handle is currently open or closed
fpga_event_type | Enum | Used to describe an event type to register for
fpga_accelerator_state | Enum | Used to describe state of accelerator resource (assigned vs unassigned)

# The following describes the files included in this request.

## opae.cpp
The file, opae.cpp, contains the code which uses [pybind11](https://github.com/pybind/pybind11) to add the bindings. This will be compiled into the shared object _opae.so inside the resulting Python package, `opae.api`.

## setup.py
Used to build the Python bindings using Python's setuptools/distutils. This is a templated file that cmake will use to create a setup.py file in the build directory. This contains the logic for building the opae Python bindings in opae.cpp.

## test_pyopae.py
Contains unit tests for the basic functionality of the bindings

## CMakeLists.txt
Contains the cmake logic to "build" the pyopae package.
The CMakeLists.txt can use Python's virtualenv module to create a virtual Python environment in the build directory. Once the virtual environment has been created, a developer may source the `bin/activate` file to activate the virtual Python environment. Once in this environment, any calls to Python or its modules will be resolved by the files in the virtual environment.

## create_python_dist.cmake
The function in create_python_dist is used to process the setup.py file and any required files.
This will copy the required files to the build directory and call on setup.py to install a development version of pyopae in the virtual environment.

# TODO
As this is  feature branch, there are still items left to do before this can be merged into the master branch. This branch will be used for continued development.

These items to add include the following:
* Add Python `docstrings` to the Python API
* Improve/expand the unit test coverage
* Continue to sync the API with any changes in opae-cxx-core
  * Remove reconfigure method and replace with Python wrapper around handle::reconfigure once it's implemented
  * Implement event.os_object once implemented in opae-cxx-core
* Move process to create Python virtual environment into cmake macro/function


pyopae/opae/__init__.py
```diff
@@ -0,0 +1,7 @@
+from _opae import properties
+from _opae import token
+from _opae import handle
+from _opae import shared_buffer
+from _opae import event
+from _opae import *
+__all__ = [ 'reconfigure', 'version', 'properties', 'token', 'handle', 'shared_buffer', 'event']
```
buie1: Do the `fpga_objype`, `fpga_status`, etc. be included here?

r-rojo: @buie1 , No, they don't since they are not Python modules. They get included with:
```Python
from _opae import *
```

PR #415, opened 2018-05-12
Quartus include of Tcl file should use SOURCE_TCL_SCRIPT_FILE

michael-adler:


michael-adler:
Arthur discovered this obvious syntax error in generating Quartus scripts. Confirmed that Tcl scripts added to Quartus projects are now run correctly.

PR #416, opened 2018-05-14
Document LimitMEMLOCK configuration for services

southerngs:


luebbers:
It's also possible to specify the limit for all systemd services using `/etc/systemd/system.conf`. See https://www.freedesktop.org/software/systemd/man/systemd-system.conf.html#DefaultLimitCPU=

PR #419, opened 2018-05-16
Support local memory widths that aren't 512 bits

michael-adler:
- ASE models whatever width the platform or AFU configure.
- AFU JSON may now request any local memory data or address width that is no
  larger than the platform's memory interface.


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2380344

@deepakcu84: The full memory model remains at fixed 512 bit wide data. At some point it needs to be updated to handle parameterization. That may be difficult. Maybe you'll have to make it wider and scale truncate the ports using a shim to whatever the AFU requests -- similar to what we would do on hardware.

PR #422, opened 2018-05-17
libopae: remove extraneous function del_dev

tswhison:
from the enumeration code.


tswhison:
Stale. Will re-create with latest master.

PR #425, opened 2018-05-21
fpgad: add mock test for AP6

tswhison:
Add AP6 mock test to increase fpgad code coverage.


tswhison:
The test is failing because it depends on the libmock change from #424

PR #427, opened 2018-05-22
c++core: rename get() fn to c_type()

r-rojo:
To avoid confusion and to better represent what this function does, it
is being renamed to `c_type()`
The plan is to also create a factory function for properties called
`get` which would be more in line with the OPAE C API
(`fpgaGetProperties`).


michael-adler:
Why not c_handle()?

michael-adler:
Never mind the previous comment if you saw it about c_handle(). I didn't read far enough to other types.

r-rojo:
c_handle for every C++ type or c_handle for the handle class, and c_token for the token class, etc...?

michael-adler:
@r-rojo please merge the change in intel-fpga-bbb (https://github.com/OPAE/intel-fpga-bbb/pull/23) when you merge this one. I put you on the review.

r-rojo:
@tswhison , I don't know if I'd call this "hand holding". It's just a more explicit API for those that prefer it.

PR #428, opened 2018-05-22
refactor enumeration code

r-rojo:
This changes the enumeration logic to only read afu_id from a port if it
has to. The conditions to when it should read afu_id are:
* No filters are given
* At least one filter includes FPGA_ACCELERATOR as the type
* At least one filter does not specify an object type property

To do this, the function `enum_fme_afu` has been split up in two:
* enum_fme
* enum_afu

The function enum_afu only gets called when the conditions listed above
are met.

Oh, and I also ran this through clang-format which would explain why it is somewhat reformatted...


r-rojo:
We don't but I wanted to throw this out here and get feedback to see if it's worth merging.
It is more of an optimization to read from AFU CSR space only if necessary. I'm OK with abandoning this if people don't see a strong need for this.
I manually tested this and it worked but adding more tests might be a good idea.

luebbers:
If it's supporting a use case / solves an issue, I don't see a problem with merging it. It does complicate the code, though, so if it's not required, I'd rather not merge it to keep the code as simple as possible.

There may be an opportunity in the future if the enumeration code gets refactored to integrate this.

In other news, this patch does include some nice simplifications (`_parent_tok`) worth keeping separate and applying no matter what.

r-rojo:
After thinking about it some more, I do feel this is worth merging as it does support instances where an accelerator may be misbehaving. This will avoid mmio timeout errors when enumerating resources as part of the PR flow.
@tswhison, what kind of tests would you like to see for this?

tswhison:
re: tests - so long as the code coverage numbers don't decrease with the change..

r-rojo:
Coverage seems to improve slightly (using the mock driver).
The top report for enum.c is for this branch. 
image

The logic that determines when to enumerate afus is all tested too:
image

PR #430, opened 2018-05-23
Add sim_afu

michael-adler:
sim_afu loads the ASE version of libopae-c, turning software compiled for use
with hardware into a simulation run.


michael-adler:
What do you all think of this? I've been struggling with the problem of having to write Makefiles for examples that either build both ASE and FPGA versions of a binary or a mode that builds one or the other. This script lets us just compile for FPGA and invoke it for ASE by forcing it to load libopae-c-ase.

Any suggestions for the script name? sim_afu is ok, but maybe there is something better.

deepakcu84:
Why can't you use a flag in the Makefile? E.g. make USE_ASE=1 builds the ASE version.

If you still want to keep the script, how about "run_as_ase", "run_in_ase" or "use_ase" instead of sim_ase?

michael-adler:
I like "run_in_ase".

The reason I don't like the USE_ASE=1 flag, especially, is because it builds a single named binary that has different semantics depending on the flag. You never know how it was compiled last, so are forced to do "make clean" every time you do a make in order to be sure it is compiled how you need it.

chipb:
I also like `run_in_ase` and this script in general.

It'd be an atypical case, but I think we should *extend* LD_PRELOAD rather than overriding entirely:
```
env LD_PRELOAD="libopae-c-ase.so${LD_PRELOAD:+ $LD_PRELOAD}" "$@"
```

abelardojarab:
Is there a consistent methodology on the motivations for naming these scripts? I think it would be useful to see what is being envisioned here.

ase/scripts/run_in_ase
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+# Run a binary using the ASE simulation library instead of the standard libopae-c.
+env LD_PRELOAD="libopae-c-ase.so${LD_PRELOAD:+:$LD_PRELOAD}" "$@"
```
chipb: Huh. I think I've only seen `LD_PRELOAD` as being space-delimited, not `:`-delimited, but I guess the dynamic linker accepts either.

chipb:
Not to dwell on something inconsequential, but another name occurred to me. How does plain `with_ase` sound?

As a sanity check to potentially add later, it might be worth verifying that `libopae-c-ase.so` appears on the search path prior to launching the command.

michael-adler:
@chipb I think you are probably right about `with_ase`. I was finding `run_in_ase` tedious to type.

Any disagreement? Not of deep consequence, but typed a lot so worth a couple extra minutes.

deepakcu84:
I'm fine with with_ase

michael-adler:
@OPAE/ci waiting for you to approve one line make change.

PR #432, opened 2018-05-24
Refactor code that examines filters

r-rojo:
Move the code that looks at filters passed in to `fpgaEnumerate`
to determine is reading afu_id is required.


r-rojo:
@luebbers , Yes, good point. We will revisit enumeration later and refactor it.
Thanks!

PR #437, opened 2018-05-24
libopae: fix issues identified by checkers

anandaravuri:
Fixed Checker issues

### DIFF
--- a/libopae/src/bitstream.c
+++ b/libopae/src/bitstream.c
@@ -168,7 +168,7 @@ fpga_result get_interface_id(fpga_handle handle, uint64_t *id_l, uint64_t *id_h)
 		return FPGA_EXCEPTION;
 	}
 
-	memset(buf, 0, sizeof(buf));
+	memset_s(buf, sizeof(buf), 0);
 
 	b = 0;
 	do {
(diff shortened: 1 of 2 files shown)


libopae/src/bitstream.c
```diff
@@ -552,7 +552,7 @@ fpga_result read_gbs_metadata(const uint8_t *bitstream,
 				e = memcpy_s(gbs_metadata->afu_image.afu_clusters.name,
 						AFU_NAME_LEN,
 						json_object_get_string(name),
-						sizeof(json_object_get_string(name)));
+						strlen(json_object_get_string(name))+1);
```
tswhison: The checkers might want strnlen_s().

libopae/src/bitstream.c
```diff
@@ -552,7 +552,7 @@ fpga_result read_gbs_metadata(const uint8_t *bitstream,
 				e = memcpy_s(gbs_metadata->afu_image.afu_clusters.name,
 						AFU_NAME_LEN,
 						json_object_get_string(name),
-						sizeof(json_object_get_string(name)));
+						strnlen_s(json_object_get_string(name), AFU_NAME_LEN));
```
r-rojo: Would it better to use `json_object_get_string_len`?
I'm not sure which is better.

PR #438, opened 2018-05-24
tools: fix issues identified by checkers

anandaravuri:


tools/extra/mmlink/mmlink_connection.cpp
```diff
@@ -44,7 +44,10 @@
 using namespace std;
 
 const char *mmlink_connection::UNKNOWN = "UNKNOWN\n";
+#define MMLINK_UNKNOWN_SIZE  9
```
r-rojo: Why is this 9?
UNKNOWN\n is eight chars.

tools/extra/mmlink/mmlink_connection.cpp
```diff
@@ -44,7 +44,10 @@
...
+
 const char *mmlink_connection::OK = "OK\n";
+#define MMLINK_OK_SIZE  4
```
r-rojo: OK\n is three chars

tools/extra/mmlink/mmlink_connection.cpp
```diff
@@ -198,7 +201,7 @@ int mmlink_connection::handle_unbound_command(char *cmd)
 				" setting to bound state\n";
 
 		bind();
-		send(OK, strnlen_s(OK, sizeof(OK)));
+		send(OK, strnlen_s(OK, MMLINK_OK_SIZE));
```
r-rojo: Seeing as this is C++ already, can't you use C++ strings instead?

anandaravuri: extra size for "\0" , added +1 to it

PR #442, opened 2018-05-25
Add with_ase (#430)

michael-adler:
with_ase loads the ASE version of libopae-c, turning software compiled for use
with hardware into a simulation run.


michael-adler:
We decided to cherry pick this into the 1.1 release because it affects how we build AFUs that are compatible with multiple releases. It would be nice to make all the current major releases consistent.

PR #444, opened 2018-05-25
ASE Klocwork fix for using safe memset() function

lzhan55:
Added the safe version memset in ase_strings.c. This should fix the klocwork issues related to memset() in ASE. 

DCP regression tests passed at:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2402072

### DIFF
--- a/ase/sw/ase_ops.c
+++ b/ase/sw/ase_ops.c
@@ -293,7 +293,7 @@ char *ase_malloc(size_t size)
 		exit(1);
 #endif
 	} else {
-		memset(buffer, 0, size);
+		ase_memset(buffer, 0, size);
 		FUNC_CALL_EXIT;
 		return buffer;
 	}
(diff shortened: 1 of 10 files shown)


lzhan55:
DCP ase regression tests passed.

ase/sw/ase_ops.c
```diff
@@ -758,6 +758,16 @@ void ase_memcpy(void *dest, const void *src, size_t n)
 	ase_memcpy_s(dest, n, src, n);
 }
 
+/*
+* ase_memset - Secure memset abstraction
+*/
+void ase_memset(void *dest, int ch, size_t n)
+{
+	// Secure implementation
+	ase_memset_s(dest, n, ch, n);
```
michael-adler: I assume you realize that, just like all these other "ase" functions, the majority of the protection offered is eliminated by passing _n_ to both the _dmax_ and _count_ fields. We are satisfying Klocwork but not accomplishing much.

lzhan55: Agree.  We might need to switch to using SDL safe string in the future.

michael-adler: That won't help. It has the same problem. The problem is we are asserting the length of the target buffer matches the length of the source, which is exactly what is supposed to be tested. Unless we have the targets being smart objects that know their length (which likely means writing in C++ instead of C), this exercise is pointless.

PR #448, opened 2018-05-30
c++core: Update hello sample to match hello_fpga

r-rojo:


libopae++/samples/hello_fpga-1.cpp
```diff
@@ -42,9 +43,9 @@ static const uint64_t CL = 64;
 static const uint64_t KB = 1024;
```
tswhison: Just a nit to consider updating names - these definitions actually implement KiB and MiB.

PR #454, opened 2018-06-05
cxx-core: Add error API for opae-cxx-core

r-rojo:
This adds the errors class in opae::fpga::types namespace which
wraps the errors API in the OPAE C API.


r-rojo:
@michael-adler , That sounds like a good idea. 👍 
I'll add it an a separate commit.

PR #461, opened 2018-06-07
CI: Add a release mode build to per checkin process

nakulkorde:


nakulkorde:
The issue with that is if there are three different values to the same environment variable, instead of starting three different builds with three different parameters to the script, Travis gets confused and doesn't start the build at all.

PR #462, opened 2018-06-07
Emit platform configuration for ASE is Makefile/CMake includes.

michael-adler:
This will allow us to import Quartus objects that are Quartus/platform
dependent.

### DIFF
--- a/platforms/scripts/platmgr/emitcfg.py
+++ b/platforms/scripts/platmgr/emitcfg.py
@@ -288,6 +288,49 @@ def emitConfig(args, afu_ifc_db, platform_db, platform_defaults_db,
 
 
 #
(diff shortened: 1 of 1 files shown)


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2432307

This code takes state from the Platform DB that is already emitted to Verilog header files and also emits it to files that can be included in make/cmake (rtl/ase_platform_config.mk and rtl/ase_platform_config.cmake). These files hold preprocessor variables defined in the "define" section of a platform DB file. make/cmake can choose to load particular Quartus atoms by looking for variables such as "PLATFORM_FPGA_FAMILY_A10".

PR #463, opened 2018-06-08
tools: Add platform specific subdirectories

r-rojo:


southerngs:
@r-rojo can comment on the purpose of this change?  Also I assume this doesn't change install location, just source location.

r-rojo:
@southerngs , part of the original plan to reorganize the tools source included adding platform specific directory names in the tools/extra.
This doesn't change the install location - only the source tree is changed.

PR #466, opened 2018-06-11
Added more test cases to test_pyopae.py.

ibarraza:


pyopae/test_pyopae.py
```diff
@@ -37,84 +37,154 @@
...
         props = opae.fpga.properties(type=opae.fpga.ACCELERATOR)
         assert props.type == opae.fpga.ACCELERATOR
+        props = opae.fpga.properties()
+        props.type = opae.fpga.ACCELERATOR
+        assert props.type == opae.fpga.ACCELERATOR
 
     def test_set_objtype_device(self):
         props = opae.fpga.properties(type=opae.fpga.DEVICE)
         assert props.type == opae.fpga.DEVICE
+        props.type = opae.fpga.DEVICE
```
r-rojo: We've already asserted that the type is of DEVICE on line 70 so we should probably set the type to something different than what it was before.

pyopae/test_pyopae.py
```diff
@@ -37,84 +37,154 @@
 MOCK_PORT_ERROR = "/tmp/class/fpga/intel-fpga-dev.0/intel-fpga-port.0/errors/errors"
 
 class TestProperties(unittest.TestCase):
+    def test_set_parent(self):
+        props = opae.fpga.properties()
```
r-rojo: We should set the first filter to type=DEVICE so that it can be a parent.

pyopae/test_pyopae.py
```diff
@@ -37,84 +37,154 @@
...
+        guid_str = props.guid
+        guid = uuid.UUID(guid_str)
+        assert str(guid).lower() == NLB0
 
     def test_set_objtype_accelerator(self):
         props = opae.fpga.properties(type=opae.fpga.ACCELERATOR)
         assert props.type == opae.fpga.ACCELERATOR
+        props = opae.fpga.properties()
+        props.type = opae.fpga.ACCELERATOR
+        assert props.type == opae.fpga.ACCELERATOR
```
r-rojo: so maybe here we create properties of type device and then set it to accelerator

pyopae/test_pyopae.py
```diff
@@ -37,84 +37,154 @@
...
 
     def test_set_objtype_device(self):
         props = opae.fpga.properties(type=opae.fpga.DEVICE)
         assert props.type == opae.fpga.DEVICE
+        props.type = opae.fpga.ACCELERATOR
+        assert props.type == opae.fpga.ACCELERATOR
```
r-rojo: and here we create a properties object of type accelerator and then set the property to device.

pyopae/test_pyopae.py
```diff
@@ -37,84 +37,154 @@
...
 
     def test_set_bus(self):
         props = opae.fpga.properties(bus=0x5e)
         assert props.bus == 0x5e
+        props = opae.fpga.properties()
+        props.bus = 0x5e
+        assert props.bus == 0x5e
```
r-rojo: set bus to something other than 0x5e the second time around

pyopae/test_pyopae.py
```diff
@@ -37,84 +37,154 @@
...
 
     def test_set_device(self):
         props = opae.fpga.properties(device=0xe)
         assert props.device == 0xe
+        props = opae.fpga.properties()
+        props.device = 0xe
+        assert props.device == 0xe
```
r-rojo: set device to something other than 0xe the second time around

PR #469, opened 2018-06-12
add the default visibility to error API

r-rojo:
Add the \_\_FPGA_API\_\_ macro to make the error API functions visible when linking
with other modules.

### DIFF
--- a/ase/api/src/error.c
+++ b/ase/api/src/error.c
@@ -31,7 +31,9 @@
 #include "common_int.h"
 #include "opae/error.h"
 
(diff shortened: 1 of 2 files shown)


chipb:
Hold on, we should also cover the ASE versions of these functions too, right?

Will try to add in another commit.

chipb:
Sigh. Tabs + Spaces > Only Spaces?

r-rojo:
@chipb , Thanks for your help with the ASE functions.

PR #470, opened 2018-06-12
tools: fpgaconf prints actual and expected  interface id

anandaravuri:
Changes:
   - Prints actual and expected  interface id with  verbosity turned up


libopae/src/bitstream_int.h
```diff
@@ -142,6 +142,19 @@ fpga_result validate_bitstream_metadata(fpga_handle handle,
...
+ * Reads interface id high and low values
+ *
+ * Reads interface id from sysfs.
+ *
+ * @param[in] handle   FME handle
+ * @param[out] id_l    Interface id low
+ * @param[out] id_h    Interface id lHigh
+ * @returns             FPGA_OK on success
+ */
+fpga_result get_interface_id(fpga_handle handle,
```
tswhison: I think this should be tagged with __FIXME_MAKE_VISIBLE__

tools/base/fpgaconf/fpgaconf.c
```diff
@@ -487,6 +488,84 @@ int read_bitstream(char *filename, struct bitstream_info *info)
...
+	fpga_properties   filter              = NULL;
+	uint32_t          num_matches         = 0;
+	int               retval              = -1;
+	uint64_t          intfc_id_l          = 0;
+	uint64_t          intfc_id_h          = 0;
+	fpga_handle       fpga_handle         = NULL;
+	fpga_result       res                 = -1;
+	fpga_token        fpga_token          = NULL;
+	fpga_guid         expt_interface_id   = {0};
+	char              guid_str[37]        = {0};
```
r-rojo: All of this extra spacing may eventually get taken out with some editor settings/tools for coding style.

libopae/src/common.c
```diff
@@ -184,3 +184,29 @@ uint64_t wsid_gen(void)
...
 }
+
+/**
+ * @brief
+ *
+ * @param guidh
+ * @param guidl
+ * @param guid
+ */
+void __FIXME_MAKE_VISIBLE__ fpga_guid_to_fpga(uint64_t guidh, uint64_t guidl, uint8_t *guid)
```
tswhison: Hey Ananda, sorry for the trouble, but I had a recent pull request that removed this function (it was unused in the API).  Do you think it makes sense to move it into fpgaconf, since that's the only entity that uses it?

PR #472, opened 2018-06-13
libopae: Reconfigure bitstream after assign afu to vm

anandaravuri:


zhangzhang10:
Does this provide the feature of PR through VF? Thanks!

PR #475, opened 2018-06-14
add getitem methods to shared_buffer class

r-rojo:
* Implement the __getitem__ protocol with shared_buffer class.
  This allows getting bytes (or byte slices) from a shared_buffer object.
* Remove `memoryview` method from shared_buffer class as it already implements
  the buffer protocol. This means the Python `memoryview` class can be used
  with a shared_buffer object.


pyopae/opae.cpp
```diff
@@ -170,16 +169,17 @@ PYBIND11_MODULE(_opae, m) {
...
-           shared_buffer_doc_to_memoryview())
       .def_buffer([](shared_buffer &b) -> py::buffer_info {
-    return py::buffer_info(const_cast<uint8_t *>(b.c_type()), b.size());
-      });
+        return py::buffer_info(
+            const_cast<uint8_t *>(b.c_type()), sizeof(uint8_t),
+            py::format_descriptor<uint8_t>::format(), b.size());
+      })
+      .def("__getitem__", shared_buffer_getitem, shared_buffer_doc_getitem())
+      .def("__getitem__", shared_buffer_getslice, shared_buffer_doc_getslice());
```
tswhison: Should both of these be "__getitem__"?

r-rojo: Yes. This is "overloaded" to support slicing.
The first method allows using the buffer like this:
```Python
# get byte at offset 256
value = buf[256]
```
The second method allows slicing:
```Python
# get the first 128 bytes
values = buf[0:128]
```

PR #477, opened 2018-06-15
libopaec: increase code coverage numbers

tswhison:
Remove unused function aal_guid_to_fpga().
Add tests for fpgaGetPropertiesFromHandle(),
fpgaPropertiesGetObjectID(), fpgaPropertiesSetObjectID().
Comment out (currently) unused function token_cleanup.
Add test event_drv_07 to stress list walk code in fpgad.


libopae/src/token_list.c
```diff
@@ -186,6 +186,7 @@ struct _fpga_token *token_get_parent(struct _fpga_token *_t)
  * Clean up remaining entries in linked list
  * Will delete all remaining entries
  */
+/*
```
r-rojo: I don't know how I feel about disabling code like this just to improve coverage.
Should we add a destructor attribute so that this gets called when the library is unloaded? I think we should be able to test it then (and get coverage).
```C
__attribute__((destructor))
```

PR #478, opened 2018-06-15
pyopae: enumerate w/ kwargs

r-rojo:
Add an overload method of enumerate that takes in kwargs for the filter.
This eliminates the need for creating properties objects before calling
enumerate and allows client code to call enumerate with kwargs only.

This change also creates pytoken.h|cpp files which contain the code for the
documentation and implementation of token methods.


pyopae/opae.cpp
```diff
@@ -136,7 +115,8 @@ PYBIND11_MODULE(_opae, m) {
                     properties_doc_accelerator_state());
 
   // define token class
-  m.def("enumerate", &token::enumerate, token_doc_enumerate());
+  m.def("enumerate", &token::enumerate, token_doc_enumerate())
+    .def("enumerate", token_enumerate_kwargs, token_doc_enumerate_kwargs());;
```
tswhison: nit: additional semicolon at the end. :)

PR #484, opened 2018-06-18
libopae-cxx-core: add test for src_location

tswhison:
Add test for src_location's copy-assignment operator.

### DIFF
--- a/tests/function/gtCxxExcept.cpp
+++ b/tests/function/gtCxxExcept.cpp
@@ -53,6 +53,22 @@ TEST(LibopaecppExceptCommonALL, except_03) {
 }
 
 /**
+ * @test except_04
+ * Given a src_location<br>
+ * When the object is copy-assigned<br>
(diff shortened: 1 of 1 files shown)


tests/function/gtCxxExcept.cpp
```diff
@@ -52,6 +52,22 @@ TEST(LibopaecppExceptCommonALL, except_03) {
...
 
+/**
+ * @test except_04
+ * Given a src_location<br>
+ * When the object is copy-assigned<br>
+ * Then the new object is a copy of the original.<br>
+ */
+TEST(LibopaecppExceptCommonALL, except_04) {
+  src_location l1(__FILE__, __func__, __LINE__);
+  src_location l2("someotherfile.txt", "some_other_fn", __LINE__);
```
r-rojo: Should the third argument be something other than __LINE__ here?

r-rojo: 🤓

PR #488, opened 2018-06-19
cxx-core: add null tests

r-rojo:
Add checks for null arguments and throw invalid_argument if check fails.

While doing negative testing from Python API, these missing checks resulted in segfaults.
Putting these checks in will enable development of negative tests from Python.

TODO:
Look into possibly using except class for throwing those exceptions.


r-rojo:
@tswhison 
This resulted when I was testing `None` (or null) arguments from the Python API.

PR #489, opened 2018-06-19
Docs: Update information on landing page

nakulkorde:
Add information to the documentation landing page.
This change adds a brief description of OPAE that we add with every release.
Please suggest if any other information should be added.


nakulkorde:
image

The page looks like this and the list of  versions go towards the end of the page.

PR #490, opened 2018-06-19
tools: fpgaconf prints actual and expected interface id #

anandaravuri:


tools/base/fpgaconf/fpgaconf.c
```diff
@@ -387,6 +387,85 @@ int parse_metadata(struct bitstream_info *info)
 	return 0;
 }
 
+/*
+* Prints Actual and Expected Interface id
+*/
+int prints_interface_id(fpga_guid actual_interface_id)
```
r-rojo: Just a small nit, can you rename this function to `print_interface_id`?

libopae/src/bitstream_int.h
```diff
@@ -142,6 +142,19 @@ fpga_result validate_bitstream_metadata(fpga_handle handle,
...
+* Reads interface id high and low values
+*
+* Reads interface id from sysfs.
+*
+* @param[in] handle   FME handle
+* @param[out] id_l    Interface id low
+* @param[out] id_h    Interface id lHigh
+* @returns             FPGA_OK on success
+*/
+fpga_result get_interface_id(fpga_handle handle,
```
tswhison: Querying the FME's guid property actually returns the PR interface ID, so this method of obtaining it isn't strictly necessary.  That's an optimization that we can make later, though.

r-rojo: I agree with @tswhison , we can (and should do this later). You should either add a TODO here or track it elsewhere.

PR #492, opened 2018-06-20
tests: add bitstream metadata test case to increase code coverage number

anandaravuri:


tests/unit/gtBSMetadata.cpp
```diff
@@ -0,0 +1,169 @@
...
+	result = read_gbs_metadata(bitstream_guid, &gbs_metadata);
+	ASSERT_NE(result, FPGA_OK);
+
+	// Invalid input bitstream
+	uint8_t bitstream_guid_invalid1[] = "XeonFPGA�GBSv001";
+	result = read_gbs_metadata(bitstream_guid_invalid1, &gbs_metadata);
+	ASSERT_NE(result, FPGA_OK);
+
+	// Invalid input bitstream
+	uint8_t bitstream_guid_invalid2[] = "XeonFPGA�GBSv001S  {\"version/\": 640, \"afu-image\":  \
```
tswhison: Looks like there are some Unicode characters in these metadata strings.  Not sure if that's what you intended?  Looks good, otherwise.

tests/unit/gtBSMetadata.cpp
```diff
@@ -0,0 +1,169 @@
...
+	result = read_gbs_metadata(bitstream_guid_invalid1, &gbs_metadata);
+	ASSERT_NE(result, FPGA_OK);
+
+	// Invalid input bitstream
+	uint8_t bitstream_guid_invalid2[] = "XeonFPGA·GBSv001S  {\"version/\": 640, \"afu-image\":  \
+		{\"clock-frequency-high\": 312, \"clock-frequency-low\": 156,\
+		\"power\": 50, \"interface-uuid\": \"1a422218-6dba-448e-b302-425cbcde1406\", \
+		\"magic-no\": 488605312, \"accelerator-clusters\"\
+		: [{\"total-contexts\": 1, \"name\": \"nlb_400\", \"accelerator-type-uuid\": \
+		\"d8424dc4-a4a3-c413-f89e-433683f9040b\"}]}, \"platform-name\": \"MCP\"}";
```
r-rojo: Seems like you have non-alphanumeric byte(s)  at the end of the string. You might also want to consider using C++ raw strings (assuming gtapi is built with c++11).
https://en.cppreference.com/w/cpp/language/string_literal

tests/unit/gtBSMetadata.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+	// Invalid input
+	struct gbs_metadata gbs_metadata ;
+	result = read_gbs_metadata(NULL, &gbs_metadata);
+	ASSERT_NE(result, FPGA_OK);
+
+	// Invalid input bitstream
+	uint8_t bitstream[10] = "abcd";
+	result = read_gbs_metadata(bitstream, NULL);
+	ASSERT_NE(result, FPGA_OK);
```
r-rojo: You might want to change some of these ASSERT_NE to EXPECT_NE so that the test doesn't terminate at the first failure it encounters. Only use ASSERT_.* when it doesn't make sense to continue the test.

tests/unit/gtBSMetadata.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+	// Invalid input bitstream
+	uint8_t bitstream_guid_invalid8[] = "XeonFPGA·GBSv001\00\00\00\00{\"version\": 640, \"afu-image\":\
+		{\"clock-frequency-high\": 312, \"clock-frequency-low\": 156, \
+		\"power\": 50, \"interface-uuid\": \"1a422218-6dba-448e-b302-425cbcde1406\", \
+		\"magic-no\": 488605312, \"accelerator-clusters\": [{\"total-contexts\":1,\
+		\"name\": \"nlb_400\", \"accelerator-type-uuid\":\
+		\"d8424dc4-a4a3-c413-f89e-433683f9040b\"}]}, \"platform-name\": \"MCP\"}";
+	result = read_gbs_metadata(bitstream_guid_invalid8, &gbs_metadata);
```
r-rojo: Are all of these invalid bitstream combinations required? Do they result in exercising different code paths?

tests/unit/gtBSMetadata.cpp
```diff
@@ -0,0 +1,169 @@
...
+	ASSERT_NE(result, FPGA_OK);
+}
+
+/**
+* @test    bs_metadata_02
+* @brief   Tests: validate_bitstream_metadata
+* @details validate_bitstream_metadata verifies metadata
+*          Then the return value is FPGA_OK<br>
+*/
+TEST(LibopaecBSMetadataCommonMOCKHW, bs_metadata_02) {
```
r-rojo: The comments for this test don't seem to match up with what the test is doing.
Also, how different is this test from the others above?

tests/unit/gtBSMetadata.cpp
```diff
@@ -48,40 +48,40 @@ using namespace common_test;
...
 	EXPECT_NE(result, FPGA_OK);
 
-	// Invalid input bitstream
+	// Invalid bitstream metadata  size
 	uint8_t bitstream_guid[] = "XeonFPGA·GBSv001S";
 	result = read_gbs_metadata(bitstream_guid, &gbs_metadata);
 	EXPECT_NE(result, FPGA_OK);
 
-	// Invalid input bitstream
+	// Zero metadata lenght
```
r-rojo: typo in `lenght`

tests/unit/gtBSMetadata.cpp
```diff
@@ -0,0 +1,167 @@
...
+	// Invalid input bitstream 
+	result = read_gbs_metadata(bitstream, &gbs_metadata);
+	EXPECT_NE(result, FPGA_OK);
+
+	// Invalid bitstream metadata  size
+	uint8_t bitstream_guid[] = "XeonFPGA·GBSv001S";
+	result = read_gbs_metadata(bitstream_guid, &gbs_metadata);
+	EXPECT_NE(result, FPGA_OK);
+
+	// Zero metadata lenght
```
r-rojo: misspelling

tests/unit/gtBSMetadata.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+	// Zero metadata length 
+	uint8_t bitstream_guid_invalid2[] = "XeonFPGA·GBSv001S {\"version/\": 640, \"afu-image\":  \
+		{\"clock-frequency-high\": 312, \"clock-frequency-low\": 156,\
+		\"power\": 50, \"interface-uuid\": \"1a422218-6dba-448e-b302-425cbcde1406\", \
+		\"magic-no\": 488605312, \"accelerator-clusters\"\
+		: [{\"total-contexts\": 1, \"name\": \"nlb_400\", \"accelerator-type-uuid\": \
+		\"d8424dc4-a4a3-c413-f89e-433683f9040b\"}]}, \"platform-name\": \"MCP\"}";
+
+	result = read_gbs_metadata(bitstream_guid_invalid2, &gbs_metadata);
```
r-rojo: Comment says zero metadata length but you have metadata?

tests/unit/gtBSMetadata.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+	// valid input bitstream
+	uint8_t bitstream_guid_invalid4[] = "XeonFPGA·GBSv001\53\02\00\00{\"version\": 640, \"afu-image\":\
+		{\"clock-frequency-high\": 312, \"clock-frequency-low\": 156, \
+		\"power\": 50, \"interface-uuid\": \"1a422218-6dba-448e-b302-425cbcde1406\", \
+		\"magic-no\": 488605312, \"accelerator-clusters\": [{\"total-contexts\":1,\
+		\"name\": \"nlb_400\", \"accelerator-type-uuid\":\
+		\"d8424dc4-a4a3-c413-f89e-433683f9040b\"}]}, \"platform-name\": \"MCP\"}";
+	result = read_gbs_metadata(bitstream_guid_invalid4, &gbs_metadata);
+	EXPECT_NE(result, FPGA_OK);
```
r-rojo: Comment said valid input bitstream but test expects result != FPGA_OK

tests/unit/gtBSMetadata.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/**
+* @test    bs_metadata_02
+* @brief   Tests: validate_bitstream_metadata
+* @details validate_bitstream_metadata validates BS metadata
+*          Retuns FPGA_OK if metadata is valid
+*/
+TEST(LibopaecBSMetadataCommonMOCKHW, bs_metadata_02) {
```
r-rojo: Why do we need one test for this? Why can't we move this expect statement up to the other test? Or, conversely,  why can't we separate the expect statements above into individual tests? For simplicity, I would opt for the former. @tswhison, what do you think?

tswhison: The test case is getting sortof long, but I would be happy to have the functionality in place for now. :)

Issue #411, opened 2018-05-11
Change OPAE tools CLI


r-rojo:
# OPAE Tools Command Line Arguments #

## Introduction ##
Currently, the CLI (command line interface) of the tools included in the OPAE SDK
are not all consistent. In other words, while some tools use `--socket-id` for
identifying the `SocketID` property, others may use `--socket`.

Another example is the use of short options for specifying properties like
bus, device, or function - some tools use `-b, -d, -f` while other use `-B, -D, -F`.

### Common Concepts ###
The following table lists the concepts that have been identified as being
common amongst most tools.

| Concept   | Description                                                            |
|-----------|------------------------------------------------------------------------|
| Bus       | The PCI bus number of an OPAE resource                                 |
| Device    | The PCI device number of an OPAE resource                              |
| Function  | The PCI function number of an OPAE resource                            |
| Socket ID | The socket ID encoded in the BBS                                       |
| Object ID | A 64-bit system-wide unique integer that identifies an OPAE resource   |
| GUID      | A globally unique identifier                                           |
| Timeout   | A maximum amount of time that a specific routine may be allowed to run |
| Help      | Request for help or usage message                                      |

### Proposed Changes ###
After careful review of the different tools and their CLI, the OPAE SW team is planning
on changing the tools identified as being inconsistent in order to have a common set of
command line options for concepts that are common. The following table lists those concepts
along with the long and short command line arguments that will be used for identifying them.

| Concept   | Short Form | Long Form            |
|-----------|------------|----------------------|
| Bus       | `-B`       | `--bus`              |
| Device    | `-D`         | `--device`           |
| Function  | `-F`         | `--function`         |
| Socket ID | `-S`       | `--socket-id`        |
| Object ID | `-O`         | `--object-id`        |
| GUID      | `-G`       | `--guid`             |
| Timeout   |  NA        | `--timeout-<unit> `  |
| Help      |  `-h`      | `--help`             |

_NOTE: Options related to timeouts will also not have short options and will only use long options. They will use the units after in place of `<unit>` in the table listed above
where the untis can be `seconds`, `msec` (for milliseconds), `usec` (for microseconds)._

### OPAE Tools ###
The following sections describe the tools that are part of OPAE SDK and any changes that will be made to bring a unified CLI across all of them.

#### fpgabist ####
`fpgabist` will have its short options changed  to uppercase. Those changes are:
* `-b` will be changed to `-B`
* `-d` will be changed to `-D`
[...]

PR #499, opened 2018-06-21
fpgabist: change CLI for consistency

buie1:
Changing CLI to be consistent with OPAE tools.
Responding to Issue: https://github.com/OPAE/opae-sdk/issues/411


buie1:
@asgibbons I've being seeing most of the tools going through these changes. I believe `fpgaconf`, and `fpgainfo` have already had these changes.  I can't say I remember the official POR, but the changes are already being done. Do we have a formal process to approve these sorts of changes? 
@r-rojo

r-rojo:
@buie1, not that I'm aware of. I took some of the feedback in issue #411 and update the issue to reflect the feedback. I never heard anything else so we proceeded.

PR #502, opened 2018-06-21
tests: mock driver add ap event and interface-id

anandaravuri:


r-rojo:
You should really put a description in pull requests that change binaries. We really can't see the changes made and why

PR #504, opened 2018-06-22
tests: add sysfs test case to increase code coverage number

anandaravuri:


tests/unit/gtsysfs.cpp
```diff
@@ -81,4 +80,92 @@ TEST(LibopaecsysfsCommonMOCKHW, fpga_sysfs_01) {
...
+
+	result = sysfs_write_u64("/tmp/class/fpga/intel-fpga-dev.0/intel-fpga-fme.0/socket_id", 0x100);
+	EXPECT_EQ(result, FPGA_OK);
+
+	//Invalid input parameters
+	fpga_guid guid;
+	result = sysfs_read_guid(NULL, NULL);
+	EXPECT_NE(result, FPGA_OK);
+
+	result = sysfs_read_guid("/sys/class/fpga/intel-fpga-dev.0/intel-fpga.0/", guid);
```
tswhison: I think you need intel-fpga-port.0 or intel-fpga-fme.0 instead of intel-fpga.0.

tests/unit/gtsysfs.cpp
```diff
@@ -81,4 +80,92 @@ TEST(LibopaecsysfsCommonMOCKHW, fpga_sysfs_01) {
...
+	EXPECT_NE(result, FPGA_OK);
+
+	//Invalid input parameters
+	result = get_port_sysfs(NULL, NULL);
+	EXPECT_NE(result, FPGA_OK);
+
+	//Invalid handle
+	result = get_port_sysfs((void*) 0x123, NULL);
```
tswhison: Passing (void *) 0x123 seems dangerous to me.  It would be better to declare a local variable that is initialized to an invalid value and pass its address.

tests/unit/gtsysfs.cpp
```diff
@@ -81,4 +80,92 @@ TEST(LibopaecsysfsCommonMOCKHW, fpga_sysfs_01) {
...
+	EXPECT_NE(result, FPGA_OK);
+
+	//Invalid path parameters
+	result = get_fpga_deviceid(NULL, NULL);
+	EXPECT_NE(result, FPGA_OK);
+
+	//Invalid sturct handle 
+	result = get_fpga_deviceid((void*) 0x123, NULL);
```
tswhison: Passing (void *) 0x123 seems dangerous to me.  It would be better to declare a local variable that is initialized to an invalid value and pass its address.

tests/unit/gtsysfs.cpp
anandaravuri: Invalid Handle ,  Handle is void*  , wrong type cast pointer

tests/unit/gtsysfs.cpp
anandaravuri: Invalid path , so open file fails

tests/unit/gtsysfs.cpp
tswhison: I feel strongly that this is a dangerous practice.  Consider what happens when that pointer value is accessed within get_port_sysfs().

```
_token = (struct _fpga_token *) _handle->token;
```
In this case, _handle has the value 0x123, so this will most likely seg fault.

libopae/src/sysfs.c
```diff
@@ -55,6 +55,11 @@ fpga_result sysfs_read_int(const char *path, int *i)
 	char buf[SYSFS_PATH_MAX];
 	int b;
 
+	if (path == NULL) {
```
tswhison: I'm on the fence about these new path checks for NULL.  On the one hand, it is a good thing to do this type of checking.  On the other, these are internal functions, and we have a lot of empirical evidence from our testing over time that we are not passing NULL to these. @r-rojo, do you have a preference here?

anandaravuri: open segfaults if user pass NULL, added null check before passing open

r-rojo: Yes, I saw `open` segfault with NULL input parameter. I think @anandaravuri uncovered with the negative tests.

tests/unit/gtsysfs.cpp
```diff
@@ -81,4 +80,158 @@ TEST(LibopaecsysfsCommonMOCKHW, fpga_sysfs_01) {
...
+
+	//Invalid input path
+	result = sysfs_read_int("/tmp/class/fpga/intel-fpga-dev.0/intel-fpga-fme.10", NULL);
+	EXPECT_NE(result, FPGA_OK);
+
+	result = sysfs_read_int("/tmp/class/fpga/intel-fpga-dev.0/intel-fpga-fme.0", NULL);
+	EXPECT_NE(result, FPGA_OK);
+
+	// Valid input path
+	result = sysfs_read_int("/tmp/class/fpga/intel-fpga-dev.0/intel-fpga-fme.0/socket_id", &i);
```
r-rojo: How is this a valid path? It's only valid when using with the mock driver. If this is run without the mock driver, on real hw, then this will fail.
Out mock implementation rewrites /sys to /tmp to technically you can put any three letter starting directory, however, that will ONLY work with mock. It would be better to assume real HW and let mock do the rewrite.

tests/unit/gtsysfs.cpp
```diff
@@ -81,4 +80,158 @@ TEST(LibopaecsysfsCommonMOCKHW, fpga_sysfs_01) {
...
+	//Invalid input parameters
+	result = sysfs_write_u64(NULL, 0);
+	EXPECT_NE(result, FPGA_OK);
+
+	result = sysfs_write_u64("/tmp/class/fpga/intel-fpga-dev.0/intel-fpga-fme.0", 0x100);
+	EXPECT_NE(result, FPGA_OK);
+
+	//valid path 
+	result = sysfs_write_u64("/tmp/class/fpga/intel-fpga-dev.0/intel-fpga-fme.0/socket_id", 0);
+	EXPECT_EQ(result, FPGA_OK);
```
r-rojo: Should we read it here too and compare with what we intended to write?

tests/unit/gtsysfs.cpp
```diff
@@ -81,4 +80,158 @@ TEST(LibopaecsysfsCommonMOCKHW, fpga_sysfs_01) {
...
+	result = get_port_sysfs(NULL, (char*) "/tmp/class/fpga/intel-fpga-dev.0/intel-fpga-fme.0/socket_id");
+	EXPECT_NE(result, FPGA_OK);
+
+	//Invalid handle
+	result = get_port_sysfs(h, (char*)"/tmp/class/fpga/intel-fpga-dev.0/intel-fpga-port.0/");
+	EXPECT_NE(result, FPGA_OK);
+
+
+	//Invalid path parameters
+	result = get_fpga_deviceid(NULL, NULL);
```
r-rojo: What is this testing? The first or second parameter?

PR #508, opened 2018-06-25
libopae: fpgaPrepareBuffer input paramater validate

anandaravuri:
Changes:
  - Check NULL bufffer
  - Check zero buffer length
  - Check Invalid handle

### DIFF
--- a/libopae/src/common.c
+++ b/libopae/src/common.c
@@ -71,9 +71,13 @@ fpga_result prop_check_and_lock(struct _fpga_properties *prop)
 {
 	ASSERT_NOT_NULL(prop);
 
-	if (pthread_mutex_lock(&prop->lock)) {
-		FPGA_MSG("Failed to lock mutex");
-		return FPGA_EXCEPTION;
+	if (prop->magic != FPGA_PROPERTY_MAGIC) {
+		FPGA_MSG("Invalid properties object");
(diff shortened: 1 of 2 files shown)


libopae/src/common.c
```diff
@@ -71,9 +71,13 @@ fpga_result prop_check_and_lock(struct _fpga_properties *prop)
 {
 	ASSERT_NOT_NULL(prop);
 
-	if (pthread_mutex_lock(&prop->lock)) {
-		FPGA_MSG("Failed to lock mutex");
-		return FPGA_EXCEPTION;
+	if (prop->magic != FPGA_PROPERTY_MAGIC) {
```
tswhison: We can't do this - it's not thread safe.

libopae/src/common.c
```diff
@@ -96,9 +100,13 @@ fpga_result handle_check_and_lock(struct _fpga_handle *handle)
 {
 	ASSERT_NOT_NULL(handle);
 
-	if (pthread_mutex_lock(&handle->lock)) {
-		FPGA_MSG("Failed to lock mutex");
-		return FPGA_EXCEPTION;
+	if (handle->magic != FPGA_HANDLE_MAGIC) {
```
tswhison: We can't do this - it's not thread safe.

PR #509, opened 2018-06-25
libopae: fpgaPrepareBuffer input paramater validate

anandaravuri:
Changes:
  Check NULL bufffer
  Check zero buffer length

### DIFF
--- a/libopae/src/buffer.c
+++ b/libopae/src/buffer.c
@@ -212,13 +212,15 @@ fpga_result __FPGA_API__ fpgaPrepareBuffer(fpga_handle handle, uint64_t len,
 			goto out_unlock;
 		}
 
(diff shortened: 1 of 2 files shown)


libopae/src/buffer.c
```diff
@@ -205,10 +205,24 @@ fpga_result __FPGA_API__ fpgaPrepareBuffer(fpga_handle handle, uint64_t len,
...
+		}
+
+		if (!len) {
+			FPGA_MSG("buffer length is zero");
+			result = FPGA_INVALID_PARAM;
+			goto out_unlock;
+		}
+
 		/* round up to nearest page boundary */
 		if (!len || (len & (pg_size - 1))) {
```
tswhison: With the above check on !len, we can remove !len from this if statement

PR #513, opened 2018-06-25
libopae-c: increase coverage for matches_filter

tswhison:


libopae/src/enum.c
```diff
@@ -81,7 +81,7 @@ static bool matches_filter(const struct dev_list *attr,
 	int err = 0;
 
 	if (pthread_mutex_lock(&_filter->lock)) {
-		FPGA_MSG("Failed to lock global mutex");
+		FPGA_MSG("Failed to lock filter mutex");
```
r-rojo: 👍 Good catch

PR #515, opened 2018-06-25
tools:fpgad fix double free memory address

anandaravuri:
Changes:
  -Bitsteam buffer freed twice in case of corrupted bitstream

### DIFF
--- a/tools/base/fpgad/ap6.c
+++ b/tools/base/fpgad/ap6.c
@@ -255,7 +255,9 @@ int read_bitstream(const char *filename, struct bitstream_info *info)
 	return 0;
 
 out_free:
-	free((void *)info->data);
+	if (info->data)
+		free((void *)info->data);
+	info->data = NULL;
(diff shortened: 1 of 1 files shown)


tools/base/fpgad/ap6.c
```diff
@@ -285,7 +288,8 @@ void *ap6_thread(void *thread_context)
 		ret = read_bitstream(c->config->null_gbs[i], &null_gbs_info);
 		if (ret < 0) {
 			dlog("ap6[%i]: \tfailed to read bitstream\n", c->socket);
-			free((void *)null_gbs_info.data);
+			if (null_gbs_info.data)
+				free((void *)null_gbs_info.data);
```
r-rojo: do we need to set null_gbs_info.data to NULL here?

anandaravuri: safe to set to NULL

PR #516, opened 2018-06-26
pyopae: implement setitem in shared_buffer

r-rojo:
This implements the shared_buffer __setitem__ protocol method.
NOTE: This currently only works on the argument type being of type `int`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,8 +65,8 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${OPAE_SDK_SOURCE}/cmake/modules")
 ############################################################################
 set(OPAE_PYTHON_VERSION 2.7 CACHE STRING "Python version to use for building/distributing pyopae")
 set_property(CACHE OPAE_PYTHON_VERSION PROPERTY STRINGS 2.7 3.6 3.5 3.4 3.3)
-find_package(PythonLibs ${OPAE_PYTHON_VERSION})
 find_package(PythonInterp ${OPAE_PYTHON_VERSION})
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -65,8 +65,8 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${OPAE_SDK_SOURCE}/cmake/modules")
 ############################################################################
 set(OPAE_PYTHON_VERSION 2.7 CACHE STRING "Python version to use for building/distributing pyopae")
 set_property(CACHE OPAE_PYTHON_VERSION PROPERTY STRINGS 2.7 3.6 3.5 3.4 3.3)
-find_package(PythonLibs ${OPAE_PYTHON_VERSION})
 find_package(PythonInterp ${OPAE_PYTHON_VERSION})
+find_package(PythonLibs ${OPAE_PYTHON_VERSION})
```
lzhan55: Is there a difference for switching this two lines?

lzhan55:
Are you planning to expand to other data types in addition to integer? Will your Python implementation work for Python 3.6+?

CMakeLists.txt
r-rojo: Yes, this way you will find the development libraries (and headers) that come packaged with the Python interpreter found.
In my case, I'm using a custom installation of Python 2.7.15 in a non-standard location.
Because of this, I can't find the cmake package until _after_ I've found the interpreter.

r-rojo:
@lzhan55, Yes, I would like to expand to other types in the future.
Regarding Python 3.6, while I haven't tested it myself personally, theoretically it should work as pybind11 claims to be compatible with both. I will eventually test with Python 3.6, however.

PR #518, opened 2018-06-26
pyopae: Add context manager helper class

r-rojo:
Add a singleton class, `buffer_registry` which is designed to be called by the
context manager enter function to register a handle. It is then called again by
the shared_buffer allocate function to associate the resulting buffer with the
handle. The context manager exit function should then call the deregister
function in the buffer_registry to release buffers associated with a handle.

### DIFF
--- a/pyopae/opae.cpp
+++ b/pyopae/opae.cpp
@@ -164,7 +164,7 @@ PYBIND11_MODULE(_opae, m) {
            py::arg("offset"), py::arg("value"), py::arg("csr_space") = 0);
 
   // define shared_buffer class
-  m.def("allocate_shared_buffer", &shared_buffer::allocate,
+  m.def("allocate_shared_buffer", shared_buffer_allocate,
         shared_buffer_doc_allocate());
   py::class_<shared_buffer, shared_buffer::ptr_t> pybuffer(
       m, "shared_buffer", py::buffer_protocol(), shared_buffer_doc());
--- a/pyopae/pycontext.cpp
+++ b/pyopae/pycontext.cpp
@@ -23,32 +23,36 @@
 // CONTRACT,  STRICT LIABILITY,  OR TORT  (INCLUDING NEGLIGENCE  OR OTHERWISE)
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
-#pragma once
-#include <Python.h>
 
-#include <opae/cxx/core/shared_buffer.h>
-#include <pybind11/pybind11.h>
-#include "pyhandle.h"
+#include "pycontext.h"
 
-const char *shared_buffer_doc();
+buffer_registry* buffer_registry::instance_ = nullptr;
 
-const char *shared_buffer_doc_allocate();
-opae::fpga::types::shared_buffer::ptr_t shared_buffer_allocate(
-    opae::fpga::types::handle::ptr_t hndl, size_t size);
-const char *shared_buffer_doc_size();
+buffer_registry::buffer_registry() : buffers_() {}
 
-const char *shared_buffer_doc_wsid();
+buffer_registry& buffer_registry::instance() {
+  if (buffer_registry::instance_ == nullptr) {
+    buffer_registry::instance_ = new buffer_registry();
+  }
+  return *buffer_registry::instance_;
+}
 
-const char *shared_buffer_doc_iova();
+void buffer_registry::register_handle(handle_t handle) {
+  buffers_[handle] = buffer_list_t();
+}
 
-const char *shared_buffer_doc_fill();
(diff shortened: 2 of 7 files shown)


pyopae/opae.cpp
```diff
@@ -164,7 +164,7 @@ PYBIND11_MODULE(_opae, m) {
            py::arg("offset"), py::arg("value"), py::arg("csr_space") = 0);
 
   // define shared_buffer class
-  m.def("allocate_shared_buffer", &shared_buffer::allocate,
+  m.def("allocate_shared_buffer", shared_buffer_allocate,
```
lzhan55: What is the difference between the original and the modified here?

r-rojo: The original refers to a static method defined in the `opae::fpga::types::shared_buffer` class.
The new one refers to a function defined in pyshared_buffer.h|cpp which does a little more that simply calling `shared_buffer::allocate`

pyopae/pyhandle.cpp
```diff
@@ -98,7 +99,10 @@ const char *handle_doc_context_enter() {
   )opaedoc";
 }
 
-handle::ptr_t handle_context_enter(handle::ptr_t hnd) { return hnd; }
+handle::ptr_t handle_context_enter(handle::ptr_t hnd) {
+  buffer_registry::instance().register_handle(hnd);
```
lzhan55: Where is the handle from?

pyopae/pycontext.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+buffer_registry& buffer_registry::instance() {
+  if (buffer_registry::instance_ == nullptr) {
+    buffer_registry::instance_ = new buffer_registry();
+  }
+  return *buffer_registry::instance_;
+}
+
+void buffer_registry::register_handle(handle_t handle) {
+  buffers_[handle] = buffer_list_t();
```
lzhan55: buffers_[handle] is the start of the buffer list? buffer_list_t() is to create a list?

pyopae/pyhandle.cpp
r-rojo: Yes, handle here is the C++ handle class in the C++ API which is wrapped by the Python bindings (as defined in opae.cpp) using pybind11.
The opae.cpp file also defines \_\_enter\_\_ and \_\_exit\_\_ methods for the handle class which get called when entering a context in Python.
```Python
with opae.fpga.open(tok, flags) as h: # this calls the __enter__ function
    # do stuff with h while it is open
    # you can allocate buffers with the handle here
    buf = opae.fpga.allocate_shared_buffer(h, size)
# __exit__ is called when the context is closed (when this scope exits)
assert not h # This should assert should pass as the 
assert buf.size() == 0 # the buffer should be released by the context manager __exit__

File: pyopae/pyshared_buffer.h [hunk 4]
```diff
@@ -33,7 +33,8 @@
 const char *shared_buffer_doc();
 
 const char *shared_buffer_doc_allocate();
-
+opae::fpga::types::shared_buffer::ptr_t shared_buffer_allocate(
```
lzhan55: The namespace for pointer type is so long. Is there a way to simplify it?

pyopae/pycontext.cpp
r-rojo: buffers_ is a map (associative array) keyed by a handle::ptr_t object.
This line is inserting an empty vector of buffer::ptr_t objects which is used later when registering a buffer::ptr_t object

pyopae/pyshared_buffer.h
r-rojo: You can shorten it (give it an alias) however it is not recommended to do in header files.

pyopae/test_pyopae.py
```diff
@@ -309,6 +309,18 @@ def test_allocate(self):
         ba = bytearray(buff1)
         assert ba[0] == 0xaa
 
+    def test_conext_release(self):
+        assert self.handle
+        self.handle.close()
+        with opae.fpga.open(self.toks[0]) as h:
+            buff = opae.fpga.allocate_shared_buffer(h, 4096)
```
lzhan55: what is the difference between the pyshared_buffer.cpp file "shared_buffer_allocate" and the "opae.fpga.allocate_shared_buffer" here

r-rojo: opae.cpp associates the module method `allocate_shared_buffer` with the function, `shared_buffer_allocate`, defined in pyshared_buffer.h|cpp

pyopae/pycontext.cpp
lzhan55: Is the map also a c++ data type? Or it is defined by pybind?

r-rojo: map is part of the c++ stl (standard library).
You get it by including `<map>` and is found in the `std` namespace.
```C++
#include <map>

int main(){
  std::map<char*, int> str_to_ints;
  str_to_ints["one"] = 1;
}
```

PR #521, opened 2018-06-26
libopae-c: refactor out get_handle_objtype

tswhison:
Refactor out get_handle_objtype, simplifying driver_register_event and
driver_unregister_event.

### DIFF
--- a/libopae/src/event.c
+++ b/libopae/src/event.c
@@ -223,65 +223,31 @@ static fpga_result send_uafu_event_request(fpga_handle handle,
 	return FPGA_OK;
 }
 
-static fpga_result get_handle_objtype(fpga_handle handle, fpga_objtype *objtype)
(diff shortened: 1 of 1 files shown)


libopae/src/event.c
```diff
@@ -223,65 +223,31 @@ static fpga_result send_uafu_event_request(fpga_handle handle,
 	return FPGA_OK;
 }
 
-static fpga_result get_handle_objtype(fpga_handle handle, fpga_objtype *objtype)
+static fpga_result check_interrupts_supported(fpga_handle handle, fpga_objtype *objtype)
```
r-rojo: You might want to add a comment explaining what this fn does and that the objtype is an output parameter.
Also, should we do a null check here or just leave it up to the call below on line 244?

PR #528, opened 2018-06-27
Zzhan68/event tests

zhangzhang10:


tests/function/gtEvent.cpp
```diff
@@ -63,8 +63,8 @@ class LibopaecEventFCommonMOCKHW : public BaseFixture, public ::testing::Test {
       EXPECT_EQ(FPGA_OK, fpgaDestroyEventHandle(&m_EventHandles[i]));
     }
 
-    EXPECT_EQ(FPGA_OK, fpgaClose(m_FMEHandle));
-    EXPECT_EQ(FPGA_OK, fpgaClose(m_AFUHandle));
+    fpgaClose(m_FMEHandle);
```
tswhison: Why was the EXPECT_EQ removed?

zhangzhang10: No good reason. It's a mistake. I've added EXPECT_EQ back. Thanks.

PR #529, opened 2018-06-27
mock: add IRQ support

tswhison:
Fix bug in test reconf_gbs_01.
Add conditional interrupt ioctl support to mock, controlled
by mock_enable_irq().
Fix bug in send_uafu_event_request().

### DIFF
--- a/libopae/src/event.c
+++ b/libopae/src/event.c
@@ -180,8 +180,9 @@ static fpga_result send_uafu_event_request(fpga_handle handle,
 	struct _fpga_handle *_handle = (struct _fpga_handle *)handle;
 	struct fpga_port_info port_info  = {.argsz = sizeof(port_info),
 						.flags = 0 };
-	struct fpga_port_uafu_irq_set uafu_irq = {.argsz = sizeof(uafu_irq),
-						.flags = 0};
+	uint8_t uafu_irq_buf[sizeof(struct fpga_port_uafu_irq_set)+sizeof(int32_t)];
+	struct fpga_port_uafu_irq_set *uafu_irq =
+		(struct fpga_port_uafu_irq_set *) uafu_irq_buf;
 
 	if (uafu_operation != FPGA_IRQ_ASSIGN && uafu_operation != FPGA_IRQ_DEASSIGN) {
 		FPGA_ERR("Invalid UAFU operation requested");
@@ -200,18 +201,21 @@ static fpga_result send_uafu_event_request(fpga_handle handle,
 			return FPGA_INVALID_PARAM;
 		}
 
+		uafu_irq->argsz = sizeof(struct fpga_port_uafu_irq_set);
+		uafu_irq->flags = 0;
+
 		if (uafu_operation == FPGA_IRQ_ASSIGN) {
-			uafu_irq.evtfd[0] = fd;
-			uafu_irq.start = flags;
+			uafu_irq->evtfd[0] = fd;
+			uafu_irq->start = flags;
 			_eh->flags = flags;
 		} else {
-			uafu_irq.start = _eh->flags;
-			uafu_irq.evtfd[0] = -1;
+			uafu_irq->start = _eh->flags;
+			uafu_irq->evtfd[0] = -1;
 		}
 
-		uafu_irq.count = 1;
+		uafu_irq->count = 1;
 
-		if (ioctl(_handle->fddev, FPGA_PORT_UAFU_SET_IRQ, &uafu_irq) != 0) {
+		if (ioctl(_handle->fddev, FPGA_PORT_UAFU_SET_IRQ, uafu_irq) != 0) {
 			FPGA_ERR("Could not set eventfd");
 			return FPGA_EXCEPTION;
 		}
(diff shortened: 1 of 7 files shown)


libopae/src/event.c
```diff
@@ -180,8 +180,9 @@ static fpga_result send_uafu_event_request(fpga_handle handle,
 	struct _fpga_handle *_handle = (struct _fpga_handle *)handle;
 	struct fpga_port_info port_info  = {.argsz = sizeof(port_info),
 						.flags = 0 };
-	struct fpga_port_uafu_irq_set uafu_irq = {.argsz = sizeof(uafu_irq),
-						.flags = 0};
+	uint8_t uafu_irq_buf[sizeof(struct fpga_port_uafu_irq_set)+sizeof(int32_t)];
```
r-rojo: Should this be initialized to zeroes?

tests/common_test.cpp
```diff
@@ -651,4 +656,27 @@ void BaseFixture::TestAllFPGA(fpga_objtype otype, bool loadbitstream,
...
+
+  if (!real_mock_enable_irq) {
+    std::cerr << "dlsym(\"mock_enable_irq\") failed: (NULL fn pointer)" << std::endl;
+    std::cerr << "Be sure that libmock.so is loaded for mock tests." << std::endl;
+    throw std::logic_error("mock_enable_irq is NULL");
+  }
+
+  return real_mock_enable_irq(enable);
+}
+
```
r-rojo: Perhaps a little comment above this snippet of code that explains what it is doing...
It took me a little while to figure it out ⌛️

tests/function/gtEvent.cpp
```diff
@@ -290,7 +290,11 @@ TEST_F(LibopaecEventFCommonMOCKHW, event_drv_06) {
   poll_fd.events  = POLLIN | POLLPRI;
   poll_fd.revents = 0;
 
-  res = poll(&poll_fd, 1, 1000);
+  do
+  {
+    res = poll(&poll_fd, 1, 1000);
+    ASSERT_GE(res, 0);
+  } while(res == 0);
```
r-rojo: Is there a chance of ever getting "stuck" here? Should we include a timeout period here or leave that up to gtest?

tests/mock.c
```diff
@@ -203,22 +211,22 @@ int ioctl(int fd, unsigned long request, ...)
...
+
+			FILE* hashfile = fopen(hashfilename, "w");
+			if (hashfile) {
+		                fwrite(&hash, sizeof(hash), 1, hashfile);
+				fclose(hashfile);
+			}
+			retval = 0;
+			errno = 0;
+			break;
+		case FPGA_FME_PORT_ASSIGN:
```
r-rojo: Good formatting fix... 👍 
I was considering doing this when I started to look at enabling more event tests.

tests/function/gtEvent.cpp
```diff
@@ -301,3 +305,139 @@ TEST_F(LibopaecEventFCommonMOCKHW, event_drv_06) {
   EXPECT_EQ(FPGA_OK, fpgaUnregisterEvent(m_AFUHandle, FPGA_EVENT_POWER_THERMAL,
                                          m_EventHandles[0]));
 }
+
+class LibopaecEventFCommonMOCK : public LibopaecEventFCommonMOCKHW {
+ protected:
+  virtual void SetUp() {
+    MOCK_enable_irq(true);
```
r-rojo: Will these tests be totally different from those that use fpgad? If not, can we combine them to be parameterized?

tests/unit/gtReconf.cpp
```diff
@@ -68,7 +68,7 @@ TEST(LibopaecReconfCommonMOCKHW, gbs_reconf_01) {
 	fpga_token tok = &_tok;
 
 
-	EXPECT_EQ(FPGA_INVALID_PARAM, set_afu_userclock(h, usrlclock_high, usrlclock_low));
+	//EXPECT_EQ(FPGA_INVALID_PARAM, set_afu_userclock(h, usrlclock_high, usrlclock_low));
```
r-rojo: ❓

libopae/src/event.c
tswhison: I think so.  Good call.

tests/common_test.cpp
tswhison: Sure

tests/function/gtEvent.cpp
tswhison: Here I'm fixing the case where we get a spurious wake event.  If we get stuck, then we have a test failure.  I suppose we could implement a counter-based approach (max number of loops).

tests/function/gtEvent.cpp
tswhison: Yes, they are significantly different to warrant a different fixture - they can't be run on true hardware, because we cannot easily force the driver to signal an interrupt event.

tests/unit/gtReconf.cpp
tswhison: This portion of the test case is invalid and was causing a segfault due to the uninitialized handle h.  Isaac hit this too, and filed an issue on it.

tswhison:
@r-rojo re: negative tests - could you elaborate?

r-rojo:
@tswhison , maybe trying invalid parameter combinations?
Something like trying to register for FPGA_EVENT_INTERRUPT with an FPGA_DEVICE object?
Or maybe trying to register for FPGA_EVENT_POWER_THERMAL?

PR #533, opened 2018-06-27
fpgad: fix uninitialized variable bug

tswhison:
Initialize local variables in poll_ap_event() so that errant event
status is no longer reported.
Fix typos in printed strings.


tools/base/fpgad/ap_event.c
```diff
@@ -86,7 +86,7 @@ static int poll_ap_event(struct fpga_ap_event *event)
 	}
 
 	if (event->ap1_last_event != AP1_STATE && ap1_event == 1) {
-		dlog("AP1 Triggerd for socket %d\n", event->socket);
+		dlog("AP1 Triggered for socket %d\n", event->socket);
```
r-rojo: 😆

PR #535, opened 2018-06-27
tests: Add Error injection support in mock driver

anandaravuri:
Changes:
  -FPGA mock driver  ioctl returns with error code


tests/mock.c
```diff
@@ -180,6 +189,14 @@ int ioctl(int fd, unsigned long request, ...)
...
 				goto out_EINVAL;
 			}
+
+			// Returns error when Error injection enabled
+			if (gEnableErrInj) {
+				retval = EINVAL;
+				errno = EINVAL;
+				break;
+			}
+
```
r-rojo: Can all of these be simplified by moving this check up to around line 168?
If the flag is set, then we can simply call
```C
goto out_EINVAL;
```

tests/common_test.cpp
```diff
@@ -692,4 +692,40 @@ bool MOCK_enable_irq(bool enable)
...
+* would fail.
+*
+* Instead, the test cases that use mock rely on libmock.so being
+* preloaded (ie, we expect the below MOCK_enable_errinj() to throw an
+* exception if libmock.so is not actually loaded). This preserves the
+* restriction that we do not hard link against libmock.so, but still
+* allows us to access the mock API when needed.
+*/
+typedef bool(*mock_enable_errinj_t)(bool);
+bool MOCK_enable_errinj(bool enable) {
```
r-rojo: can we rename this to something that relates to bad ioctl calls?
something like `MOCK_enable_ioctl_err`?

r-rojo:
seems like you may have introduced formatting change in mock.c. The diff is too big now. Can you discard those other (unrelated) changes?

anandaravuri:
formatting changes got removed

PR #539, opened 2018-06-28
pyopae: add pycontext to setup.py

r-rojo:
Installation from package will fail without compiling this as this file
implements functions needed for context management.

### DIFF
--- a/pyopae/setup.py
+++ b/pyopae/setup.py
@@ -34,6 +34,7 @@ class pybind_include_dirs(object):
 extensions = [
     Extension("opae.fpga._opae",
               sources=["pyproperties.cpp",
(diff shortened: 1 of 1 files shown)


lzhan55:
Is compilation needed for installation of the package?

r-rojo:
Short answer, yes.
The long anser:
Our cmake build system builds the extension module using cmake only. This is done to allow building in different build types (Debug, Coverage, Release, ...).
The cmake build system includes another step that processes the setup.py file to create both:
* A source distribution (as a tar.gz file)
* A universal wheel package which includes the compiled extension module.

When I was testing the binary distribution I noticed it was missing some symbols (the ones provided by that pycontext file).

PR #541, opened 2018-06-28
mock: fix build warnings

tswhison:
Fix build warnings regarding unused result of write().

### DIFF
--- a/tests/mock.c
+++ b/tests/mock.c
@@ -306,7 +306,9 @@ int ioctl(int fd, unsigned long request, ...)
 			if (gEnableIRQ && fme_irq->evtfd >= 0) {
 				uint64_t data = 1;
 				// Write to the eventfd to signal one IRQ event.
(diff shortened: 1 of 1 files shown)


tests/mock.c
```diff
@@ -306,7 +306,9 @@ int ioctl(int fd, unsigned long request, ...)
 			if (gEnableIRQ && fme_irq->evtfd >= 0) {
 				uint64_t data = 1;
 				// Write to the eventfd to signal one IRQ event.
-				write(fme_irq->evtfd, &data, sizeof(data));
+				if (write(fme_irq->evtfd, &data, sizeof(data)) != sizeof(data)) {
+					FPGA_ERR("IRQ write < 8 bytes");
```
r-rojo: `!=` is no the same as `<`

tswhison: True, but it would't be greater and it's not equal, so it must be less.. :)

PR #544, opened 2018-06-28
Create fpga-dma shared library

buie1:
This PR will create the `libfpga-dma.so` library in the lib64 directory.
This is one step to our DCP hopefully being able to remove duplicated code from the fpgabist tool, opencl mmd and afu repositories.


southerngs:
I don't think we should be including a gzipped tar file in this repo.  We probably need to figure out how we want to distribute libfpga-dma.so before we use start linking to it.

r-rojo:
What is in the tar file?

deepakcu84:
Agreed. The tar file should be removed. I believe it was accidentally added to the repo.

southerngs:
Assuming that we are removing the tar file, then how do we plan to distribute libfpga-dma.so?  I think we need to know how users will get access to that file before we add steps to opae-sdk that use it.

deepakcu84:
So, the proposal is that we let cmake build and install libfpga-dma.so under the same path where other OPAE shared libraries are installed (<path to opae install>/lib64/). Do you see any problem with this approach? Can you clarify what you mean by "how do we plan to distribute libfpga-dma.so"

southerngs:
The problem I see with this is that we are adding a component to opae-sdk that is not part of core OPAE functionality, and that we plan to distribute separately.  If in the future we have the DMA available as a standalone BBB then how will users choose which version of libfpga-dma.so to use?  How do they choose between the one that would come with opae-sdk and the one in DMA BBB?

deepakcu84:
Our goal is to keep the driver in one place. I am open to keeping it in OPAE or BBB repo as long as we only maintain a single copy. It doesn't really matter.

I see two possibilities.

1. Keep just one version of the DMA driver and that version resides in OPAE.
If we were to distribute DMA as a BBB, software libraries for BBB driver will pull sources from OPAE.  
The nice thing about this flow is OpenCL and AFU/BBB repo are already dependent on OPAE, so keeping the common dependency in OPAE is a natural choice. This flow is also more straightforward as we just need to generate a shared lib from the driver source  that's already out there. This flow is somewhat similar to NLB, where we keep the hardware sources in BBB/AFU repo, but refer the application to the version in OPAE.

2.  Keep just one version of the DMA driver, but that version resides in the BBB/AFU repo. Both BIST and OpenCL will have to refer to that copy by linking shared lib. Two downsides I see with this flow are that 1) it introduces external repo dependencies in OPAE and 2) we'll need the DMA BBB to go public before the refactoring can happen because BIST can't refer to the internal repo.

While I agree DMA isn't a core feature, it is quite similar to how HSSI is to OPAE. OPAE already creates shared libs for HSSI.

Do you have any alternate suggestions?

southerngs:
@deepakcu84 I think that's a good summary of the options.

I'm in favor of packaging DMA driver with the BBB, despite the downsides, but I'd also be interested to hear if anyone else has thoughts on this issue.

r-rojo:
I think I would opt for keeping it with the BBB (similar to how the current BBS include SW).
If, however, we decide it makes more sense in keeping it in OPAE, then I would suggest packaging that up separably from the base libraries. It can either be bundled in one of the extra tools or if it is more standalone, then perhaps we can look at creating an "libs-extra" package.

deepakcu84:
Noted. As I said, I don't mind keeping the driver in BBB repo. We'll have to wait till DMA makes to public BBB repo before BIST can be updated.

deepakcu84:
One advantage of keeping driver in the BBB repo is that hw and sw sources can be kept together (like other BBBs). Remember that for NLB, hw sources are kept in AFU repo, but sw app is located in OPAE. This special structure required us to write and document custom simulation scripts (and they don't look pretty). I'd try to keep everything consistent unless there is a strong reason not to do so.

buie1:
Closing PR. There is a larger conversation to be had about where to put the DMA afu code, etc. May revisit in the future.

PR #550, opened 2018-06-30
Fix issues found during static analysis

r-rojo:


libopae/src/event.c
```diff
@@ -618,14 +618,17 @@ fpga_result __FPGA_API__ fpgaRegisterEvent(fpga_handle handle,
 	struct _fpga_token *_token;
 	int err;
 
-	result = handle_check_and_lock(_handle);
-	if (result)
-		return result;
-
+	// lock event handle first as other event API functions only lock the
```
tswhison: We should follow the same ordering as other places which lock both a handle and an event_handle (handle first, followed by the event_handle).

libopae/src/event.c
```diff
@@ -680,14 +683,17 @@ fpga_result __FPGA_API__ fpgaUnregisterEvent(fpga_handle handle,
 	struct _fpga_event_handle *_eh = (struct _fpga_event_handle *) event_handle;
 	struct _fpga_token *_token;
 
-	result = handle_check_and_lock(_handle);
-	if (result)
-		return result;
-
+	// lock event handle first as other event API functions only lock the
```
tswhison: Same comment as above re: the ordering.  Not sure why this order would avoid a deadlock.

r-rojo:
@tswhison, Yeah, I thought about this some more afterwards and realized that this might not be the correct fix.

The deadlock I'm trying to avoid is a scenario where one thread has a handle mutex locked and then tries to lock an event handle mutex while another thread has the same event handle mutex locked and then tries to lock the same handle mutex.
So thinking about this some more, I can't imagine such a scenario seeing as even handle mutexes are released before locking an event handle mutex. Thoughts?

PR #552, opened 2018-07-02
fpgad: fix issue identified by checkers

tswhison:


tools/base/fpgad/ap6.c
```diff
@@ -277,7 +277,7 @@ void *ap6_thread(void *thread_context)
 	uint32_t num_matches = 0;
 
 	struct bitstream_info null_gbs_info ;
-	memset(&null_gbs_info, 0, sizeof(null_gbs_info));
+	memset_s(&null_gbs_info, sizeof(null_gbs_info), 0);
```
r-rojo: No need to check return value?

tswhison: I don't think so.  &null_gbs_info will always be non-NULL.

PR #553, opened 2018-07-02
Reverting changes due to #546

dasm:
Reverted uppercases back to lowercases.


dasm:
Something is still not right.

dasm:
Updated head. Now it is working.

PR #554, opened 2018-07-02
hssi: override `assign` fn to cfg i2c

r-rojo:
Implment e100::assign method so that it instantiates an i2c instance in 2-byte
byte-address mode as the i2c controller used for AK uses 2 bytes for the byte
address field.

Also update the copyright text for the e100.cpp file.


tools/extra/integrated/hssi/e100.cpp
```diff
@@ -1,5 +1,31 @@
+// Copyright(c) 2018, Intel Corporation
```
tswhison: If this was part of a prior release that went out last year, you could use 2017-2018.  Looks good.

PR #555, opened 2018-07-02
tests: Add clearing mock errors in setup/teardown

r-rojo:
Implement SetUp/TearDown functions for mock event tests to write 0 to the mock
errors file for more repeatable tests involving mock sysfs.


r-rojo:
Going to hold off as the real issue is a race condition with our current mock tests (and no sort of IPC synchronization).

PR #562, opened 2018-07-05
pyopae: add device_id property to Python API

r-rojo:
Also reword some of the doc strings

### DIFF
--- a/pyopae/pyproperties.cpp
+++ b/pyopae/pyproperties.cpp
@@ -371,7 +371,7 @@ void properties_set_vendor_id(properties::ptr_t props, uint32_t vendor_id) {
 const char *properties_doc_device_id() {
   return R"opaedoc(
     Get or set the device ID  property of a resource.
-    The device ID is part of the PCI ID and is assibned by the
+    The device ID is part of the PCI ID and is assigned by the
     vendor.
    )opaedoc";
 }
--- a/pyopae/pyproperties.cpp
+++ b/pyopae/pyproperties.cpp
@@ -63,31 +63,33 @@ const char *properties_doc_get() {
 
       type (fpga_objtype): The object type - DEVICE or ACCELERATOR.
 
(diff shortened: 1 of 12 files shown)


pyopae/pyproperties.cpp
```diff
@@ -363,6 +367,23 @@ void properties_set_vendor_id(properties::ptr_t props, uint32_t vendor_id) {
   props->vendor_id = vendor_id;
 }
 
+// device id
+const char *properties_doc_device_id() {
+  return R"opaedoc(
+    Get or set the device ID  property of a resource.
+    The device ID is part of the PCI ID and is assibned by the
```
tswhison: assigned

r-rojo: doh!
Good catch!

pyopae/pyproperties.cpp
```diff
@@ -63,31 +63,33 @@ const char *properties_doc_get() {
...
 
-      vendor_id: The vendor ID in PCI config space.
+      vendor_id (uint16_t): The vendor ID in PCI config space.
+
+      device_id (uint16_t): The device ID in PCI config space.
 
       model (str): The model of the FPGA.
 
-      local_memory_size: The size (in bytes) of the FPGA local memory.
+      local_memory_size (uint64_t): The size (in bytes) of the FPGA local memory.
```
tswhison: Do we support this property currently?  Last I remembered, it was a TODO item.

r-rojo: It looks like it's still not supported 😞 
In this case, the C++ backing code will throw a `not_supported` exception which will get converted into a Python RuntimeError with the message generated by the `not_supported` exception.
I think this is fine for now but we should probably look at either defining analogous exception classes in Python or convert the C++ exceptions to standard Python exceptions that make sense for the situation.

PR #578, opened 2018-07-18
cxx-core: add segment property to properties class

r-rojo:


libopaecxx/src/properties.cpp
```diff
@@ -38,6 +38,7 @@ properties::properties()
     : props_(nullptr),
       type(&props_, fpgaPropertiesGetObjectType, fpgaPropertiesSetObjectType),
       num_errors(&props_, fpgaPropertiesGetNumErrors, fpgaPropertiesSetNumErrors),
+      segment(&props_, fpgaPropertiesGetSegment, fpgaPropertiesSetSegment),
```
asgibbons: I don't see where these methods are defined.  Is that a different commit?

tests/function/gtCxxProperties.cpp
```diff
@@ -144,3 +144,16 @@ TEST(LibopaecppPropsCommonALL, get_num_errors) {
...
+ * @test segment
+ * Given a properties properties object with the segment property set to a
+ * known value
+ * When I get the segment property
+ * Then the number is the expected value
+ */
+TEST(LibopaecppPropsCommonALL, get_segment) {
+  auto p = properties::get();
+  p->segment = 9090;
+  EXPECT_EQ(static_cast<uint16_t>(p->segment), 9090);
```
asgibbons: So, you're assigning a value to a variable, then expecting it to be equal to the value you just assigned to it?  I don't see where this is actually testing anything, but there may be some C++ magic here somewhere.  Can you clarify?

libopaecxx/src/properties.cpp
r-rojo: which methods? fpgaProperties[GS]etSegment?

tests/function/gtCxxProperties.cpp
r-rojo: line 157 is exercising the assignment operator in the pvalue template class which will ultimately call fpgaPropertiesSetSegment.
line 158 is exercising the conversion operator in the pvalue template classs which will call fpgaPropertiesGetSegment to retrieve the value.

PR #580, opened 2018-07-20
pyopae: Add `segment` property to Python API

r-rojo:


pyopae/pyproperties.cpp
```diff
@@ -63,6 +62,8 @@ const char *properties_doc_get() {
 
       type (fpga_objtype): The object type - DEVICE or ACCELERATOR.
 
+      segment (uint16_t) : The PCIe segment (or domain) numer.
```
tswhison: number

PR #582, opened 2018-07-23
Remove json flag from afu_synth_setup example

stefanstoeckigt:
The following example doesn't work with the --json flag, not supported by afu_synth_setup.


lzhan55:
Hi Abelardo, could you take a look at this?

PR #583, opened 2018-07-23
Scottgi/fpgainfo c

asgibbons:
Initial "complete" implementation.


tools/base/fpgainfo-c/powerinfo.c
```diff
@@ -108,23 +113,28 @@ static fpga_result get_power_info(fpga_token tok, struct power_info *finfo)
...
+	       "\n");
+}
+
+static void print_power_info(fpga_properties props)
+{
+	fpga_result res = FPGA_OK;
+
+	fpgainfo_print_common("//****** POWER ******//", props);
+
+	res = bmc_print_values(get_sysfs_path(props, FPGA_DEVICE, NULL),
```
r-rojo: Will this only work on DCP platform?

asgibbons: This reads IPMI-formatted binary from a sysfs file.  If other platforms provide it, then it will work.  But for now, it's A10 (Rush Creek) specific.

tools/base/fpgainfo-c/bmcdata.c
```diff
@@ -0,0 +1,726 @@
...
+			 L"packets",
+			 L"messages",
+			 L"characters",
+			 L"error",
+			 L"correctable error",
+			 L"uncorrectable error",
+			 L"fatal error",
+			 L"grams"};
+
+size_t max_base_units = (sizeof(base_units) / sizeof(base_units[0]));
```
r-rojo: This will probably get flagged by static analysis.

tools/base/fpgainfo-c/bmcdata.c
```diff
@@ -0,0 +1,726 @@
...
+				    "Discrete (07h)",
+				    "'digital' Discrete (08h)",
+				    "'digital' Discrete (09h)",
+				    "Discrete (0Ah)",
+				    "Discrete (0Bh)",
+				    "Discrete (0Ch)"};
+
+
+size_t max_event_reading_type_code = (sizeof(event_reading_type_codes)
+				      / sizeof(event_reading_type_codes[0]));
```
r-rojo: This will probably get flagged by static analysis

tools/base/fpgainfo-c/bmcdata.c
```diff
@@ -0,0 +1,726 @@
...
+			       "Real-Time Clock",
+			       "Reserved",
+			       "Air Inlet",
+			       "Reserved",
+			       "Air Inlet",
+			       "Processor / CPU",
+			       "Baseboard / Main System Board"};
+
+size_t max_entity_id_code =
+	(sizeof(entity_id_codes) / sizeof(entity_id_codes[0]));
```
r-rojo: This will probably get flagged by static analysis

tools/base/fpgainfo-c/bmcinfo.c
```diff
@@ -247,63 +62,44 @@ fpga_result bmc_filter(fpga_properties *filter, int argc, char *argv[])
...
 {
 	Values *val = (Values *)calloc(1, sizeof(Values));
 
+	(void)header;
+
 	if (NULL == val)
 		return NULL;
 
-	memset(val, 0, sizeof(val));
+	memset(val, 0, sizeof(*val));
```
r-rojo: Doesn't `calloc` already zero the memory? If so, this isn't necessary.

r-rojo:
I'm also wondering if it is worthwhile including a "strings table"  for string management. What are your thoughts on this?

tools/base/fpgainfo-c/bmcinfo.c
asgibbons: Good catch.

asgibbons:
@r-rojo The only wide character support in safe_strings is for wmemcpy_s and wmemmove_s.  The reason for wide characters is to emulate the Python version's printing of the degree symbol.

As far as a strings table is concerned, it's pretty complicated in vanilla "C", so unless we can use C++ I'd say it's a no-go.  We'll probably need to re-factor fpgainfo for Darby Creek (and MCP?) anyway, so we could consider adding it at that time.

tools/base/fpgainfo-c/tempinfo.c
```diff
@@ -23,35 +23,50 @@
...
-	snprintf_s_ss(path, sizeof(path), "%s/%s", dev->sysfspath,
-		      "thermal_mgmt/temperature");
-	res = sysfs_read_u32(path, &temperature);
+	// BMC thermal sensors are more accurate than sysfs - don't print sysfs
+#ifdef PRINT_SYSFS_TEMPERATURE
+	char path[SYSFS_PATH_MAX];
+	uint32_t temperature = -1;
+	snprintf_s_ss(path, sizeof(path), "%s/%s", sysfspath,
+		      SYSFS_THERMAL_FILE);
+	res = fpgainfo_sysfs_read_u32(path, &temperature);
```
buie1: Is there anyway that temperature can have another API call for temperature? Seems like it would make sense imo
@r-rojo

asgibbons: This was in response to an email - apparently the BMC sensor has less variability than the board's sensor.  PM didn't want to display the value for the less accurate sensor.

I think OPAE is working on a metrics API that would potentially replace all of this sysfs stuff, if that's what you mean.

asgibbons:
@r-rojo I believe it's ready to merge.  What is remaining to do?

PR #589, opened 2018-07-26
Add generic APIs for reading/writing objects

r-rojo:
Add six APIs for reading and writing from/to an object in an FPGA
resource. The objects are identified by a string key.
For Linux, this is equivalent to reading/writing to sysfs
files with 'key' as a path relative to the resource directory in sysfs.

For each operation (read, write), three functions are defined.
* One for operating on a 32-bit value
* One for operating a 64-bit value
* One for operating on a sequence of bytes

### DIFF
--- a/common/include/opae/sys.h
+++ b/common/include/opae/sys.h
@@ -97,43 +97,49 @@ fpga_result fpgaReadObjectBytes(fpga_token token, const char *key,
  * @brief Write a 32-bit value to an FPGA resource object as identified by a
  * string key. On Linux, this is equivalent to writing to a sysfs node with
  * 'key' as the path relative to the sysfs directory of the resource identified
- * by the token.
+ * by the handle.
  *
- * @param[in] token Token identifying a resource (accelerator or device)
+ * @param[in] handle Handle to an open resource (accelerator or device)
  * @param[in] key A key identifying the object with respect to the resource
  * @param[in] value The 32-bit value to write to the object
  *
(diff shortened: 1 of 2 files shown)


r-rojo:
@asgibbons , Yes, key is a path relative to either of those sysfs directories.
@tswhison , Should we add a mutex to the token or do you think opening a resource is required in order to operate on its sysfs files?

tswhison:
I think opening the resource is required.  Say that you have a token, but unbeknownst to you someone else has a handle to the same resource.  I think this could cause collisions, unless we're banking on the kernel serializing the sysfs access.  Conceptually, you could read from a device you don't own, but I think writing is definitely off limits.

asgibbons:
@r-rojo , What would Windows expect?  Does it require an open resource before having the data available?

r-rojo:
@tswhison , So maybe we can still keep token in the API but just be sure the implementation of reading and writing is different:
* read operations use token as is
* write operations use token to open a handle (in shared mode?)

@asgibbons , I'm not familiar with the Windows implementation but we can probably impose similar restrictions (open handle for writing operations)

asgibbons:
@rrojo , Differing behavior is OK with me.

Is the default mode for fpgaOpen SHARED?  If not, the open will fail.

r-rojo:
I've changed the function signatures for write operations to require a handle.
This can enforce more thread safety and avoid doing opens that may result in failure.

PR #592, opened 2018-07-27
Add sys.c (for generic sysfs operations)

r-rojo:
Implment the functions for the fpgaObject API for reading/writing from
generic resource objects.
On Linux, this translates to reading/writing sysfs files relative to the
resource's sysfs path rooted at:

/sys/class/fpga/intel-fpga-dev.\d/intel-fpga-(fme|port).\d

This also adds a unit test executable that compiles in mock.c.
TODO:
* Implement the (Read|Write)Bytes functions
* Add more unit tests

### DIFF
--- a/libopae/src/sys.c
+++ b/libopae/src/sys.c
@@ -35,9 +35,11 @@
 #include "safe_string/safe_string.h"
 
 #define NULL_CHECK(var)                                                        \
-	if (var == NULL) {                                                     \
-		FPGA_MSG(#var " is NULL");                                     \
-	}
+	do {                                                                   \
+		if (var == NULL) {                                             \
+			FPGA_MSG(#var " is NULL");                             \
+		}                                                              \
+	} while (false);
 
 fpga_result cat_token_sysfs_path(char *dest, fpga_token token, const char *path)
 {
--- a/libopae/src/sys.c
+++ b/libopae/src/sys.c
@@ -1,0 +1,158 @@
+// Copyright(c) 2017-2018, Intel Corporation
+//
+// Redistribution  and  use  in source  and  binary  forms,  with  or  without
+// modification, are permitted provided that the following conditions are met:
+//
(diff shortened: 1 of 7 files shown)


libopae/src/sys.c
```diff
@@ -0,0 +1,158 @@
+// Copyright(c) 2017-2018, Intel Corporation
```
tswhison: Different name for the file, perhaps?  We already have sysfs.c.

r-rojo:
@tswhison, Yeah. I was thinking that too. I'll rename the files in a future commit.

libopae/src/sys.c
```diff
@@ -0,0 +1,160 @@
...
+	NULL_CHECK(len);
+	(void)offset;
+
+	char objpath[SYSFS_PATH_MAX];
+	fpga_result res = cat_token_sysfs_path(objpath, token, key);
+	if (res) {
+		return res;
+	}
+
+
```
asgibbons: I'm assuming this still needs work?

libopae/src/sys.c
```diff
@@ -0,0 +1,160 @@
...
+	NULL_CHECK(key);
+	NULL_CHECK(buffer);
+	(void)offset;
+	(void)len;
+	char objpath[SYSFS_PATH_MAX];
+	fpga_result res = cat_handle_sysfs_path(objpath, handle, key);
+	if (res) {
+		return res;
+	}
+	return res;
```
asgibbons: And this...

tswhison:
I think this is intentional.  The read/write bytes functions were called out as TODO. @r-rojo?

r-rojo:
@asgibbons , Yes. This is incomplete. I've called out what else needs to be done in the TODO.
This is the reason why I'm doing this on feature branch. Those TODO items can be done on this feature branch.

PR #593, opened 2018-07-31
Rename sys file, implement RW Bytes ops

r-rojo:
* Rename sys.h and sys.c to sysobject.h and sysobject.c
* Implement fpgaWriteObjectBytes, and fpgaReadObjectBytes functions
* Also change fn signature of fpgaWriteObjectBytes to use len output
param
* Rename gtunit in CMake to gtunit-sysobject

### DIFF
--- a/common/include/opae/sysobject.h
+++ b/common/include/opae/sysobject.h
@@ -143,7 +143,8 @@ fpga_result fpgaWriteObject64(fpga_handle handle, const char *key,
  * @param[in] key A key identifying the object with respect to the resource
  * @param[in] buffer Pointer to memory to write to the the object
  * @param[in] offset Where to start writing bytes in the object
- * @param[in] len The number of bytes to write
+ * @param[in] len The number of bytes to read. This is updated with the
+ * actual number of bytes read
  *
  *
  * @return FPGA_OK on success. FPGA_INVALID_PARAM if any of the supplied
@@ -154,7 +155,7 @@ fpga_result fpgaWriteObject64(fpga_handle handle, const char *key,
  * possible race conditions
  */
 fpga_result fpgaWriteObjectBytes(fpga_handle handle, const char *key,
-				 uint8_t *buffer, size_t offset, size_t len);
+				 uint8_t *buffer, size_t offset, size_t *len);
 
 #ifdef __cplusplus
 } // extern "C"
--- a/common/include/opae/sysobject.h
+++ b/common/include/opae/sysobject.h
@@ -80,15 +80,20 @@ fpga_result fpgaReadObject64(fpga_token token, const char *key,
  *
  * @param[in] token Token identifying a resource (accelerator or device)
  * @param[in] key A key identifying the object with respect to the resource
- * @param[out] buffer Pointer to memory to store the object's value
+ * @param[out] buffer Pointer to memory to store the object's bytes - a null
+ * pointer indicates a request for the total number of bytes required
  * @param[in] offset Start of byte sequence to read
- * @param[in] len The number of bytes to read. This is updated with the
+ * @param[out] len The number of bytes to read. This is updated with the
  * actual number of bytes read
  *
  *
  * @return FPGA_OK on success. FPGA_INVALID_PARAM if any of the supplied
  * parameter is invalid. FPGA_NOT_SUPPORTED if this function is not supported
  * by the current implementation of this API.
+ *
+ * @note A null output buffer can be used to get the total number of bytes the
+ * caller should allocate relative to the offset. A null buffer and an offset
+ * of 0 will get the total number of bytes used by the object.
  */
 fpga_result fpgaReadObjectBytes(fpga_token token, const char *key,
 				uint8_t *buffer, size_t offset, size_t *len);
--- a/common/include/opae/sysobject.h
+++ b/common/include/opae/sysobject.h
@@ -155,7 +155,7 @@ fpga_result fpgaWriteObject64(fpga_handle handle, const char *key,
  * possible race conditions
  */
 fpga_result fpgaWriteObjectBytes(fpga_handle handle, const char *key,
-				 uint8_t *buffer, size_t offset, size_t *len);
+				 uint8_t *buffer, size_t offset, size_t len);
 
 #ifdef __cplusplus
 } // extern "C"
--- a/common/include/opae/sysobject.h
+++ b/common/include/opae/sysobject.h
@@ -83,7 +83,7 @@ fpga_result fpgaReadObject64(fpga_token token, const char *key,
  * @param[out] buffer Pointer to memory to store the object's bytes - a null
  * pointer indicates a request for the total number of bytes required
  * @param[in] offset Start of byte sequence to read
- * @param[out] len The number of bytes to read. This is updated with the
+ * @param[inout] len The number of bytes to read. This is updated with the
  * actual number of bytes read
  *
  *
--- a/libopae/src/sysobject.c
+++ b/libopae/src/sysobject.c
@@ -28,6 +28,9 @@
 #include <config.h>
 #endif // HAVE_CONFIG_H
 
+#include <sys/types.h>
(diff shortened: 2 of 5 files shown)


libopae/src/sysobject.c
```diff
@@ -98,20 +101,38 @@ fpga_result __FPGA_API__ fpgaReadObjectBytes(fpga_token token, const char *key,
...
+				 strerror(errno));
+		} else {
+			FPGA_ERR("Offset after lseek < offset argument");
+		}
+		return FPGA_EXCEPTION;
+	}
+	*len = read(fd, buffer, *len);
 
-	return res;
+	return FPGA_OK;
```
asgibbons: Need to close the file.

libopae/src/sysobject.c
```diff
@@ -144,17 +165,37 @@ fpga_result __FPGA_API__ fpgaWriteObject64(fpga_handle handle, const char *key,
...
+			FPGA_ERR("Error with lseek operation: %s",
+				 strerror(errno));
+		} else {
+			FPGA_ERR("Offset after lseek < offset argument");
+		}
+		return FPGA_EXCEPTION;
+	}
+	*len = write(fd, buffer, *len);
+
+	return FPGA_OK;
```
asgibbons: Need to close the file.

tests/unit/gtsys.cpp
```diff
@@ -38,7 +38,7 @@ extern "C" {
 #include "gtest/gtest.h"
 #include "common_test.h"
 
-TEST(sys, read32) {
+TEST(sysobject, read32) {
```
tswhison: More tests coming in a future commit?

r-rojo: Yes, more tests coming

r-rojo:
@asgibbons, I changed the ReadBytes function to use partial read.
According to Linux kernel documentation on sysfs, doing a partial read with offset of 0 will cause the kernel driver to fill its buffer. Would it be useful to include a flag in the ReadBytes function that will force a buffer reload by seeking to zero?

libopae/src/sysobject.c
```diff
@@ -98,18 +101,42 @@ fpga_result __FPGA_API__ fpgaReadObjectBytes(fpga_token token, const char *key,
...
 	if (res) {
 		return res;
 	}
 
+	fd = open(objpath, O_RDONLY);
+	if (fd < 0) {
+		return FPGA_NOT_FOUND;
+	}
+
+	count = pread(fd, buffer, *len, offset);
```
tswhison: Why not loop here if the call fails and errno == EINTR?  man pread/pwrite says that the fn's can fail and set any error code that read/write can set.

libopae/src/sysobject.c
```diff
@@ -144,17 +171,41 @@ fpga_result __FPGA_API__ fpgaWriteObject64(fpga_handle handle, const char *key,
...
+	res = cat_handle_sysfs_path(objpath, handle, key);
 	if (res) {
 		return res;
 	}
+	fd = open(objpath, O_WRONLY);
+	if (fd < 0) {
+		return FPGA_NOT_FOUND;
+	}
+
+	count = pwrite(fd, buffer, *len, offset);
```
tswhison: Same comment as above.  Loop when pwrite fails with EINTR.

libopae/src/sysobject.c
```diff
@@ -146,15 +173,84 @@ fpga_result __FPGA_API__ fpgaWriteObjectBytes(fpga_handle handle,
...
+
+
+	rw_buffer = calloc(fsize, sizeof(char));
+	if (rw_buffer == NULL) {
+		res = FPGA_NO_MEMORY;
+		goto out_close;
+	}
+
+	// copy bytes to write to rw_write buffer
+	memcpy(rw_buffer + offset, buffer, len);
```
tswhison: probably memcpy_s

asgibbons:
@r-rojo The scenario is this:  I read a header (4-10 bytes) that contains the length of the record following.  I then read the first structure (size is count from header) immediately following the header.  Then (immediately after the record) there is another header with size.  This continues until EOF.  With ReadBytes, there will be an open, seek, read, close sequence for each block, which could be very expensive.

Any way I could just get an fd returned?  I could then do reads at will.

asgibbons:
A strong argument can be made that this *does* comply - the SDR sysfs file contains an array of sensor description records, and the sensors file contains an array of sensor readings.  Each of these files is an array of the same type.  The only sticky part is that the type has variable-length constituents.  This is not unlike the sysfs file ["/sys/devices/system/node/node0/numastat"](/sys/devices/system/node/node0/numastat) which displays an array of statistics for NUMA interactions.  The lines within this file are variable-length, but all of the same "type."

I guess I'd be OK with the API returning the entire file contents, which I could subsequently parse.  It should probably have a mechanism to return the size of the buffer required so we can be efficient with memory allocation.  It should read the minimum of the <number of bytes in the file> or <the passed-in size>.

r-rojo:
@asgibbons, Hmm... I guess it's sort of a grey area... having multiple headers within one file is the tricky part. So as far as this API goes, what do you think should be the agreement?
* Caller allocates and frees buffer
  This requires the caller to have some method in the API to determine how much to allocate
* API allocates and frees buffer
  The API function ReadBytes allocates the buffer and returns the size in the output parameter.
  Another API function is made to free
* API allocates but caller frees
  The API function ReadBytes allocates the buffer and returns the size in the output parameter.
  The caller simply calls free on the buffer

Another idea is that we could introduce an additional opaque type to represent device attributes. 
```C
fpgaGetAttribute(token|handle, key, fpga_attribute*);
fpgaDestroyAttribute(tfpga_attribute*);
fpgaReadAttribute64(fpga_attribute, uint64_t*);
fpgaReadAttributeByes(fpga_attribute, uint8_t*, count, offset);
```

Thoughts?

asgibbons:
@r-rojo I would prefer option 1, as I believe it's best for the caller to manage the memory.  I would recommend we follow fpgaEnumerate's lead - fpgaReadObjectBytes() gets passed a NULL buffer pointer and it returns the file size perhaps?

common/include/opae/sysobject.h
```diff
@@ -80,15 +80,20 @@ fpga_result fpgaReadObject64(fpga_token token, const char *key,
  *
  * @param[in] token Token identifying a resource (accelerator or device)
  * @param[in] key A key identifying the object with respect to the resource
- * @param[out] buffer Pointer to memory to store the object's value
+ * @param[out] buffer Pointer to memory to store the object's bytes - a null
+ * pointer indicates a request for the total number of bytes required
  * @param[in] offset Start of byte sequence to read
- * @param[in] len The number of bytes to read. This is updated with the
+ * @param[out] len The number of bytes to read. This is updated with the
```
asgibbons: This is actually '[in/out]'.

libopae/src/sysobject.c
```diff
@@ -41,6 +45,38 @@
 		}                                                              \
 	} while (false);
 
+static inline ssize_t eintr_pread(int fd, void *buf, size_t count, size_t offset)
+{
+	ssize_t bytes_read = 0, total_read = 0;
+	char *ptr = buf;
+	while (total_read < (ssize_t)count) {
+		bytes_read = pread(fd, ptr + total_read, count, offset);
```
tswhison: Don't you need..
  count - total_read
  offset + total_read

libopae/src/sysobject.c
```diff
@@ -41,6 +45,38 @@
...
+	}
+	return total_read;
+}
+
+static inline ssize_t eintr_write(int fd, void *buf, size_t count)
+{
+	ssize_t bytes_written = 0, total_written = 0;
+	char *ptr = buf;
+	while (total_written < (ssize_t)count) {
+		bytes_written = write(fd, ptr + total_written, count);
```
tswhison: Don't you need..
  count - total_written

PR #594, opened 2018-08-01
Adding bus support to hello_fpga

trenduchintala:
Enabling the user to target a bus when using a multi card system. 

Example of usage:
hello_fpga -B 04
hello_fpga --bus 0x04


samples/hello_fpga.c
```diff
@@ -106,6 +113,132 @@ void print_err(const char *s, fpga_result res)
...
+		const char *tmp_optarg = optarg;
+		/* Checks to see if optarg is null and if not it goes to value of optarg */
+		if ((optarg) && ('=' == *tmp_optarg)){
+			++tmp_optarg;
+		}
+
+		switch (getopt_ret){
+		case 'B': /* bus */
+			if (NULL == tmp_optarg)
+				break;
```
asgibbons: This is an error condition, as there is a required argument for -B.  Should print an error and exit instead of break and continue.

samples/hello_fpga.c
```diff
@@ -106,6 +113,132 @@ void print_err(const char *s, fpga_result res)
...
+
+out:
+	return err_cnt;
+}
+
+	
+/* function to get the bus number when there are multiple buses */
+struct bus_info{
+	uint8_t bus;
+};
```
asgibbons: Why is this a struct?  You should replace this struct with a uint8_t bus only, unless you suspect it will be expanded in the future (most likely not).

samples/hello_fpga.c
```diff
@@ -115,7 +248,8 @@ int main(int argc, char *argv[])
 	fpga_token         accelerator_token;
```
asgibbons: filter is unused.  See comment below.

samples/hello_fpga.c
```diff
@@ -126,56 +260,50 @@ int main(int argc, char *argv[])
...
-	/* TODO: Add selection via BDF / device ID */
-
-	res = fpgaEnumerate(&filter, 1, &accelerator_token, 1, &num_matches);
-	ON_ERR_GOTO(res, out_destroy_prop, "enumerating accelerators");
-
-	if (num_matches < 1) {
+	res = find_fpga(guid, &accelerator_token, &num_matches_accelerators, &num_matches_fpga);
+	if (num_matches_accelerators <= 0) {
 		fprintf(stderr, "accelerator not found.\n");
 		res = fpgaDestroyProperties(&filter);
```
asgibbons: filter has not been referenced and is NULL.

samples/hello_fpga.c
trenduchintala: I left it as a struct just in case we would like to print information about the device and function if needed in the future.

PR #598, opened 2018-08-03
Remove 32-bit sysfs write operations

r-rojo:
* Remote 32-bit sysfs write operations
  Generic 32-bit sysfs write operations are not supported yet.
  The kernel documentation for sysfs recommends reading entire
  contents of a sysfs file, modifying it with the intended bytes, and then
  writing back the buffer. Rather that implementing that for 32-bit
  objects, this removes this operation for now.
* Rename mock unit test executable

### DIFF
--- a/common/include/opae/sysobject.h
+++ b/common/include/opae/sysobject.h
@@ -99,26 +99,6 @@ fpga_result fpgaReadObjectBytes(fpga_token token, const char *key,
 				uint8_t *buffer, size_t offset, size_t *len);
 
 /**
- * @brief Write a 32-bit value to an FPGA resource object as identified by a
- * string key. On Linux, this is equivalent to writing to a sysfs node with
- * 'key' as the path relative to the sysfs directory of the resource identified
- * by the handle.
- *
- * @param[in] handle Handle to an open resource (accelerator or device)
- * @param[in] key A key identifying the object with respect to the resource
- * @param[in] value The 32-bit value to write to the object
- *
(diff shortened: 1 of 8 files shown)


r-rojo:
@asgibbons, that seems to go in line with the sysfs documentation:
```
- The buffer will always be PAGE_SIZE bytes in length. On i386, this
  is 4096. 
```
Let me take a look at how I can refactor this taking this into consideration.

asgibbons:
@r-rojo Another thing to consider is to add globbing to the file name.  I find this is needed for the bmc code, as the directory under which the BMC sensors are located is `avmmi-bmc.*.auto/bmc_info/sensors`, where `*` is a number.  I would recommend the following:

	#include <glob.h>

	glob_t pglob;
	int gres = glob(sysfspath, GLOB_NOSORT, NULL, &pglob);
	if ((gres) || (1 != pglob.gl_pathc)) { // Only allow single match
		globfree(&pglob);
		return FPGA_NOT_FOUND;
	}

	fd = open(pglob.gl_pathv[0], O_RDONLY);

This would allow for wildcard usage in pathnames.

r-rojo:
@asgibbons , Is this number the same as the device instance id? Would it look something like the following?
```
.../intel-fpga-dev.<number>/intel-fpga-fme.<number>/avmmi-bmc.<number>.auto/bmc_info/sensors
```

asgibbons:
@r-rojo No.  The numbers are different.  I don't know the scheme.

r-rojo:
@asgibbons , I'd like to check in this PR and address the requested change in a new PR

PR #599, opened 2018-08-06
Remove old RAS error strings.

michael-adler:
Strings were moved to other structures but the old instances remained.
New C++ compilers trigger errors on the unused strings.


michael-adler:
gcc 6.1.0 failed to compile the module because the strings I removed defined but not used. It appears that the removed strings are remnants of previous error logging since the same strings are present in other, newer data structures in the same module.

PR #603, opened 2018-08-08
Lzhan55/ase refactor9

lzhan55:
- Removed unneeded code


michael-adler:
Looks cleaner. What's the seed used for?

lzhan55:
Seed is used for random number generation. We don't really need to save the seed in a file.

michael-adler:
Right, I forgot. Last time I tried I couldn't get ASE to produce the same result from run to run. Perhaps I was setting the seed wrong. Does it work now if you force a seed, producing identical results each time?

lzhan55:
The seed was either read from ase_config file, or randomly set based on the current time. If you set the seed in config file, it should be consistent.

PR #604, opened 2018-08-08
Add `instance` to internal _fpga_token

r-rojo:
Add an instance number field to the internal sturcture, _fpga_token.
Parse that from sysfs path during enumeration and store it in the token.
This can then be used in other calls that require formatting paths using
that number.


libopae/src/token_list.c
```diff
@@ -57,6 +57,22 @@ struct _fpga_token *token_add(const char *sysfspath, const char *devpath)
...
+	/* get the device instance id */
+	if (ptr == NULL) {
+		FPGA_MSG("sysfspath does not meet exptected format");
+		return NULL;
+	}
+
+	num = strtoul(ptr, &endptr, 10);
+	/* no digits in path */
+	if (num == 0 && endptr == ptr) {
+		FPGA_MSG("sysfspath does not meet exptected format");
```
tswhison: expected

libopae/src/types_int.h
```diff
@@ -99,6 +99,7 @@
 
 /** System-wide unique FPGA resource identifier */
 struct _fpga_token {
+  uint32_t instance;
```
tswhison: indentation is off

libopae/src/token_list.c
```diff
@@ -57,6 +57,22 @@ struct _fpga_token *token_add(const char *sysfspath, const char *devpath)
 	struct token_map *tmp;
 	errno_t e;
 	int err = 0;
+	uint32_t num = 0;
+	char *endptr = NULL;
+	const char *ptr = strrchr(sysfspath, '.');
+
+	/* get the device instance id */
+	if (ptr == NULL) {
+		FPGA_MSG("sysfspath does not meet exptected format");
```
tswhison: expected

PR #607, opened 2018-08-15
Dunnikri/add tcl to qsys tags

deepakcu84:
Fixes the following issues in ASE simulation flow involving Qsys systems:

1. Avoid manipulating file system hierarchy when copying over qsys systems.   

When copying over Qsys systems, generate_ase_environment.py adds a suffix (qsys_sim/target_dir_<id>/*) to the original directory string which breaks Qsys systems that rely on relative paths to locate instantiated IP/other qsys systems.

2. Sub directories under the location where .qsys files are located are not copied over in generate_ase_environment, breaking generation.

3. Qsys systems that instantiate other IPs require their sources (*_hw.tcl format) in the simulation directory. Adds support in rtl_src_config and generate_ase_environment to copy over TCL sources.

TCL files should be specified in filelist.txt,  with the tag “SI:”. rtl_src_config emits TCL with other simulation sources. They get filtered and copied into the simulation directory.


michael-adler:
The existing suffix (.tcl) is effectively a tag, too. I think your change already causes .tcl files to be emitted by rtl_src_config only when --tcl is specified, right? If so, you've already eliminated them from Quartus builds and SI: is redundant (unnecessary). You just need to gather the .tcl files using --tcl in generate_ase_environment. You're basically doing that already and, I suspect, don't need the extra check for the suffix in that script.

deepakcu84:
Re "I think your change already causes .tcl files to be emitted by rtl_src_config only when --tcl is specified, right? If so, you've already eliminated them from Quartus builds"

Not really. All files except the ones with "SI:" tag got added to QII compile flow. Here's what I tried

# Emit sources and Quartus/simulator includes
for c in cfg:
     if(tag SI: exists)
              print(sim file)
     elif(tag QII: exists)
              print(QII file)
     else
            if(opts.sim)
                  print(sim file that exists in sim tag map) 
            elif(opts.tcl) // causes QII compile to pick up the TCL
                   print(tcl file)
             ....

I didnt see a switch to explicitly select the QII compile flow. i,e QII compile is implied by the absence of --sim switch. Did I miss something?

michael-adler:
No, you're right. I forgot that .tcl files could be QII files outside of Qsys, so have to be included in the QII group. You've defined the category of SI:*.tcl to be files that need to be handled specially for simulation, which is fine.

I still think you can simplify the new generate_ase_environment code, taking advantage of the fact that rtl_src_config will filter out just the tcl files for you.

deepakcu84:
Ok. Let me look at it.

deepakcu84:
@michael-adler @OPAE/ase can you approve now? All travis checks passed.

deepakcu84:
Thanks Michael and Tracy.

PR #612, opened 2018-08-22
Add PLATFORM_INTENDED_DEVICE_FAMILY to generated platform configuration.

michael-adler:
The value is derived from the fpga-family encoded in the platform
database and may be passed to megafunctions as "intended_device_family"
to get proper simulated semantics for a target architecture.


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=2617494
This change would be an all-or-nothing failure, so the couple of Qsys failures are something in my environment.

PR #614, opened 2018-08-22
initial commit to tests refactor

r-rojo:
This adds:
* testing/mock/mock.c which implements system calls like `open`, `close`, `ioctl`...
* testing/mock/test_system.*
* testing/mock/c_test_system.h
* testing/unit/munit-enum-c.cpp

testing/mock/mock.c calls C functions declared in c_test_system.h.
Those functions are implemented in test_system.cpp and call into
test_system singleton object for calls like open and opendir.

Unit tests (in testing/unit directory) are to call the test_system to
initialize it. For example, munit-enum-c.cpp initializes the test system
and also prepares a mock sysfs directory (as a subdirectory of cwd).

Calls destined to sysfs paths that are normally handled by the kernel
driver are handled by test_system. Calls to open devices in /dev, are
stashed in a map, keyed off the file descriptor, and are used when ioctl
is called with them.


libopae/src/enum.c
```diff
@@ -74,9 +74,13 @@ struct dev_list {
 	struct dev_list *fme;
 };
 
-static bool matches_filter(const struct dev_list *attr,
-			   const fpga_properties filter)
+bool matches_filter(const struct dev_list *attr, const fpga_properties filter)
 {
+	if (!filter) {
```
tswhison: This is caught by matches_filters() in the case that num_filters is 0.  (there is logic that makes sure that filters is NULL if num_filters is 0 just inside fpgaEnumerate.  It is not technically an error for filter to be NULL.  That is the "match all" case.  You're right that filter mustn't be NULL here, but the code is structured so that we don't need to check.

testing/CMakeLists.txt
```diff
@@ -0,0 +1,82 @@
+cmake_minimum_required (VERSION 2.8)
+
+# projectname is the same as the main-executable
+project(testing)
+find_package(GTest)
+
+If(${GTest_FOUND})
+  message(STATUS Found GTest)
+else()
+  Build_GTEST()
```
ibarraza: I think you have to add a "include(test_config)" in order for Build_GTEST() to be visible.

PR #619, opened 2018-08-27
add events unit tests

r-rojo:
* add new executable, munit-opae-c-events which compiles in some files
from the events api and fpgad
* add basic events unit test (parameterized) fixture that controls fpgad
server_thread thread

### DIFF
--- a/testing/unit/events_test.cpp
+++ b/testing/unit/events_test.cpp
@@ -1,0 +1,133 @@
+// Copyright(c) 2017-2018, Intel Corporation
+//
+// Redistribution  and  use  in source  and  binary  forms,  with  or  without
+// modification, are permitted provided that the following conditions are met:
+//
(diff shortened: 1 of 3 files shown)


testing/unit/events_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+  std::array<fpga_token, 2> tokens_;
+  fpga_handle handle_;
+  uint32_t num_matches_;
+  test_platform platform_;
+  test_system *system_;
+  std::thread fpgad_;
+  fpga_event_handle eh_;
+};
+
+TEST_P(events_p, register_event) {
```
tswhison: Seems like we had a few more tests.  Adding them later?

r-rojo: yes

PR #620, opened 2018-08-27
Update mock device object model

r-rojo:
* Add sysclass path (/sys/class/fpga/intel-...) to c'tor
* Add device_id to c'tor
* Add accessor methods to get the values mentioned above
* Change the ioctl function to be virtual with the default
implementation throwing an exception if called
* Change the test_system::open call to figure out if the file being
opened is a driver attribute or a device file and use the appropriate
c'tor for creating the mock device
* Add a regex class in test_utils.h that wraps the POSIX.2 regex API


jelonanX:
Inconsistent device id type definition between source code sysfs.h (uint64_t) and mock test_system.h (uint16_t)

r-rojo:
The API defines it as 32-bit. I'll change it to 32-bit to be consistent with the API.

PR #622, opened 2018-08-28
Jelonanx/tests refactor

jelonanX:
Improved sysfs.c coverage. 
Imported common_test namespace from old mock driver to help unit tests.

### DIFF
--- a/testing/CMakeLists.txt
+++ b/testing/CMakeLists.txt
@@ -25,7 +25,6 @@ set(MUNIT_SRC mock/mock.c
               ${CMAKE_SOURCE_DIR}/libopae/src/common.c
               ${CMAKE_SOURCE_DIR}/libopae/src/enum.c
               ${CMAKE_SOURCE_DIR}/libopae/src/error.c
-              ${CMAKE_SOURCE_DIR}/libopae/src/event.c
               ${CMAKE_SOURCE_DIR}/libopae/src/hostif.c
               ${CMAKE_SOURCE_DIR}/libopae/src/init.c
               ${CMAKE_SOURCE_DIR}/libopae/src/manage.c
@@ -54,6 +53,7 @@ include_directories(
     ${CMAKE_SOURCE_DIR}/common/include
     ${PROJECT_BINARY_DIR}/include
     ${CMAKE_SOURCE_DIR}/libopae/src
+    ${CMAKE_SOURCE_DIR}/tools/base
     mock
     ${GTEST_INCLUDE_DIRS}
     {CMAKE_CURRENT_SOURCE_DIR}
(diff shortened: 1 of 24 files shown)


testing/mock/test_system.cpp
```diff
@@ -52,10 +52,9 @@ mock_object::mock_object(const std::string &devpath,
       device_id_(device_id),
       type_(type) {}
 
-int mock_fme::ioctl(int request, va_list argp) {
-  (void)request;
-  (void)argp;
-  return 0;
+  int mock_fme::ioctl(int request, va_list argp) {
+    (void)request;
+    (void)argp;
```
r-rojo: This should return  0

testing/mock/common_test.h
```diff
@@ -0,0 +1,390 @@
+// Copyright(c) 2017, Intel Corporation
```
r-rojo: I would prefer not to include these two files as is (common_test.h|cpp)

testing/mock/test_system.h
```diff
@@ -30,7 +30,7 @@
 #include <map>
 #include <string>
 #include <vector>
-
+#include "opae/fpga.h"
```
r-rojo: Not needed here

testing/unit/munit-sysfs.cpp
```diff
@@ -29,169 +29,433 @@
...
+	EXPECT_NE(result, FPGA_OK);
+
+	// Pass Port handle insted of FME handle
+	result = get_fpga_deviceid(NULL, &deviceid);
+	EXPECT_NE(result, FPGA_OK);
+
+	//result = fpgaOpen(tok, &h, 0);
+	//EXPECT_EQ(FPGA_OK, result) << "This is the result: \t" << result;
+
+	token_for_afu0(&_tok);
```
r-rojo: I would prefer not to use this for now

testing/mock/test_system.h
jelonanX: I could be wrong but I need to include that to define fpga_version and fpga_accelerator type inside test_device struct. There is probably other ways to do this.

testing/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ set(MUNIT_SRC mock/mock.c
               unit/properties_test.cpp
               unit/munit-wsid_list.cpp
               unit/munit-sysfs.cpp
+              unit/munit-mmio-c.cpp
```
ibarraza: This file does not exist. Did you mean to add him?

testing/mock/test_system.h
r-rojo: Sorry, it should be <opae/fpga.h>

PR #624, opened 2018-08-29
Lzhan55/ase cov2

lzhan55:
- Added ASE unit test application to test ASE simulator internal functions 
- Added test cases for ase_ops.c , error_report.c and ase_strings.c


cmake/modules/coverage.cmake
```diff
@@ -152,7 +152,7 @@ function(set_target_for_coverage_local target_name)
   add_custom_target(${name}
 
     # Cleanup lcov
-    COMMAND ${LCOV_EXECUTABLE} --directory . --zerocounters
+    # COMMAND ${LCOV_EXECUTABLE} --directory . --zerocounters
```
nakulkorde: Since this file is used to generate coverage for other parts for the project as well, do you think not resetting the counters can have a negative effect? Do we need to have a separate version of this just for ASE?

lzhan55: @nakulkorde I think it should still work. @abelardojarab , can you comment on this?

ase/sw/ase_common.h
```diff
@@ -441,15 +443,30 @@ bool remove_existing_lock_file(char *);
...
 void ase_buffer_t_to_str(struct buffer_t *, char *);
 void ase_str_to_buffer_t(char *, struct buffer_t *);
 int ase_dump_to_file(struct buffer_t *, char *);
 uint64_t ase_rand64(void);
 void ase_eval_session_directory(void);
 int ase_instance_running(void);
-void remove_spaces(char *);
-void remove_tabs(char *);
+
+#ifdef __cplusplus
```
abelardojarab: shouldnt't the 'extern c_plusplus wrap all the functions in this file? 

i dont see any function on this file which it is implemented in a C++ implementation file..

cmake/modules/coverage.cmake
abelardojarab: mm i think best should be add a flag whose default behavior (if flag is not specified) is not to flush coverage counters.

ase/sw/ase_common.h
abelardojarab: yes i understand the reason but can we enclose the complete set of function prototypes in this file in 'extern c_pluspus' since i think all of these functions are implemented in C, no one is implemented with C++

cmake/modules/coverage.cmake
abelardojarab: sounds good

ase/api/src/mmio.c
```diff
@@ -50,13 +50,14 @@ fpga_result __FPGA_API__ fpgaWriteMMIO32(fpga_handle handle,
...
 {
 	UNUSED_PARAM(mmio_num);
-	struct _fpga_handle *_handle = (struct _fpga_handle *) handle;
 
 	if (NULL == handle) {
 		FPGA_MSG("handle is NULL");
 		return FPGA_INVALID_PARAM;
 	}
 
+	struct _fpga_handle *_handle = (struct _fpga_handle *) handle;
```
abelardojarab: nice catch, very nice.

ase/api/src/event.c
```diff
@@ -51,16 +50,12 @@ fpga_result __FPGA_API__ fpgaCreateEventHandle(fpga_event_handle *handle)
 	_eh->fd = eventfd(0, 0);
 	if (_eh->fd < 0) {
 		FPGA_ERR("eventfd : %s", strerror(errno));
-		result = FPGA_NOT_FOUND;
-		goto out_free;
+		free(_eh);
```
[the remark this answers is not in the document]
lzhan55: @abelardojarab I was trying to improve the code coverage for this by moving the free(_eh) here.

lzhan55:
ase_coverage_status

lzhan55:
Previously ASE coverage data was: 
se_coverage_pre

lzhan55:
@michael-adler For OPAE API tests with ASE, it used the same gtests for OPAE. The additional ASE unit test of gtAseU was added just for ASE simulater internal code..

PR #625, opened 2018-08-30
Add mmio unit test

jelonanX:
Migrate tests from old system to this new mock infrastructure.

### DIFF
--- a/scripts/cover.sh
+++ b/scripts/cover.sh
@@ -40,6 +40,7 @@ make munit-opae-c munit-opae-c-events
 
 lcov --directory . --zerocounters
 lcov -c -i -d . -o coverage.base
-./bin/munit-opae-c --gtest_filter="enum*:buffer*:*properties*:sysfs*"
+./bin/munit-opae-c --gtest_filter="mmio*"
+#./bin/munit-opae-c --gtest_filter="enum*:buffer*:*properties*:sysfs*"
 ./bin/munit-opae-c-events
 
--- a/testing/unit/munit-mmio-c.cpp
+++ b/testing/unit/munit-mmio-c.cpp
@@ -1,0 +1,126 @@
+// Copyright(c) 2017-2018, Intel Corporation
+//
+// Redistribution  and  use  in source  and  binary  forms,  with  or  without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of  source code  must retain the  above copyright notice,
+//   this list of conditions and the following disclaimer.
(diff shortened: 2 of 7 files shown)


scripts/cover.sh
```diff
@@ -40,6 +40,7 @@ make munit-opae-c munit-opae-c-events
 
 lcov --directory . --zerocounters
 lcov -c -i -d . -o coverage.base
-./bin/munit-opae-c --gtest_filter="enum*:buffer*:*properties*:sysfs*"
+./bin/munit-opae-c --gtest_filter="mmio*"
```
tswhison: only mmio?

testing/unit/munit-mmio-c.cpp
```diff
@@ -0,0 +1,141 @@
...
+    struct fpga_port_region_info *rinfo = va_arg(argp, struct fpga_port_region_info *);
+    if (!rinfo) {
+      FPGA_MSG("rinfo is NULL");
+      goto out_EINVAL;
+    }
+    if (rinfo->argsz != sizeof(*rinfo)) {
+      FPGA_MSG("wrong structure size");
+      goto out_EINVAL;
+    }
+    if (rinfo->index != 0) {
```
tswhison: Index can be 0 or 1 for the AFU.  1 == signal tap

scripts/cover.sh
r-rojo: I think this was done because he was testing only mmio. I don't think this script is final. It was meant as something to get up and running with coverage quickly.

testing/unit/munit-mmio-c.cpp
```diff
@@ -0,0 +1,141 @@
...
+};
+
+TEST_P (mmio_c_p, test_map_mmio) {
+  uint64_t * mmio_ptr = NULL;
+  EXPECT_TRUE(((struct _fpga_handle*)handle_)->mmio_root == NULL);
+
+  ASSERT_EQ(FPGA_OK, fpgaMapMMIO(handle_, 0, &mmio_ptr));
+  EXPECT_FALSE(mmio_ptr == NULL);
+  EXPECT_FALSE(((struct _fpga_handle *)handle_)->mmio_root == NULL);
+  EXPECT_EQ(FPGA_OK, fpgaUnmapMMIO(handle_, 0));
```
r-rojo: These can also be written as
```C++
EXPECT_EQ(lhs, nullptr);
// OR
EXPECT_NE(lhs, nullptr);
```
Which is comparing the left-hand side (lhs) against null.

PR #626, opened 2018-08-30
Add in version check to new test infrastructure 

jelonanX:


testing/unit/munit-version.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+using namespace opae::testing;
+/**
+ * @test       version_01
+ *
+ * @brief      When I retrieve fpga_version information using
+ *             fpgaGetOPAECVersion using a NULL pointer as the return buffer,
+ *             the function returns FPGA_INVALID_PARAM.
+ */
+TEST(version_c_p, version_01) {
```
r-rojo: You should probably remove the _p from the test class (`vercion_c_p`). The `p` is for parameterized tests.
I know you copied from previous tests but let's try to make the test names should be more meaningful.
Like this one, can be:
```C++
TEST(version_c, null_output) {
//...
}
```

testing/unit/munit-version.cpp
```diff
@@ -0,0 +1,154 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#ifdef __cplusplus
+
+extern "C" {
+#endif
+
+#include <opae/fpga.h>
+#include "config_int.h"
```
tswhison: Let's move away from config_int.h to config.h.  All of the OPAE source files have the following block:
#ifdef HAVE_CONFIG_H
#include <config.h>
#endif

PR #630, opened 2018-09-04
Add properties from handle in cxx and Python APIs

r-rojo:
Add a new factory function for creating properties objects from a
handle.
Add a Python binding around that function to get properties from a
handle in Python.


common/include/opae/cxx/core/properties.h
```diff
@@ -101,6 +101,12 @@ class properties {
    */
   static properties::ptr_t get(fpga_token t);
 
+  /** Retrieve the properties for a given handle object.
+   * @param[in] t A handle identifying the accelerator resource.
```
tswhison: The param name is h. :)

r-rojo: Thanks!

PR #632, opened 2018-09-05
Fix incorrect tests for command line arguments in userclk program.

michael-adler:


michael-adler:
I finally looked into why "userclk -S 0" didn't work correctly on a dual-FPGA system. The test for the command line arguments was looking for > 0, which is wrong. I updated the code based on the version in fpgaconf and tested on the various platforms.

PR #633, opened 2018-09-05
Fix errors being treated as warnings in gcc 5

nakulkorde:


tests/function/gtBuffer.cpp
```diff
@@ -215,6 +215,11 @@ TEST(LibopaecBufCommonALL, Prep0B) {
 
   EXPECT_EQ(FPGA_INVALID_PARAM, fpgaReleaseBuffer(h, wsid));
 #endif
+  UNUSED_PARAM(buf_len);
+  UNUSED_PARAM(buf_addr);
+  UNUSED_PARAM(wsid);
+  UNUSED_PARAM(flags);
+
```
r-rojo: Can this be rewritten to something like:
```C++
#ifdef BUILD_ASE
  UNUSED_PARAM(buf_len);
  //...



#else
  buf_len =  0;
  flags = 0;
  // ...
#endif
```

nakulkorde: Yes, I think that can be done.. let me make those changes and push a new commit.

tests/function/gtBuffer.cpp
```diff
@@ -199,22 +199,23 @@ TEST(LibopaecBufCommonALL, Prep0B) {
...
+  uint64_t* buf_addr;
+  uint64_t wsid = 1;
+  int flags;
   buf_len = 0;
   flags = 0;
   EXPECT_EQ(FPGA_INVALID_PARAM,
             fpgaPrepareBuffer(h, buf_len, (void**)&buf_addr, &wsid, flags));
 
   EXPECT_EQ(FPGA_INVALID_PARAM, fpgaReleaseBuffer(h, wsid));
 #endif
```
r-rojo: Nice!

PR #634, opened 2018-09-05
Add unit tests for fpgaOpen and fpgaClose

jelonanX:


testing/unit/munit-open-close.cpp
```diff
@@ -0,0 +1,158 @@
...
+  ASSERT_EQ(FPGA_INVALID_PARAM, res);
+}
+
+/**
+ * @test       open_02
+ *
+ * @brief      When the fpga_handle * parameter to fpgaOpen is NULL, the
+ *             function returns FPGA_INVALID_PARAM.
+ */
+TEST_P(openclose_c_p, open_02) {
```
tswhison: open_01 and open_02 appear to be the same test.  Maybe a copy/paste glitch?  I think open_02 probably needs to verify another condition.

testing/unit/munit-open-close.cpp
```diff
@@ -0,0 +1,158 @@
...
+ * @test       open_05
+ *
+ * @brief      When the flags parameter to fpgaOpen is invalid, the
+ *             function returns FPGA_INVALID_PARAM and FPGA_NO_DRIVER.
+ *
+ */
+TEST_P(openclose_c_p, open_05) {
+  fpga_handle handle_;
+  fpga_result res;
+  struct _fpga_token* _token = (struct _fpga_token*)tokens_[0];
```
tswhison: Let's move away from using the internal data structures - _fpga_token, _fpga_handle etc.  The new plugin-api will bomb on these type of tests.

testing/unit/munit-open-close.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+  res = fpgaOpen(tokens_[0], &handle_, 42);
+  ASSERT_EQ(FPGA_INVALID_PARAM, res);
+
+  _token->magic = FPGA_TOKEN_MAGIC;
+  res = fpgaOpen(tokens_[0], &handle_, FPGA_OPEN_SHARED);
+  ASSERT_EQ(FPGA_OK, res);
+
+  strcpy(_token->devpath,"/dev/intel-fpga-fme.01");
```
tswhison: Did you intend to use intel-fpga-fme.01?  The leading 0 makes the path invalid.

testing/unit/munit-open-close.cpp
jelonanX: good catch.

testing/unit/munit-open-close.cpp
jelonanX: I did that intentionally to cover one of the cases.

testing/unit/munit-open-close.cpp
jelonanX: Actually, the first test tests for the token when is initialized to NULL and the second test for an declared token.

tswhison: I see the difference now.  Actually, open_02 is invalid, because tok is uninitialized.  We cannot protect against this scenario.

testing/unit/munit-open-close.cpp
tswhison: Ok.. We still need to move away from peeking inside the internal structures.  The handles and tokens we get back from the plugin api have an extra layer on top of the _fpga_token and _fpga_handle structures.

testing/unit/munit-open-close.cpp
jelonanX: ok, I will leave this test out.

asgibbons: Just a thought - you *can* test this by:

char tmp[256] = { 0 };
fpga_handle fpga = (fpga_handle) &tmp[0];

The fpgaOpen() checks the signature in the handle.  Just a thought.

tswhison: Actually, if there is no valid pthread_mutex_t in the underlying object, we'll segfault inside pthread_mutex_lock().  That's the root of why we can't guard against uninitialized handles.  There is race condition between the validity of the mutex object and the magic.  Which do you do first, lock the mutex or check the magic?  If you check the magic first, you're not thread safe.  If you lock the mutex first you risk attempting to lock a garbage mutex object..

asgibbons: This is an open, and there's no thread protection on tokens.  Perhaps that is an omission in the token struct, but today, there's no mutex protection present for tokens.

PR #635, opened 2018-09-06
Restructure sysfs unit tests. 

jelonanX:


scripts/cover.sh
```diff
@@ -40,6 +40,6 @@ make munit-opae-c munit-opae-c-events
 
 lcov --directory . --zerocounters
 lcov -c -i -d . -o coverage.base
-./bin/munit-opae-c --gtest_filter="enum*:buffer*:*properties*:sysfs*:mmio*"
-./bin/munit-opae-c-events
 
+./bin/munit-opae-c --gtest_filter="enum*:buffer*:*properties*:sysfs*"
```
ibarraza: Why are we excluding mmio tests here?

testing/unit/munit-sysfs.cpp
```diff
@@ -41,18 +41,12 @@ extern "C" {
...
-#include <map>
-#include <memory>
-#include <string>
-#include <vector>
 #include "gtest/gtest.h"
 #include "test_system.h"
 
 using namespace opae::testing;
 
+using namespace opae::testing;
```
ibarraza: Duplicate namespace.

scripts/cover.sh
jelonanX: It's only excluded for this branch. It will be included in the later PRs.

testing/unit/munit-sysfs.cpp
```diff
@@ -110,29 +128,163 @@ TEST_P(sysfs_c_p, null_device_id){
...
+  //NULL sysfs path 
+  result = get_port_sysfs(h, NULL);
+  EXPECT_NE(result, FPGA_OK);
+  
+  //NULL handle
+  result = get_port_sysfs(NULL, (char*) "/sys/class/fpga/intel-fpga-dev.0/intel-fpga-fme.0/socket_id");
+  EXPECT_NE(result, FPGA_OK);
+  
+  //Invalid handle
+  result = get_port_sysfs(h, (char*)"/sys/class/fpga/intel-fpga-dev.0/intel-fpga-port.0/");
```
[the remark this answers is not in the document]
tswhison: I think line 246 does what Isaac mentioned.

tswhison:
Let's resolve the test cases that are trying to use the handle without initializing it.  Those will definitely be problematic.

Issue #631, opened 2018-09-05
Potential problem with packaged /usr/bin/afu_sim_setup in 1.1.0 release


dwilson1394:
When using the deb package for ASE 1.1.0 (on a 64-bit Ubuntu 16.04 virtual machine) and attempting to use "afu_sim_setup" in the first tutorial in https://github.com/OPAE/intel-fpga-bbb/tree/master/samples/tutorial/01_hello_world, afu_sim_setup throws an IO exception at line 108 indicating **dst** doesn't exist.

https://github.com/OPAE/opae-sdk/blob/0bf28fa812e9be6e4e6a1a25a229667022d37a76/ase/scripts/afu_sim_setup#L103-L110

Upon further investigation, dst doesn't exist because it is never created via **copy_tree()** from **ase_base** which is set to **/home/lab/sys_aalsdk/opae-sdk-x/ase**. This corresponds to a value returned by getASESrcPath() which uses the local variables **project_src_dir** and **ase_share_dir** . In the package's **/usr/bin/afu_sim_setup**, they appear as:

``` python
def getASESrcPath():
    # CMake will update any variables marked by @ with the proper values.
    project_src_dir = '/home/lab/sys_aalsdk/opae-sdk-x/ase'
    ase_share_dir = '/home/lab/sys_aalsdk/opae-sdk-x/ase'
```

Since ase_share_dir is a non-valid full path on my machine, afu_sim_setup won't work since line 68 (`ase_share = os.path.join(parent_dir, ase_share_dir) `) returns **/home/lab/sys_aalsdk/opae-sdk-x/ase**.  

https://github.com/OPAE/opae-sdk/blob/0bf28fa812e9be6e4e6a1a25a229667022d37a76/ase/scripts/afu_sim_setup#L46-L68

I would guess that ASE_SHARE_DIR variable was originally a relative path and some change in the cmake files has changed it to the packager's full path that results in the os.path.join() in being incorrect.

michael-adler:
I'm confused. The user specifies "dst". Is the original complaint that the exception fires because "dst" doesn't exist wrong? Is the reason really that "ase_base" (the source) doesn't exist?

dwilson1394:
Sorry about that, I jumped a few steps there.

[...]

PR #648, opened 2018-09-08
Fix incorrect recording of ASE installation path

michael-adler:
The meaning of ASE_SHARE_DIR changed in commit 575fb262, which broke
afu_sim_setup's ability to find the installed ASE share tree.

This fixes issue #631

### DIFF
--- a/ase/CMakeLists.txt
+++ b/ase/CMakeLists.txt
@@ -45,14 +45,17 @@ set(OPAE_SHARE_DIR     ${OPAE_SDK_SOURCE})
 set(OPAE_SHARE_DIR     ${OPAE_SDK_SOURCE} PARENT_SCOPE) # do not remove, they are not the same
 include(ase_variables_config)
 
+# This is used in scripts to find the relative installed path to ASE files
+set(ASE_INST_SHARE_DIR share/opae/ase)
+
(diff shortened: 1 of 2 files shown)


michael-adler:
The problem appears only when the OPAE SDK source for an installation is _not_ present when afu_sim_setup is invoked. This would happen, e.g., when installing OPAE from an RPM instead of building from source. We are clearly not testing this case well and should.

Confirmed the solution by first making the bug happen in the DCP test environment. I removed the opae_sdk_x tree and ran ASE regressions using current master. Everything broke (http://sw-pert.altera.com/pert/pert.php?test_run_id=2654936). I reran the experiment using the change here and ASE then ran correctly (http://sw-pert.altera.com/pert/pert.php?test_run_id=2654937).

This will have to be cherry picked into all releases derived from 575fb26.

lzhan55:
@michael-adler Will this fix work on machines without OPAE installation such as OPAE development machine? The fix looks good to me. Thanks.

michael-adler:
@lzhan55, I'm not sure what you mean by "without OPAE installation". Do you mean "without OPAE sources"? If so, then yes afu_sim_setup will again work without OPAE SDK sources present after this commit.

PR #650, opened 2018-09-10
Add in optionparser unit tests

jelonanX:


testing/unit/test_optionparser_c.cpp
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * @test       open_06
+ *
+ * @brief      When the fpga_token parameter to fpgaOpen is NULL, the
+ *             function returns FPGA_INVALID_PARAM.
+ */
+//TEST(LibopaecOpenCommonALL, open_06) {
+//  fpga_handle h;
+//
+//  EXPECT_EQ(FPGA_INVALID_PARAM, fpgaOpen(NULL, &h, 0));
```
ibarraza: This is not using the handle. It is populating it.

testing/unit/test_optionparser_c.cpp
```diff
@@ -0,0 +1,88 @@
...
+ *             function returns FPGA_INVALID_PARAM.
+ *
+ */
+//TEST(LibopaecOpenCommonALL, open_08) {
+//  struct _fpga_token _tok;
+//  fpga_token tok = &_tok;
+//  fpga_handle h;
+//
+//  token_for_afu0(&_tok);
+//  EXPECT_EQ(FPGA_INVALID_PARAM, fpgaOpen(tok, &h, 42));
```
ibarraza: This is not using the handle. It is populating it.

PR #658, opened 2018-09-11
Add cpp coding style checker

nakulkorde:
This adds the cpp coding style checker script which uses clang-format with Google's coding style.


scripts/test-codingstyle-cpp.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+pushd $(dirname $0)
+
+FILES=$(find ../libopaecxx/src ../libopaecxx/samples \( -iname "*.cpp" -or -iname "*.h" \))
+clang-format-3.9 -i -style=Google $FILES
+
+if [ "git diff" != NULL ]; then
```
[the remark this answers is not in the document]
nakulkorde: Thanks!

scripts/test-codingstyle-cpp.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+pushd $(dirname $0)
+
+FILES=$(find ../libopaecxx/src ../libopaecxx/samples ../common/include/opae/cxx/core \( -iname "*.cpp" -or -iname "*.h" \))
+clang-format-3.9 -i -style=Google $FILES
+
+if [ x$(git diff) != x ]; then
+   echo "Coding style issues occured. Please fix them based on the following suggestions. You can run clang-format on these files in order to automatically format your code."
```
r-rojo: You might want to rephrase to something like:
```
Coding style check failed. Please fix ...
```

nakulkorde: yes I have tested it and I got the same output as with Travis. Since clang format suggests changes to the files as seen here: https://travis-ci.org/OPAE/opae-sdk/jobs/427459877, my PR will not go through unless they are modified. @r-rojo @tswhison  what do you think should be done since you have previously used clang-format on these files?

nakulkorde:
yes I have tested it and I got the same output as with Travis. Since clang format suggests changes to the files as seen here: https://travis-ci.org/OPAE/opae-sdk/jobs/427459877, my PR will not go through unless they are modified. @r-rojo @tswhison  what do you think should be done since you have previously used clang-format on these files?

r-rojo:
you can run clang-format on the files and fix them 😁

PR #662, opened 2018-09-12
plugin-api: fix coding style errors

tswhison:


tswhison:
I've got the style checkers passing now, it's just that the tools are failing to build (for known reasons).

PR #676, opened 2018-09-14
plugin-api: refactor properties API

tswhison:
Make the property get/set routines of the properties API local
functions of opae-c, because they don't need to be plugin-enabled.
Remove the notion of wrapped properties from opae-c. Adjust the
test cases accordingly.


libopae/api-shell.c
```diff
@@ -123,36 +119,15 @@ fpga_result fpgaGetPropertiesFromHandle(fpga_handle handle,
...
-		res = FPGA_NO_MEMORY;
-		dres = wrapped_handle->adapter_table->fpgaDestroyProperties(
-			&pr);
-	}
-
-	*prop = wrapped_properties;
 
-	return res != FPGA_OK ? res : dres;
+	return wrapped_handle->adapter_table->fpgaGetPropertiesFromHandle(
+		wrapped_handle->opae_handle, prop);
```
r-rojo: Don't forget to wrap the parent token if present in the properties.

libopae/api-shell.c
```diff
@@ -164,1123 +139,34 @@ fpga_result fpgaGetProperties(fpga_token token, fpga_properties *prop)
...
-
-		return wrapped_properties->adapter_table
-			->fpgaPropertiesSetNumErrors(
-				wrapped_properties->opae_properties,
-				num_errors);
-	}
-
-	return opae_properties_set_num_errors(
-		wrapped_properties->opae_properties, num_errors);
+		wrapped_token->opae_token, prop);
```
r-rojo: Don't forget to wrap the parent token if present in the properties.

libopae/plugins/xfpga/properties.c
```diff
@@ -57,9 +57,8 @@ xfpga_fpgaGetPropertiesFromHandle(fpga_handle handle, fpga_properties *prop)
...
 }
 
-
-fpga_result __FPGA_API__
-xfpga_fpgaGetProperties(fpga_token token, fpga_properties *prop)
+fpga_result __FPGA_API__ xfpga_fpgaGetProperties(fpga_token token,
+						 fpga_properties *prop)
 {
 	struct _fpga_properties *_prop;
 	fpga_result result = FPGA_OK;
```
r-rojo: Should we call the API `fpgaGetProperties` with a null token to allocate the properties structure?

tswhison: Solved by commit 'Fix code review comments.'

libopae/api-shell.c
```diff
@@ -123,36 +124,33 @@ fpga_result fpgaGetPropertiesFromHandle(fpga_handle handle,
...
+			res = FPGA_NO_MEMORY;
+		} else {
+			sres = fpgaPropertiesSetParent(*prop, wrapped_parent);
+		}
 	}
 
-	*prop = wrapped_properties;
-
-	return res != FPGA_OK ? res : dres;
+	return res != FPGA_OK ? res : sres;
```
r-rojo: Uhh... this is a little confusing to me..  how do you feel about something like:
```C
res = wrapped_h->at->fpgaGetPropertiesFromHandle(...);
if (res) {
        return res;
}
if (fpgaPropertiesGetParent(*prop, &parent) == FPGA_OK) {
        opae_wrapped_token *wp = opae_allocate_wrapped_token(...);
        if (!wp) {
            OPAE_ERR(...);
            return FPGA_NO_MEMORY;
        }
        return fpgaPropertiesSetParent(*prop, wp);
}
return res;
```

libopae/api-shell.c
```diff
@@ -164,1123 +162,53 @@ fpga_result fpgaGetProperties(fpga_token token, fpga_properties *prop)
...
-			res = FPGA_NO_MEMORY;
-			dres = wrapped_token->adapter_table
-				       ->fpgaDestroyProperties(&pr);
+			if (!wrapped_parent) {
+				OPAE_ERR("malloc failed");
+				res = FPGA_NO_MEMORY;
+			} else {
+				sres = fpgaPropertiesSetParent(pr,
+							       wrapped_parent);
+			}
```
r-rojo: Perhaps we can refactor this routine into one function?

libopae/api-shell.c
r-rojo: But I don't see the change here ... do we need to do it here too?

libopae/api-shell.c
tswhison: thanks!

libopae/api-shell.c
tswhison: thanks!

libopae/api-shell.c
tswhison: thanks!

tswhison:
@r-rojo Thanks for pointing out those omissions.  Does it look ok now?

PR #680, opened 2018-09-17
fix fpgaObjectwrite64 to use 64 bit hex format

r-rojo:


tswhison:
There is one thing being pointed out by the coding style checker..
../libopae/plugins/xfpga/sysobject.c:205: ERROR: do not use assignment in if condition

PR #681, opened 2018-09-17
Add `-j 4` in make command to speen up build

r-rojo:


asgibbons:
Why restrict this to 4 cores?  Using '-j' with no argument will use all available resources.

r-rojo:
Oops, forgot about this.

PR #687, opened 2018-09-18
Fix missing newlines in a few ASE error messages.

michael-adler:


michael-adler:
@lzhan55 please cherry pick this at least to release/1.1.1. (I didn't actually check, but assume the bug wound up in that release.)

PR #688, opened 2018-09-18
plugin-api: add platform detection

tswhison:
Add platform detection for MCP/DCP (xfpga).


libopae/init.c
```diff
@@ -108,11 +109,17 @@ __attribute__((constructor)) static void opae_init(void)
 	if (g_logfile == NULL)
 		g_logfile = stdout;
 
-	fpgaInitialize(NULL);
+	// If the environment hasn't requested explicit initialization,
+	// perform the initialization implicitly here.
+	if (getenv("OPAE_EXPLICIT_INITIALIZE") == NULL)
```
r-rojo: Now that I think about it some more, what do you think 💭 about making this a compile time check instead of a runtime check?

tswhison: I'm torn.  Let's propose the problem internally and solicit advice.

PR #693, opened 2018-09-19
tests: augment the malloc intercept

tswhison:
Add an optional count and symbol name to test_system::invalidate_malloc().
This will cause malloc to fail after the Nth call when the return
address is that of the given string. Note that this will not work
for inlined functions, irrespective of whether inlining is disabled.
Add fpgaOpen() test case that uses the functionality.

### DIFF
--- a/testing/mock/test_system.cpp
+++ b/testing/mock/test_system.cpp
@@ -66,7 +66,7 @@ void *malloc(size_t size) {
         dladdr(caller, &info);
         if (!info.dli_sname)
             res = 1;
-	else
+        else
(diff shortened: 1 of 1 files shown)


testing/mock/test_system.cpp
```diff
@@ -28,19 +28,56 @@
...
+
+    } else {
+        void *caller = __builtin_return_address(0);
+        int res;
+        Dl_info info;
+
+        dladdr(caller, &info);
+        if (!info.dli_sname)
+            res = 1;
+	else
```
r-rojo: Is your alignment off?

tswhison:
pesky tabs

PR #694, opened 2018-09-19
Add open close unit tests

jelonanX:


testing/xfpga/test_open_close_c.cpp
```diff
@@ -51,7 +100,13 @@ class openclose_c_p
   }
 
   virtual void TearDown() override {
-    EXPECT_EQ(fpgaDestroyProperties(&filter_), FPGA_OK);
+    EXPECT_EQ(xfpga_fpgaDestroyProperties(&filter_), FPGA_OK);
```
ibarraza: I don't think this function exists.

PR #695, opened 2018-09-19
Add error unit tests

jelonanX:


testing/xfpga/test_error_c.cpp
```diff
@@ -0,0 +1,546 @@
...
+      std::string cmd = "rm -rf " + tmpsysfs_;
+      std::system(cmd.c_str());
+    }
+    system_->finalize();
+  }
+
+  std::string tmpsysfs_;
+  fpga_properties filter_;
+  std::array<fpga_token, 2> tokens_ = {};
+  fpga_handle handle_;
```
ibarraza: I don't think handle_ or tokens_ are even used in these tests, is that right?

tswhison:
Test build issue.  Let's get this sorted out and the tests checked in.

PR #697, opened 2018-09-20
tests: scrub opae properties tests for leaks

tswhison:
Used valgrind memcheck to reveal resource leaks in the
properties tests for opae-c. Fixed 3..
1) fpgaEnumerate was leaking wrapped tokens.
2) the gcov instrumentation causes test_system::opae_create_ to
return duplicate fd's somehow. This was leaking mock_object *'s.
3) A couple of the properties tests were overwriting tokens_[i],
leaking tokens.

### DIFF
--- a/libopae/api-shell.c
+++ b/libopae/api-shell.c
@@ -201,14 +201,18 @@ fpga_result fpgaGetPropertiesFromHandle(fpga_handle handle,
 					fpga_properties *prop)
 {
 	fpga_result res;
-	fpga_token parent = NULL;
 	opae_wrapped_handle *wrapped_handle =
 		opae_validate_wrapped_handle(handle);
+	struct _fpga_properties *p;
+	int err;
 
 	ASSERT_NOT_NULL(wrapped_handle);
 	ASSERT_NOT_NULL(prop);
 	ASSERT_NOT_NULL_RESULT(
 		wrapped_handle->adapter_table->fpgaGetPropertiesFromHandle,
+		FPGA_NOT_SUPPORTED);
+	ASSERT_NOT_NULL_RESULT(
+		wrapped_handle->adapter_table->fpgaCloneToken,
 		FPGA_NOT_SUPPORTED);
 
 	res = wrapped_handle->adapter_table->fpgaGetPropertiesFromHandle(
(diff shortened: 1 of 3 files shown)


tswhison:
@anandaravuri would you mind having a look at this since @r-rojo is out today? Thanks!

libopae/api-shell.c
```diff
@@ -218,19 +222,34 @@ fpga_result fpgaGetPropertiesFromHandle(fpga_handle handle,
...
+			p->parent, &parent);
 
-		return fpgaPropertiesSetParent(*prop, wrapped_parent);
+		if (res == FPGA_OK) {
+			opae_wrapped_token *wrapped_parent =
+				opae_allocate_wrapped_token(
+					parent, wrapped_handle->adapter_table);
+
+			if (wrapped_parent) {
+				p->parent = wrapped_parent;
```
r-rojo: If we clone p->parent, and then assign p->parent to wrapped_parent, what happens to the original p->parent? Isn't that leaked?

libopae/api-shell.c
```diff
@@ -269,18 +292,33 @@ fpga_result fpgaGetProperties(fpga_token token, fpga_properties *prop)
...
+						parent, wrapped_token->adapter_table);
+
+				if (wrapped_parent) {
+					p->parent = wrapped_parent;
+				} else {
+					OPAE_ERR("malloc failed");
+					res = FPGA_NO_MEMORY;
+				}
+
+			}
```
r-rojo: We should probably refactor this routine into one function and avoid code duplication

libopae/api-shell.c
```diff
@@ -304,19 +346,34 @@ fpga_result fpgaUpdateProperties(fpga_token token, fpga_properties prop)
...
+					parent, wrapped_token->adapter_table);
+
+			if (wrapped_parent) {
+				p->parent = wrapped_parent;
+			} else {
+				OPAE_ERR("malloc failed");
+				res = FPGA_NO_MEMORY;
+			}
+
+		}
```
r-rojo: Same thing here about refactoring into a common function

libopae/api-shell.c
tswhison: I wanted to talk to you about this.  I'm following the precedent that was established in the properties API.  Look at fpgaPropertiesGetParent().  You'll notice that it is cloning the parent token and handing back the clone.  I question whether we need to do that in either place actually.  Why is fpgaPropertiesGetParent() cloning the token?

libopae/api-shell.c
tswhison: Let's wait and see what pans out on the clone discussion.

libopae/api-shell.c
tswhison: Let's wait and see what pans out on the clone discussion.

PR #698, opened 2018-09-20
Add VHDL files to qsys_sim_files.list

deepakcu84:
VHDL files from qsys-generate aren't added to qsys_sim_files.list which breaks HLS simulations.


lzhan55:
Could you add a link for your regression tests with this change?

lzhan55:
Not needed anymore.

PR #700, opened 2018-09-20
tests: Add test coverage for opae-c properties

tswhison:
Increase coverage for fpgaGetProperties[FromHandle].
Add intercept for calloc(), modeled after malloc() intercept.
Call through the adpater to destroy the cloned parent token
if the wrapped token object allocation fails (fpgaGetProperties[FromHandle]
and fpgaUpdateProperties).

### DIFF
--- a/libopae/api-shell.c
+++ b/libopae/api-shell.c
@@ -327,8 +327,9 @@ fpga_result fpgaUpdateProperties(fpga_token token, fpga_properties prop)
 
 	if (FIELD_VALID(p, FPGA_PROPERTY_PARENT) &&
 	    (p->flags & OPAE_PROPERTIES_FLAG_PARENT_ALLOC)) {
-		wrapped_parent = (opae_wrapped_token *) p->parent;
-		p->parent = wrapped_parent->opae_token;
+		wrapped_parent = opae_validate_wrapped_token(p->parent);
+		if (wrapped_parent)
+			p->parent = wrapped_parent->opae_token;
 	}
 
 	res = wrapped_token->adapter_table->fpgaUpdateProperties(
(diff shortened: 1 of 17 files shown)


libopae/api-shell.c
```diff
@@ -1425,7 +1433,7 @@ fpga_result fpgaGetUserClock(fpga_handle handle, uint64_t *high_clk,
 	ASSERT_NOT_NULL(handle);
 	ASSERT_NOT_NULL(low_clk);
 	ASSERT_NOT_NULL(high_clk);
-	ASSERT_NOT_NULL_RESULT(wrapped_handle->adapter_table->fpgaSetUserClock,
+	ASSERT_NOT_NULL_RESULT(wrapped_handle->adapter_table->fpgaGetUserClock,
```
r-rojo: good catch

testing/opae-c/test_buffer_c.cpp
```diff
@@ -0,0 +1,134 @@
...
+ *             and returns FPGA_OK.<br>
+ */
+TEST_P(buffer_c_p, ioaddr) {
+  void *buf_addr = nullptr;
+  uint64_t wsid = 0;
+  uint64_t io = 0xdecafbadbeefdead;
+  ASSERT_EQ(fpgaPrepareBuffer(accel_, (uint64_t) pg_size_,
+                              &buf_addr, &wsid, 0), FPGA_OK);
+  EXPECT_EQ(fpgaGetIOAddress(accel_, wsid, &io), FPGA_OK);
+  EXPECT_NE(io, 0xdecafbadbeefdead);
```
r-rojo: It would be cool if we could verify the io address using /proc/self/pagemap 😄

PR #704, opened 2018-09-22
Added unit test for manage

jelonanX:


testing/xfpga/test_manage_c.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include "test_system.h"
+#include "xfpga.h"
+#include "gtest/gtest.h"
+#include <opae/utils.h>
+#include <opae/fpga.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+fpga_result xfpga_fpgaAssignToInterface(fpga_handle, fpga_token, uint32_t, int);
```
tswhison: Are these necessary when xfpga.h is included?

PR #705, opened 2018-09-22
fix some memory leaks

r-rojo:
Fix memory leak related to mkdtemp
Fix other leaks from not destroying tokens in enum_c tests

### DIFF
--- a/libopae/pluginmgr.c
+++ b/libopae/pluginmgr.c
@@ -64,7 +64,7 @@ static platform_data platform_data_table[] = {
 };
 
 static int initialized;
-static platforms_detected;
+static int platforms_detected;
 
 static opae_api_adapter_table *adapter_list = (void *)0;
 static pthread_mutex_t adapter_list_lock =
--- a/libopae/pluginmgr.c
+++ b/libopae/pluginmgr.c
@@ -182,6 +182,7 @@ int opae_plugin_mgr_finalize_all(void)
 
 	adapter_list = NULL;
 	initialized = 0;
+	platforms_detected = 0;
 
 	opae_mutex_unlock(res, &adapter_list_lock);
 
--- a/libopae/pluginmgr.c
+++ b/libopae/pluginmgr.c
@@ -64,7 +64,6 @@ static platform_data platform_data_table[] = {
 };
 
 static int initialized;
-static int platforms_detected;
 
 static opae_api_adapter_table *adapter_list = (void *)0;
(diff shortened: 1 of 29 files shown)


libopae/pluginmgr.c
```diff
@@ -64,6 +64,7 @@ static platform_data platform_data_table[] = {
 };
 
 static int initialized;
+static platforms_detected;
```
tswhison: You're thinking this would allow us to have another shot at platform detection if we failed the first time?

testing/mock/test_system.cpp
```diff
@@ -206,17 +207,18 @@ test_system *test_system::instance() {
   return test_system::instance_;
 }
 
-std::string test_system::prepare_syfs(const test_platform &platform) {
-  std::string tmpsysfs = "tmpsysfs-XXXXXX";
+void test_system::prepare_syfs(const test_platform &platform) {
+  char tmpsysfs[]{ "tmpsysfs-XXXXXX\0" };
```
tswhison: The explicit \0 is redundant.  C terminate all string constants with \0.

libopae/pluginmgr.c
```diff
@@ -221,6 +222,7 @@ STATIC void opae_plugin_mgr_detect_platform(uint16_t vendor, uint16_t device)
 
 		if (platform_data_table[i].vendor_id == vendor &&
 		    platform_data_table[i].device_id == device) {
+			platforms_detected++;
```
tswhison: The rationale is that we would be able to try platform detection again if we failed to detect anything the first time?

tswhison:
Now that I think of it, why do we need multiple static globals to signify "initialized"?  Why don't we make platforms_detected a local, but let initialized remain the single point of saying we initialized successfully?

libopae/pluginmgr.c
r-rojo: Yes

testing/mock/test_system.cpp
r-rojo: I was just testing you...

r-rojo:
@tswhison , platforms_detected local to what function?

tswhison:
It could be a local in opae_plugin_mgr_initialize.

r-rojo:
@tswhison 
Somewhere here?
https://github.com/OPAE/opae-sdk/blob/c80fe05bf96a9beccebcd134bf28043191ae4954/libopae/pluginmgr.c#L325-L331

PR #715, opened 2018-09-26
Clean up delete errors and add more check for fpgaClearError

jelonanX:


testing/xfpga/test_error_c.cpp
```diff
@@ -407,16 +407,19 @@ TEST_P(error_c_p, error_06) {
   ASSERT_EQ(FPGA_OK, fpgaPropertiesGetNumErrors(filter_, &n));
   printf("Found %d FME error registers\n", n);
 
-  //EXPECT_EQ(FPGA_OK, xfpga_fpgaClearError(t, 0));
-  auto ret = delete_errors("fme","errors");
-  EXPECT_EQ(FPGA_EXCEPTION, xfpga_fpgaClearError(t, 0));
+  struct error_list *p = fake_fme_token_.errors;
+  if (p->info.can_clear){
```
tswhison: Let's loop until we find one that we can clear.  You'll have to keep track of the index, also.

PR #718, opened 2018-09-26
tests: begin adding unit tests for fpgad

tswhison:


testing/fpgad/test_evt_c.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+using namespace opae::testing;
+
+class fpgad_evt_c_p : public ::testing::TestWithParam<std::string> {
+ protected:
+  fpgad_evt_c_p()
+      : tmpfpgad_log_("tmpfpgad-XXXXXX.log") {}
+
+  virtual void SetUp() override {
+    tmpfpgad_log_ = mkstemp(const_cast<char *>(tmpfpgad_log_.c_str()));
```
r-rojo: This may get flagged as a memory leak. That's one of the reasons why I moved away from doing this.
A better approach would be something like:
```C++
  char tmpname[]{ "tmpfpgad-XXXXXX.log" };
  tmpfpgad_log_ = mkstemp(tmpname);
```

testing/fpgad/test_evt_c.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+  evt_notify_ap6_and_null(&err);
+
+  ASSERT_TRUE(do_poll(evt_fds[1]));
+
+  close(evt_fds[1]);
+}
+
+INSTANTIATE_TEST_CASE_P(fpgad_evt_c, fpgad_evt_c_p,
+                        ::testing::ValuesIn(test_platform::keys()));
```
r-rojo: This assumes that fpgad would work on other platforms, like PAC. I'm not sure if it's compatible yet...

testing/fpgad/test_evt_c.cpp
tswhison: I was just looking at mkstemp.  The logic we're using is wrong, because mkstemp returns an open fd to the file, not the substituted name.  Thinking of how to solve this.

testing/fpgad/test_evt_c.cpp
tswhison: How would I make it specific to MCP?  Do a lookup on skx-p?

testing/mock/test_system.cpp
```diff
@@ -433,6 +435,55 @@ int test_system::open(const std::string &path, int flags, mode_t mode) {
   return fd;
 }
 
+static bool _invalidate_read = false;
+static uint32_t _invalidate_read_after = 0;
+static const char * _invalidate_read_when_called_from = nullptr;
```
r-rojo: I'm thinking we should make these variables (and the invalidate_malloc ones) members of test_system (in another PR of course). Thoughts?

testing/mock/test_system.h
```diff
@@ -149,6 +150,8 @@ class test_system {
 
   int open(const std::string &path, int flags);
   int open(const std::string &path, int flags, mode_t m);
+  void invalidate_read(uint32_t after=0, const char *when_called_from=nullptr);
+  ssize_t read(int fd, void *buf, size_t count);
```
r-rojo: I like it!

testing/mock/test_system.cpp
tswhison: Yeah :)

tswhison:
umsg test failure.

PR #719, opened 2018-09-27
feature/dma_api development branch.

abelardojarab:
Please do not merge, just for review purposes.


common/include/opae/dma_types.h
```diff
@@ -0,0 +1,94 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Callback for asynchronous DMA transfers
+typedef void (*fpga_dma_transfer_cb)(void *context);
+
+// DMA transfer type
+typedef enum {
+	HOST_MM_TO_FPGA_ST = 0, // host to AFU sreaming
```
tswhison: Does it make sense to prefix all the enumerators in this file with either OPAE_ or XFPGA_?  I'd like to start cutting down on the affect on the global namespace.  An enumerator like TERMINATE_THREAD is too generic and poses a high risk of name conflict with other source.

PR #720, opened 2018-09-27
Added more tests to buffer

jelonanX:


ibarraza:
Remove test_umsg_c.cpp from this PR. Also, STATIC should already be set. Is your branch up to date with the head of feature/plugin-api?

PR #723, opened 2018-09-28
wsid's bound to shared buffers must be unique

michael-adler:
The hash computation was generating aliases when about 250k pages were allocated.
Use a simple atomic counter, which is also faster, instead.

### DIFF
--- a/libopae/src/common.c
+++ b/libopae/src/common.c
@@ -34,7 +34,6 @@
 #include <stdio.h>
 #include <stdarg.h>
 #include <stdlib.h>
-#include <stdatomic.h>
 
 // Buffer Allocation constants
 #define KB 1024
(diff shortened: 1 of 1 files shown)


michael-adler:
Tested using MPF's test_random torture test and 1 million individual 4KB pages.

michael-adler:
@r-rojo, no problem. I was just glad that the SEGV in my test was caused by a bug here instead of in my code. :-)

r-rojo:
We've been working on redoing our testing infrastructure and now include unit tests for wsid_list. We had to put in a hack in our testing until we could fix wsid_gen. Thanks again! 
https://github.com/OPAE/opae-sdk/blob/2cb49496e9e84c3544974a9beba297d79fad91ab/testing/xfpga/test_wsid_list_c.cpp#L132-L135

PR #724, opened 2018-09-28
Use a hash to track the wsid space.

michael-adler:
Unique workspace IDs are assigned to each allocated shared buffer. The IDs are used
for discovering the PA of a buffer and for passing the buffer size to the kernel when a page
is deallocated.  The data structure was a list, which didn't scale.  A million allocated
pages took over 1.5 hours to free.  The new code accomplishes the same in about 4 seconds.

### DIFF
--- a/libopae/src/buffer.c
+++ b/libopae/src/buffer.c
@@ -255,7 +255,7 @@ fpga_result __FPGA_API__ fpgaPrepareBuffer(fpga_handle handle, uint64_t len,
 	*wsid = wsid_gen();
 
 	/* Add to workspace id in order to store buffer length */
-	if (!wsid_add(&_handle->wsid_root,
+	if (!wsid_add(_handle->wsid_root,
 		      *wsid,
 		      dma_map.user_addr,
 		      dma_map.iova,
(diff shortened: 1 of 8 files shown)


michael-adler:
@OPAE/api - I don't think the remaining Travis errors are mine. Is something wrong with the test system?

michael-adler:
Tested using MPF's test_random torture test by allocating 1 million 4KB pages and then every second deallocating and reallocating a random number of pages. At the end, all pages are deallocated by walking the generated data structures. Failure to find a wsid would trigger an error.

tswhison:
Neat implementation.  Could you please run a memory analysis, something like valgrind memcheck, before we check it in?

michael-adler:
@tswhison good idea. Found one. Now it's clean:

==109681== 
==109681== HEAP SUMMARY:
==109681==     in use at exit: 0 bytes in 0 blocks
==109681==   total heap usage: 1,126 allocs, 1,126 frees, 714,773 bytes allocated
==109681== 
==109681== All heap blocks were freed -- no leaks are possible

PR #725, opened 2018-09-28
add default_ioctl_handler function in test_system

r-rojo:
Using default_ioctl_handler registers a handler as a default.
ioctl handlers are cleared in the finalize routine and the default
handlers are registered in the initialize routine.
This allows for starting with the default handlers earch time a test
fixture calls test_system::initialize()

### DIFF
--- a/testing/mock/test_system.cpp
+++ b/testing/mock/test_system.cpp
@@ -302,6 +302,9 @@ void test_system::initialize() {
   ASSERT_FN(xstat_);
   ASSERT_FN(lstat_);
   ASSERT_FN(scandir_);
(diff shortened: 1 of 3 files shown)


testing/mock/test_system.cpp
```diff
@@ -302,6 +302,9 @@ void test_system::initialize() {
   ASSERT_FN(xstat_);
   ASSERT_FN(lstat_);
   ASSERT_FN(scandir_);
+  for (const auto &kv : default_ioctl_handlers_) {
```
tswhison: Good catch!

PR #727, opened 2018-09-28
Consolidate code checking in single script

nakulkorde:


scripts/test-codingstyle-all.sh
```diff
@@ -0,0 +1,138 @@
+#!/bin/bash
+
+# Function defintion to check cpp coding style
+check_c () {
+    pushd $(dirname $0)
+
+    CHECKPATCH=checkpatch.pl
+    FILES=$(find ../libopae ../common/include/opae ../ase/api/src ../ase/sw ../tools/base/fpgaconf ../tools/base/fpgad \( -iname "*.c" -or -iname "*.h" \) -and \( ! -name "srv.c" \) -and \( ! -path "../libopae/src/usrclk/*" \) -and \( ! -path "../common/include/opae/cxx/*" \) )
```
tswhison: Just a heads-up that libopae/src is going away once we merge feature/plugin-api back to master.  We'll have to scrub things then.

nakulkorde: yes, that's right. I will update it once the plugin api is merged into master.

r-rojo: checkpatch should only run on C files.

nakulkorde: checkpatch is only looking at C files and header files.

r-rojo: sorry, I missed the `!`

PR #730, opened 2018-09-28
Fix a couple memory leaks in C++ FPGA properties wrapper

michael-adler:


michael-adler:
Found this when checking the wsid hash code.

PR #731, opened 2018-09-28
Fix compilation of the tests (gcc-warnings in feature/plugin-api)

abelardojarab:
When using gcc-6 or gcc-7 there are several warnings that are marked as errors and thus compilation cannot succeed.

Main causes:
* unused variables
* gcc also complains about dangling 'if' without 'else' unless you surround 'if' positive branch with { .. }

i think people (me) will need this to build their branches ...

### DIFF
--- a/testing/mock/test_system.cpp
+++ b/testing/mock/test_system.cpp
@@ -248,6 +248,7 @@ test_system *test_system::instance() {
 }
 
 void test_system::prepare_syfs(const test_platform &platform) {
+  int result = 0;
   char tmpsysfs[]{ "tmpsysfs-XXXXXX" };
   if (platform.mock_sysfs != nullptr) {
     char* tmp = mkdtemp(tmpsysfs);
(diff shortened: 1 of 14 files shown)


testing/mock/test_system.cpp
```diff
@@ -257,11 +258,13 @@ void test_system::prepare_syfs(const test_platform &platform) {
     root_ = std::string(tmp);
     std::string cmd = "tar xzf " + std::string(platform.mock_sysfs) + " -C " +
                       root_ + " --strip 1";
-    std::system(cmd.c_str());
+    result = std::system(cmd.c_str());
```
r-rojo: We should probably wrap with something like:
```C++
auto retval = std::system(cmd.c_str());
if (retval != 0) {
  if (retval == -1) {
    throw std::runtime_error("Failed to fork process");
  }
  std::string msg = "Error running command " + cmd;
  msg += ": exit code is " + std::to_string(retval);
  throw std::runtime_error(msg);
}
```

testing/mock/test_system.cpp
```diff
@@ -270,12 +273,12 @@ void test_system::remove_sysfs() {
     }
     if (S_ISDIR(st.st_mode)){
       auto cmd = "rm -rf " + root_;
-      std::system(cmd.c_str());
+      result = std::system(cmd.c_str());
```
r-rojo: We should probably wrap with a check like:
```C++
auto retval = std::system(cmd.c_str());
if (retval != 0) {
  if (retval == -1) {
    std::cerr << "Failed to fork process\n";
  }
  std::string msg = "Error running command " + cmd;
  msg += ": exit code is " + std::to_string(retval);
  std::cerr << msg << "\n";
}
```

Because this is in the cleanup phase, we probably don't care as much and so we shouldn't throw an exception.

PR #738, opened 2018-10-01
Resolved fpgaGetUmsgPtr test and added free_umsg_buffer tests

jelonanX:
Previous fpgaGetUmsgPtr test called an ioctl that doesn't exist, hence the failure. That test has been removed and added new tests for free_umsg_buffer.

### DIFF
--- a/testing/xfpga/test_umsg_c.cpp
+++ b/testing/xfpga/test_umsg_c.cpp
@@ -23,7 +23,12 @@
 // CONTRACT,  STRICT LIABILITY,  OR TORT  (INCLUDING NEGLIGENCE  OR OTHERWISE)
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
+
+extern "C"{
 #include "types_int.h"
+fpga_result free_umsg_buffer(fpga_handle);
+}
+
 #include "xfpga.h"
 #include "intel-fpga.h"
 #include <cstdarg>
@@ -164,13 +169,13 @@ class umsg_c_p
   virtual void TearDown() override {
     EXPECT_EQ(fpgaDestroyProperties(&filter_), FPGA_OK);
 
-    for (auto t : tokens_) {
-      if (t != nullptr) {
+    for (auto &t : tokens_) {
+      if (t) {
         EXPECT_EQ(FPGA_OK, xfpga_fpgaDestroyToken(&t));
       }
     }
 
-    if (handle_ != nullptr) EXPECT_EQ(xfpga_fpgaClose(handle_), FPGA_OK);
+    if (handle_) EXPECT_EQ(xfpga_fpgaClose(handle_), FPGA_OK);
     system_->finalize();
   }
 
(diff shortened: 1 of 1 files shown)


testing/xfpga/test_umsg_c.cpp
```diff
@@ -266,13 +303,76 @@ TEST_P(umsg_c_p, get_umsg_ptr_ioctl_err_02) {
...
+  // register an ioctl handler that will return -1 and set errno to EINVAL
+  system_->register_ioctl_handler(FPGA_PORT_UMSG_DISABLE, dummy_ioctl<-1,EINVAL>);
+  EXPECT_EQ(FPGA_OK, free_umsg_buffer(handle_));
 
+  // register an ioctl handler that will return -1 and set errno to EINVAL
+  res = xfpga_fpgaGetUmsgPtr(handle_, &umsg_ptr);
+  EXPECT_EQ(FPGA_OK, res);
+  system_->register_ioctl_handler(FPGA_PORT_UMSG_DISABLE, dummy_ioctl<0,EINVAL>);
+  system_->register_ioctl_handler(FPGA_PORT_UMSG_SET_BASE_ADDR, dummy_ioctl<-1,EINVAL>);
+  EXPECT_EQ(FPGA_OK, free_umsg_buffer(handle_));
```
ibarraza: Wait, if free_umsg_buffer() fails to perform the exact task that it was defined to do, it still returns FPGA_OK? Ok.

jelonanX: Looking at the source. It executes the ioctl and reports FPGA_ERR upon failure. Inside free_umsg_buffer, it calls free_buffer will return FPGA_OK unless munmap fails.

testing/xfpga/test_umsg_c.cpp
```diff
@@ -266,13 +303,76 @@ TEST_P(umsg_c_p, get_umsg_ptr_ioctl_err_02) {
...
+ *             "Failed to disable UMSG" and returns FPGA_OK
+ *             When ioctl fails on FPGA_PORT_UMSG_SET_BASE_ADDR, FPGA_ERR outputs
+ *             "led to zero UMSG address" and returns FPGA_OK
+ *
+ */
+TEST_P(umsg_c_p, invalid_free_umsg_buffer) {
+  uint64_t* umsg_ptr = NULL;
+  uint64_t *value = 0;
+  system_->register_ioctl_handler(FPGA_PORT_UMSG_SET_BASE_ADDR, umsg_set_base_addr);
+  system_->register_ioctl_handler(FPGA_PORT_UMSG_ENABLE, dummy_ioctl<0,EINVAL>);
```
tswhison: Did you mean to set errno to EINVAL here if returning 0?

testing/xfpga/test_umsg_c.cpp
tswhison: Yeah, looks like none of the errors are being checked, except for handle_check_and_lock.  Should probably do better if we ever enable Umsgs in the hw.

testing/xfpga/test_umsg_c.cpp
jelonanX: yeah,

PR #739, opened 2018-10-01
Removed LibopaecCloseFCommonALL.03 as an invalid test.

ibarraza:


tests/function/gtOpenClose.cpp
```diff
@@ -229,7 +255,7 @@ TEST_F(LibopaecCloseFCommonALL, 03) {
 
     EXPECT_EQ(FPGA_OK, fpgaClose(h));
 
-    EXPECT_EQ((void *)NULL, p->mmio_root);
```
michael-adler: I think the tests are wrong here.

Line 250 should expect mmio_map_is_empty(p->mmio_root) to be TRUE.
Line 254 should expect mmio_map_is_empty(p->mmio_root) to be FALSE.
Line 258 should expect p->mmio_root to be NULL.

At 250 the hash table is allocated but empty.
At 254 there should be an MMIO translation, so no longer empty.
At 258 the handle is closed and the hash table deallocated.

ibarraza: Lines 250 and 254 can be changed to use mmio_map_is_empty, but changing 258 to expect p->mmio_root to be NULL fails the test.

michael-adler: My mistake. You should delete the test at 258. At that point the handle with mmio_root has been freed and referring to it at all is a bug.

michael-adler: Works for me. You can see in https://github.com/OPAE/opae-sdk/blob/master/libopae/src/close.c that mmio_root is contained in _handle, which is freed right before fpgaClose() returns.

PR #743, opened 2018-10-03
Addressing Klocwork issues:

asgibbons:
1. Dereference 'ws' which may be NULL.
2. Free tracker info when open returns.

### DIFF
--- a/libopae/src/mmio.c
+++ b/libopae/src/mmio.c
@@ -201,6 +201,10 @@ static fpga_result find_or_map_wm(fpga_handle handle, uint32_t mmio_num,
 			return result;
 		}
 		wm = wsid_find_by_index(_handle->mmio_root, mmio_num);
+		if(!wm) {
+			FPGA_ERR("unable to map wsid for mmio region %d", mmio_num);
+			return FPGA_NO_MEMORY;
+		}
 	}
 
 	*wm_out = wm;
@@ -228,7 +232,7 @@ fpga_result __FPGA_API__ fpgaWriteMMIO32(fpga_handle handle,
 		return result;
 
 	result = find_or_map_wm(handle, mmio_num, &wm);
-	if (result || (NULL == wm))
+	if (result)
 		goto out_unlock;
 
 	if (offset > wm->len) {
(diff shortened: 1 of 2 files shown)


michael-adler:
Wouldn't it be better to check that wm isn't NULL after line 203 and return an error status in that case? You'd only have to change one line and the error return from find_or_map_wm() would return to being meaningful.

asgibbons:
@michael-adler I'm confused by your comment.  `result` will be set as the return value from the function, which also returns `wm`.  If `result` is non-zero or `wm` is `NULL`, then the function returns `result` - the return value from `find_or_map_wm()`.  Isn't that what is desired here?

michael-adler:
@asgibbons I now understand the confusion. No, find_or_map_wm() should never return NULL when result is non-zero. When wsid_find_by_index() returns NULL, map_mmio_region() is called. This is supposed to then guarantee that wsid_find_by_index() returns a non-NULL result. As a result, you should be able to simply confirm that wsid_find_by_index() returns non-NULL and, if it does, raise some error.

asgibbons:
@michael-adler I finally see your point.  Thanks for the explanation. 👍

asgibbons:
@OPAE/api Can I get an approval please?

asgibbons:
@anandaravuri Can I get an approval please?

libopae/src/mmio.c
```diff
@@ -201,6 +201,10 @@ static fpga_result find_or_map_wm(fpga_handle handle, uint32_t mmio_num,
 			return result;
 		}
 		wm = wsid_find_by_index(_handle->mmio_root, mmio_num);
+		if (!wm) {
+			FPGA_ERR("unable to map wsid for mmio region %d", mmio_num);
+			return FPGA_NO_MEMORY;
```
r-rojo: Good fix. Would it make sense to return FPGA_EXCEPTION or FPGA_NOT_FOUND?

asgibbons: I think NO_MEMORY is better since this is effectively a memory allocation.

PR #745, opened 2018-10-03
ajaraber/compilation fixes

abelardojarab:
final compilation fixes for builtin_return_adress(2)

### DIFF
--- a/testing/mock/test_system.cpp
+++ b/testing/mock/test_system.cpp
@@ -172,7 +172,9 @@ test_device test_device::unknown() {
                      .vendor_id = 0x1234,
                      .device_id = 0x1234,
                      .fme_num_errors = 0x1234,
-                     .port_num_errors = 0x1234};
+                     .port_num_errors = 0x1234,
+                     .gbs_guid = "C544CE5C-F630-44E1-8551-59BD87AF432E",
+		     .mdata = ""};
 }
 
 typedef std::map<std::string, test_platform> platform_db;
--- a/testing/mock/test_system.cpp
+++ b/testing/mock/test_system.cpp
@@ -248,6 +248,7 @@ test_system *test_system::instance() {
 }
 
 void test_system::prepare_syfs(const test_platform &platform) {
+  int result = 0;
   char tmpsysfs[]{ "tmpsysfs-XXXXXX" };
   if (platform.mock_sysfs != nullptr) {
     char* tmp = mkdtemp(tmpsysfs);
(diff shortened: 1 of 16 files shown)


testing/mock/test_system.cpp
```diff
@@ -172,7 +172,9 @@ test_device test_device::unknown() {
                      .vendor_id = 0x1234,
                      .device_id = 0x1234,
                      .fme_num_errors = 0x1234,
-                     .port_num_errors = 0x1234};
+                     .port_num_errors = 0x1234,
+                     .gbs_guid = "C544CE5C-F630-44E1-8551-59BD87AF432E",
+		     .mdata = ""};
```
r-rojo: alignment

testing/mock/test_system.cpp
```diff
@@ -276,11 +280,13 @@ void test_system::prepare_syfs(const test_platform &platform) {
     root_ = std::string(tmp);
     std::string cmd = "tar xzf " + std::string(platform.mock_sysfs) + " -C " +
                       root_ + " --strip 1";
-    std::system(cmd.c_str());
+    result = std::system(cmd.c_str());
   }
+  return (void) result;
```
r-rojo: rather than doing this, can we throw an exception if the result is non-zero?
I put a sample of how to do it in an earlier review.

testing/mock/test_system.cpp
```diff
@@ -290,9 +296,10 @@ void test_system::remove_sysfs() {
     }
     if (S_ISDIR(st.st_mode)) {
       auto cmd = "rm -rf " + root_;
-      std::system(cmd.c_str());
+      result = std::system(cmd.c_str());
     }
   }
+  return (void) result;
```
r-rojo: Can we warn to stderr if result is non-zero?

testing/xfpga/test_error_c.cpp
```diff
@@ -426,24 +428,21 @@ TEST_P(error_c_p, error_06) {
...
+
   std::string errpath = sysfs_fme + "/errors";
   ASSERT_EQ(FPGA_OK, xfpga_fpgaGetProperties(t, &filter_));
 
   // build errors and immediately remove errors dir
   build_error_list(errpath.c_str(), &fake_fme_token_.errors);
-  auto ret = delete_errors("fme","errors");
-  if (ret) { EXPECT_EQ(FPGA_EXCEPTION, xfpga_fpgaClearError(t, 0)); }
+  auto result = delete_errors("fme","errors");
+  (void) result;
```
r-rojo: this isn't needed, right?

PR #749, opened 2018-10-04
Added cxx buffer tests.

ibarraza:


testing/opae-cxx/test_buffer_cxx_core.cpp
```diff
@@ -0,0 +1,222 @@
...
+/**
+ * @test shared_buffer::allocate_no_len
+ * Calling shared_buffer::allocate with buffer length = 0 should throw
+ * invalid_param.
+ */
+TEST_P(buffer_cxx_core, allocate_no_len) {
+  size_t length = 4096;
+  shared_buffer::ptr_t buf;
+
+  EXPECT_THROW(shared_buffer::allocate(handle_, 0), std::exception);
```
r-rojo: You should be more explicit about the exception

testing/opae-cxx/test_buffer_cxx_core.cpp
```diff
@@ -0,0 +1,222 @@
...
+/**
+ * @test shared_buffer::allocate_null
+ * Calling shared_buffer::allocate with a null handle should throw
+ * invalid_param.
+ */
+TEST_P(buffer_cxx_core, allocate_null) {
+  size_t length = 4096;
+  shared_buffer::ptr_t buf;
+
+  EXPECT_THROW(shared_buffer::allocate(NULL, length), std::exception);
```
r-rojo: You should be more explicit about the exception

testing/opae-cxx/test_buffer_cxx_core.cpp
ibarraza: In my tests, no other exception type matched the thrown type. Nothing in except.h would match.

testing/opae-cxx/test_buffer_cxx_core.cpp
ibarraza: In my tests, no other exception type matched the thrown type. Nothing in except.h would match.

PR #753, opened 2018-10-04
Added plugin error cxx tests

jelonanX:


testing/opae-cxx/test_errors_cxx_core.cpp
```diff
@@ -0,0 +1,96 @@
...
+ * Given an OPAE resource token<br>
+ * When I call error::get() with that token<br>
+ * Then I get a non-null
+ * And I am able to read information about the error
+ */
+TEST_P(errors_cxx_core, get_error) {
+  for (auto &t : tokens_) {
+    auto props = properties::get(t);
+    for (int i = 0; i < static_cast<uint32_t>(props->num_errors); ++i) {
+      auto err = error::get(t, i);
```
tswhison: Assert not NULL for err here before using it?

scripts/cover.sh
```diff
@@ -31,7 +31,7 @@ mkdir -p coverage_files
 rm -rf coverage_files/*
 
 echo "Making tests"
-make -j test_unit xfpga
+make -j4 test_unit xfpga
```
tswhison: Why do this?  Did you guys see an issue?

jelonanX: Build was failing on virtual memory exhausted. Nakul suspected this was caused by cmake not rationing the resources with -j and suggested this change.

PR #756, opened 2018-10-05
add fpgaObjectGetSize API function

r-rojo:
* Add fpgaObjectGetSize API function to get the size of the underlying
object. This can be used to prepare buffers for reading from objects.
* Add a test in xfpga and in opae-c for this API
* Add a helper function in test_system to normalize GUID strings
to all lower-case, omitting any new line character at the end, with or
without hyphens.


testing/mock/test_system.cpp
```diff
@@ -442,6 +442,41 @@ FILE *test_system::register_file(const std::string &path) {
   return fptr;
 }
 
+void test_system::normalize_guid(std::string &guid_str, bool with_hyphens) {
+  const size_t std_guid_str_size = 36;
+  const size_t char_guid_str_size = 32;
+  if (guid_str.back() == '\n') {
+    guid_str.erase(guid_str.end()-1);
+  }
```
tswhison: Maybe add some comments to illustrate the different cases you're addressing?

PR #757, opened 2018-10-05
Added support for DCP platform in cxx properties test.

ibarraza:
A cxx properties test was failing an enumeration step for DCP platform due to the input guid always set to MCP. This changes it to retrieve the actual platform device afu_guid field for enumeration.

### DIFF
--- a/testing/opae-cxx/test_properties_cxx_core.cpp
+++ b/testing/opae-cxx/test_properties_cxx_core.cpp
@@ -54,7 +54,7 @@ protected:
 
   void string_to_guid(const char *guid, fpga_guid *result)
   {
-    ASSERT_GE(0, uuid_parse(guid, *result));
+    ASSERT_EQ(0, uuid_parse(guid, *result));
     ASSERT_NE(nullptr, result);
   }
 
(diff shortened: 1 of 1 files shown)


testing/opae-cxx/test_properties_cxx_core.cpp
```diff
@@ -85,8 +88,19 @@ TEST_P(properties_cxx_core, get_no_filter) {
  */
 TEST_P(properties_cxx_core, get_guid_valid) {
   std::vector<token::ptr_t> tokens;
+  const char *guid = nullptr;
+  fpga_guid valid_guid;
+
+  // Retrieve first platform device afu guid.
+  for (auto device : platform_.devices) {
```
tswhison: guid = platform_.devices[0].afu_guid;  :)

testing/opae-cxx/test_properties_cxx_core.cpp
```diff
@@ -52,15 +52,18 @@ class properties_cxx_core : public ::testing::TestWithParam<std::string> {
 
   virtual void TearDown() override { system_->finalize(); }
 
+  void string_to_guid(const char *guid, fpga_guid *result)
+  {
+    ASSERT_EQ(0, uuid_parse(guid, *result));
+    ASSERT_NE(nullptr, result);
```
tswhison: Shouldn't the order of these two lines be switched?

PR #760, opened 2018-10-05
tests: add fpgaconf unit tests

tswhison:
Re-enable problematic fpgad test.

### DIFF
--- a/testing/CMakeLists.txt
+++ b/testing/CMakeLists.txt
@@ -93,6 +93,17 @@ target_include_directories(fpgad-static PUBLIC
   $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>
   $<INSTALL_INTERFACE:include>)
 
+add_library(fpgaconf-static
+    ${OPAE_SDK_SOURCE}/tools/base/fpgaconf/fpgaconf.c
+    )
+
+target_compile_definitions(fpgaconf-static PRIVATE main=fpgaconf_main)
+
+target_include_directories(fpgaconf-static PUBLIC
+  $<BUILD_INTERFACE:${OPAE_INCLUDE_DIR}>
+  $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>
+  $<INSTALL_INTERFACE:include>)
+
 set(CMAKE_C_FLAGS "-DSTATIC='' -DHAVE_CONFIG_H ${CMAKE_C_FLAGS} -pthread")
 
 add_custom_target(test_unit)
@@ -345,6 +356,14 @@ add_mock_test(test_fpgad_fpgad_c fpgad-static
 
 add_mock_test(test_fpgad_ap6_c fpgad-static
     fpgad/test_ap6_c.cpp
+)
+
+############################################################################
+# fpgaconf tests ###########################################################
+############################################################################
+
+add_mock_test(test_fpgaconf_fpgaconf_c fpgaconf-static
+    fpgaconf/test_fpgaconf_c.cpp
 )
 
 add_custom_command(TARGET test_unit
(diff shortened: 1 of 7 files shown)


testing/fpgaconf/test_fpgaconf_c.cpp
```diff
@@ -0,0 +1,853 @@
...
+  strcpy(five, "0x5e");
+  strcpy(six, "-D");
+  strcpy(seven, "0xab");
+  strcpy(eight, "-F");
+  strcpy(nine, "3");
+  strcpy(ten, "-S");
+  strcpy(eleven, "2");
+  strcpy(twelve, "-A");
+  strcpy(thirteen, "-I");
+  strcpy(fourteen, "file.gbs");
```
r-rojo: Would this be easier to read if it were more like:
```C++
std::array<char*, 15> argv = { "fpgaconf", "-v", "-n", "--force",
                                "-B", "0x5e" /*, ...*/ };
```

tools/base/fpgaconf/fpgaconf.c
```diff
@@ -95,8 +96,8 @@ struct bitstream_info {
 	fpga_guid interface_id;
 };
 
-static fpga_result get_bitstream_ifc_id(const uint8_t *bitstream,
-					fpga_guid *guid)
+fpga_result get_bitstream_ifc_id(const uint8_t *bitstream,
```
r-rojo: What about STATIC?

testing/CMakeLists.txt
```diff
@@ -93,6 +93,17 @@ target_include_directories(fpgad-static PUBLIC
   $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>
   $<INSTALL_INTERFACE:include>)
 
+add_library(fpgaconf-static
+    ${OPAE_SDK_SOURCE}/tools/base/fpgaconf/fpgaconf.c
+    )
+
+target_compile_definitions(fpgaconf-static PRIVATE main=fpgaconf_main)
```
r-rojo: oh, nice!

testing/fpgaconf/test_fpgaconf_c.cpp
tswhison: You would get warnings/errors about mixing const char * and char *.  That's why I coded it this way.  Note that argv is char *argv[].

tools/base/fpgaconf/fpgaconf.c
tswhison: It's not necessary.  Each of the other helper fn's in this file could have theoretically been declared static..  Since it's an app, I don't think it makes much difference.

testing/fpgaconf/test_fpgaconf_c.cpp
r-rojo: And so is the "std::array<char*, 15>" 😈 
You you just have to change the function call to use `argv.data()`

tools/base/fpgaconf/fpgaconf.c
r-rojo: 👍

testing/fpgaconf/test_fpgaconf_c.cpp
tswhison: But in your example, the string constants like "fpgaconf" are const char *, but the definition of your array is char *.  That's essentially the issue. :)

r-rojo: Ok, you're right... boo! :)

tswhison: No harm done in questioning..  Glad to have convinced you. :)

PR #766, opened 2018-10-09
Add support for FPGA_BUF_PREALLOCATED to fpgaPrepareBuffer() in ASE

michael-adler:
Allowing the mapping of pre-allocated memory forces a rearchitecting of the way the RTL simulator and the application share memory. It becomes impossible to share a buffer between the two processes with mmap. Instead, we now forward all memory read and write requests from the simulator to the application.

- Update ASE's copy of OPAE API modules to recent versions (e.g. buffer.c).
- Add a structure similar to a page table to track the simulation of pinned memory.
- Improve detection of pinned and mapped pages. More refined error detection now reports accesses to addresses where the virtually mapped pages have been unmapped improperly.
- Improved pinned buffer management allows many more pages to be mapped without performance loss.
- Various logging bug fixes, e.g. logging of pinned buffer addresses and 64 bit MMIO write data.


michael-adler:
DCP tests pass:
    http://sw-pert.altera.com/pert/pert.php?test_run_id=2724110
    http://sw-pert.altera.com/pert/pert.php?test_run_id=2724124

@lzhan55 can you please run this branch on your ASE test suite?

lzhan55:
Will do.

-Tracy

lzhan55:
I will check with tools/test maintainers to see what happened here.

lzhan55:
[100%] Building CXX object tests/CMakeFiles/gtase.dir/function/gtGetProperties.cpp.o
In file included from /home/lab/lzhan55/opae-sdk/mybuild/tests/gtest/src/gtest/googletest/include/gtest/gtest.h:58:0,
                 from /home/lab/lzhan55/opae-sdk/tests/function/gtMMIO.cpp:32:
/home/lab/lzhan55/opae-sdk/tests/function/gtMMIO.cpp: In member function ‘virtual void LibopaecMmioCommonALL_mmio_drv_positive_init_01_Test::TestBody()’:
/home/lab/lzhan55/opae-sdk/tests/function/gtMMIO.cpp:93:69: error: cannot convert ‘wsid_tracker*’ to ‘wsid_map*’ for argument ‘1’ to ‘bool mmio_map_is_empty(wsid_map*)’
   EXPECT_TRUE(mmio_map_is_empty(((struct _fpga_handle*)h)->mmio_root));
                                                                     ^
/home/lab/lzhan55/opae-sdk/mybuild/tests/gtest/src/gtest/googletest/include/gtest/internal/gtest-internal.h:1189:34: note: in definition of macro ‘GTEST_TEST_BOOLEAN_’
       ::testing::AssertionResult(expression)) \
                                  ^
/home/lab/lzhan55/opae-sdk/tests/function/gtMMIO.cpp:93:3: note: in expansion of macro ‘EXPECT_TRUE’
   EXPECT_TRUE(mmio_map_is_empty(((struct _fpga_handle*)h)->mmio_root));
   ^
/home/lab/lzhan55/opae-sdk/tests/function/gtMMIO.cpp: In member function ‘virtual void LibopaecMmioCommonALL_mmio_drv_negative_map_mmio_02_Test::TestBody()’:
/home/lab/lzhan55/opae-sdk/tests/function/gtMMIO.cpp:156:69: error: cannot convert ‘wsid_tracker*’ to ‘wsid_map*’ for argument ‘1’ to ‘bool mmio_map_is_empty(wsid_map*)’
   EXPECT_TRUE(mmio_map_is_empty(((struct _fpga_handle*)h)->mmio_root));
                                                                     ^
/home/lab/lzhan55/opae-sdk/mybuild/tests/gtest/src/gtest/googletest/include/gtest/internal/gtest-internal.h:1189:34: note: in definition of macro ‘GTEST_TEST_BOOLEAN_’
       ::testing::AssertionResult(expression)) \
                                  ^
/home/lab/lzhan55/opae-sdk/tests/function/gtMMIO.cpp:156:3: note: in expansion of macro ‘EXPECT_TRUE’
   EXPECT_TRUE(mmio_map_is_empty(((struct _fpga_handle*)h)->mmio_root));
   ^
make[2]: *** [tests/CMakeFiles/gtase.dir/function/gtMMIO.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [tests/CMakeFiles/gtase.dir/all] Error 2
make: *** [all] Error 2

michael-adler:
I can fix that one.

ase/api/src/buffer.c
```diff
@@ -26,122 +26,402 @@
...
+#else
+#define ADDR (void *)(0x0UL)
+#define FLAGS_4K (MAP_PRIVATE | MAP_ANONYMOUS)
+#define FLAGS_2M (FLAGS_4K | MAP_HUGETLB)
+#define FLAGS_1G (FLAGS_2M | MAP_1G_HUGEPAGE)
+#endif
+
+
+/*
+ * In simulation the physical page assignment is synthetic. We can afford to
```
deepakcu84: comments are very helpful

michael-adler: I agree comments are helpful, but buffer.c here matches the OPAE buffer.c as closely as possible. My goal when adding files in the api directory was to be as close as possible to OPAE. Some day I'd like to share code between OPAE and ASE and have simulator/hardware functions separated out.

I think you'll find the ASE-specific changes have better comments.

lzhan55:
I will need one more day to review the changes. Thank you!

lzhan55:
Since DCP tests and OPAE tests passed, it is OK to check in.

ase/sw/ase_host_memory.c
```diff
@@ -0,0 +1,430 @@
...
+// ========================================================================
+//
+//	Maintain a simulated page table in order to track pinned pages.
+//
+// ========================================================================
+
+/*
+ * The page table here has a single job: indicate whether a given physical
+ * address (iova) is pinned. Since the simulator uses a simple XOR function
+ * for mapping virtual to physical addresses, the table is not needed for
```
deepakcu84: I like this simple hash function. But, consider the case where memory isn't backed by real hugepages (i.e. prepareBuffer uses sim_huge_mmap). What if the application spawns more than one process, each invoking fpgaPrepareBuffer separately? You could end up in a situation where identical VAs from both processes are mapped to the same PA. If the first process unpins the buffer, the buffer will be automatically unpinned for the second process right?

ase/api/src/buffer.c
```diff
@@ -26,122 +26,402 @@
...
+ * be more lenient about the availability of huge pages on the host that
+ * is running a simulation. This function first tries to use real huge pages.
+ * If unavailable, it instead allocates a virtual region that is properly
```
deepakcu84: why even rely on real huge pages on the host? I don't see any advantage here.

ase/sw/ase_host_memory.c
```diff
@@ -0,0 +1,430 @@
...
+
+/*
+ * Index of a 512 entry set of page pointers in the table, given a level.
+ */
+static inline int ase_pt_idx(uint64_t iova, int pt_level)
+{
+	// The low 12 bits are the 4KB page offset. Groups of 9 bits above that
+	// correspond to increasing levels in the page table hierarchy.
+	assert(pt_level <= 3);
+	return 0x1ff & (iova >> (12 + pt_level * 9));
```
deepakcu84: Neat. but, requires some effort to understand. consider replacing 12 and 9 with LOG2_PAGE_SIZE, LOG2_GROUP_SIZE or some other meaningful macro

ase/sw/protocol_backend.c
```diff
@@ -219,35 +224,108 @@ void sv2c_script_dex(const char *str)
...
-			ase_fakeaddr_to_vaddr((uint64_t) phys_addr);
 
 		// Write to memory
-		ase_memcpy(wr_target_vaddr, (char *) pkt->qword,
-			   CL_BYTE_WIDTH);
+		wr_req.req = HOST_MEM_REQ_WRITE_LINE;
+		wr_req.addr = phys_addr;
+		ase_memcpy(wr_req.data, (char *) pkt->qword, CL_BYTE_WIDTH);
+
+		mqueue_send(sim2app_membus_wr_req_tx, (char *) &wr_req, sizeof(wr_req));
```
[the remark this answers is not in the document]
michael-adler: mqueue_send() currently has no return value (void). This is pervasive and should be on the list of things to fix, but is bigger than just this change since it involves figuring out how to tear down the simulation cleanly.

ase/sw/protocol_backend.c
```diff
@@ -262,59 +340,109 @@ void wr_memline_dex(cci_pkt *pkt)
...
-	// Get cl_addr, deduce rd_target_vaddr
 	phys_addr = (uint64_t) pkt->cl_addr << 6;
-	rd_target_vaddr = ase_fakeaddr_to_vaddr((uint64_t) phys_addr);
 
-	// Read from memory
-	ase_memcpy((char *) pkt->qword, rd_target_vaddr, CL_BYTE_WIDTH);
+	rd_req.req = HOST_MEM_REQ_READ_LINE;
+	rd_req.addr = phys_addr;
+
+	mqueue_send(sim2app_membus_rd_req_tx, (char *) &rd_req, sizeof(rd_req));
```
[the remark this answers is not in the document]
michael-adler: Same void problem.

ase/sw/protocol_backend.c
```diff
@@ -262,59 +340,109 @@ void wr_memline_dex(cci_pkt *pkt)
...
+	while (true) {
+		status = mqueue_recv(app2sim_membus_rd_rsp_rx, (char *) &rd_rsp, sizeof(rd_rsp));
+
+		if (status == ASE_MSG_PRESENT) {
+			if (rd_rsp.status != HOST_MEM_STATUS_VALID) {
+				memline_addr_error("READ", rd_rsp.status, rd_rsp.pa, rd_rsp.va);
+			}
+
+			// Read from memory
+			ase_memcpy((char *) pkt->qword, rd_rsp.data, CL_BYTE_WIDTH);
```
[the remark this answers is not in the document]
michael-adler: Yes, and this goes all the way down to the Verilog part of the simulator. I don't think it has much impact on simulation speed since each line is necessarily delivered in separate simulated cycles. The version I'm replacing was logically similar -- one call to the DPI read function per line, even in multi-line requests.

ase/sw/protocol_backend.c
```diff
@@ -1236,21 +1391,9 @@ void start_simkill_countdown(void)
 	final_ipc_cleanup();
 
 	// wait for server shutdown
+	pthread_cancel(socket_srv_tid);
```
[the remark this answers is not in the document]
michael-adler: Good catch. Yes, it was causing a hang on exit.

ase/sw/ase_host_memory.h
```diff
@@ -0,0 +1,129 @@
...
+// is already mapped and in use by the application.  It is nearly impossible
+// in Linux to reclassify an existing memory page as shared.  Handling all
+// accesses on the application side makes shared mapping unnecessary.
+//
+
+//
+// Requests types that may be sent from simulator to application.
+//
+typedef enum {
+	HOST_MEM_REQ_READ_LINE,
```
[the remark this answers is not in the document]
michael-adler: Yes.

ase/sw/ase_host_memory.h
```diff
@@ -0,0 +1,129 @@
...
+//
+// Read response, application to simulator.
+//
+typedef struct {
+	uint8_t data[CL_BYTE_WIDTH];
+
+	// Simulated host physical address
+	uint64_t pa;
+	// Virtual address in application space
+	void *va;
```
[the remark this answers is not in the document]
michael-adler: The simulator doesn't. The human running it does. Illegal (unpinned) address access messages are printed by the RTL simulator, triggered by this data structure. See memline_addr_error() in protocol_backend.c.

ase/api/src/buffer.c
michael-adler: The flip-side is why not try to behave as much like a hardware run as possible and let users with properly configured machines see that their huge page requests work by checking /proc/meminfo, etc. even in simulation? I thought it might be even more confusing when a knowledgeable user doesn't get a huge page in response to a properly formatted request.

Do you object to keeping the code as it is?

ase/sw/ase_host_memory.c
michael-adler: Good observation. I thought about this too. The way ASE is currently structured I don't believe it could possibly support a pair of processes both talking to the RTL simulator. There are lots of statics and there is only one instance of channels to an application from the simulator. Updating the page mapping by adding an address allocator would be comparatively minor.

I think the address mapping is pretty well contained. We could switch to a range allocator and the largest change would be updating the ase_pt code in ase_host_memory to use the existing table to map pa to va. The memory footprint would grow, since the current table uses bit vectors (address present/not present) and the new table would need a full 64 bit va instead of a flag.

So, do we build it now or later if we go to multi-process? If now, we could commit the current code and I could add an allocator as a separate, smaller commit. What do you think?

ase/sw/protocol_backend.c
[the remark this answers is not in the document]
deepakcu84: Ok. That clarifies.

ase/sw/ase_host_memory.c
deepakcu84: Multiprocess capability can be added later. Just wanted to highlight it here. I don't think it is a serious issue. I guess you could also create unique PAs hashing virtual address and process ID.

ase/sw/protocol_backend.c
```diff
@@ -262,59 +340,101 @@ void wr_memline_dex(cci_pkt *pkt)
...
+			ase_memcpy((char *) pkt->qword, rd_rsp.data, CL_BYTE_WIDTH);
+			break;
+		}
+
+		// Error?  Probably channel closed and the simulator will be closing
+		// soon.
+		if (status == ASE_MSG_ERROR) {
+			pkt->success = 0;
+			break;
+		}
```
r-rojo: This loop is essentially polling on mqueue_recv, yes? Do we need a sleep in this loop?

michael-adler: It's performance critical, so a pure spin loop seemed appropriate. In normal situations it doesn't have to wait since it doesn't check for a response until the simulated cycle after the request.

It's a simulator, too, so my threshold for caring about power is lower. :-)

r-rojo: fair enough 😃

PR #771, opened 2018-10-09
Added missing properties functional tests from old mock

jelonanX:


testing/xfpga/test_properties_c.cpp
```diff
@@ -74,32 +73,98 @@ class properties_p1 : public ::testing::TestWithParam<std::string> {
...
+ *             the function returns FPGA_OK, and
+ *             the returned properties are valid for that handle.
+ */
+TEST_P(properties_c_p, valid_gets) {
+  fpga_objtype objtype;
+
+  ASSERT_EQ(FPGA_OK, xfpga_fpgaGetPropertiesFromHandle(accel_, &props_));
+  ASSERT_EQ(FPGA_OK, fpgaPropertiesGetObjectType(props_, &objtype));
+  EXPECT_EQ(FPGA_ACCELERATOR, objtype);
+}
```
ibarraza: This test is missing the case for FME. And probably should explicitly set the object type to something that is not expected, prior to getting the object type in properties.

testing/xfpga/test_properties_c.cpp
```diff
@@ -74,32 +73,98 @@ class properties_p1 : public ::testing::TestWithParam<std::string> {
...
+
+/**
+ * @test       valid_get
+ *
+ * @brief      When fpgaGetPropertiesFromHandle is called
+ *             with a valid handle,
+ *             the function returns FPGA_OK, and
+ *             the returned properties are valid for that handle.
+ */
+TEST_P(properties_c_p, valid_gets) {
```
ibarraza: Test case name different from test header.

ibarraza:
In a lot of these tests, props_ gets overwritten prior to destroying it (I don't think it even needs to be populated in SetUp for these tests), which would leak things I think.

testing/xfpga/test_properties_c.cpp
```diff
@@ -45,61 +46,193 @@ class properties_p1 : public ::testing::TestWithParam<std::string> {
...
-  EXPECT_EQ(xfpga_fpgaGetPropertiesFromHandle(accel_, &props_), FPGA_OK);
+/**
+ * @test    from_handle
+ * @brief   Tests: xfpga_fpgaGetPropertiesFromHandle
+ * @details Given a null fpga_properties* object<br>
+ *          When I call xfpga_fpgaGetPropertiesFromHandle with an valid handle,
+ *          expected result is FPGA_OK.<br>
+ */
+TEST_P(properties_c_p, from_handle) {
+  EXPECT_EQ(xfpga_fpgaGetPropertiesFromHandle(accel_, &prop_), FPGA_OK);
```
tswhison: prop_ is leaked in all cases where it is passed as a parameter to fpgaGetProperties[FromHandle].

testing/xfpga/test_properties_c.cpp
```diff
@@ -45,61 +46,189 @@ class properties_p1 : public ::testing::TestWithParam<std::string> {
...
     num_matches_ = 0xc01a;
     invalid_device_ = test_device::unknown();
   }
 
   virtual void TearDown() override {
-    EXPECT_EQ(fpgaDestroyProperties(&filter_), FPGA_OK);
-    EXPECT_EQ(fpgaDestroyProperties(&props_), FPGA_OK);
-    EXPECT_EQ(xfpga_fpgaClose(accel_), FPGA_OK);
-    for (auto &t : tokens_) {
+    if (prop_) { EXPECT_EQ(fpgaDestroyProperties(&prop_), FPGA_OK); };
```
tswhison: Probably need to initialize prop_ to nullptr in the SetUp routine.

PR #772, opened 2018-10-09
tests: add coreidle unit tests

tswhison:


tools/extra/coreidle/main.c
```diff
@@ -240,8 +239,9 @@ int main( int argc, char** argv )
 	ON_ERR_GOTO(result, out_exit, "destroying properties object");
 
 out_exit:
+	if (coreidleCmdLine.gbs_data)
+		free(coreidleCmdLine.gbs_data);
```
ibarraza: coreidleCmdLine.gbs_data does not get freed in the case of fpga resource not found, such as in line 185.

tswhison: good catch!

testing/mock/test_system.cpp
```diff
@@ -591,6 +602,47 @@ FILE *test_system::fopen(const std::string &path, const std::string &mode) {
   return fopen_(syspath.c_str(), mode.c_str());
 }
 
+FILE *test_system::popen(const std::string &cmd, const std::string &type) {
+  // Is this something we're interested in?
+  if (0 == cmd.compare(0, 5, "rdmsr")) {
```
r-rojo: Aww, no regex?

testing/mock/test_system.cpp
```diff
@@ -591,6 +602,47 @@ FILE *test_system::fopen(const std::string &path, const std::string &mode) {
...
+
+    size_t last_spc = cmd.find_last_of(' ');
+    std::string msr(cmd.substr(last_spc+1));
+
+    if (0 == msr.compare("0x35")) {
+      fprintf(fp, "0x0000000000180030");
+    } else if (0 == msr.compare("0x610")) {
+      fprintf(fp, "0x000388d000148758");
+    } else if (0 == msr.compare("0x606")) {
+      fprintf(fp, "0x00000000000a0e03");
```
r-rojo: we could use regex and a map to do this... 😄

PR #774, opened 2018-10-09
Added missing errors functional tests from old mock

jelonanX:


testing/xfpga/test_error_c.cpp
```diff
@@ -643,4 +641,30 @@ TEST(error_c, error_06) {
...
 
-INSTANTIATE_TEST_CASE_P(error_c, error_c_p, ::testing::ValuesIn(test_platform::keys(true)));
+
+/**
+ * @test       error_07
+ *
+ * @brief      When passed pathname longer than FILENAME_MAX
+ *             build_error_list() should return and not build anything
+ *
+ *@note        Must set env-variable LIBOPAE_LOG=1 to run this test.
```
tswhison: Should we update scripts/cover.sh to do this?  @r-rojo ?

r-rojo: We can also set the environment variable in the cmake test target. I think that would be a little closer to the test so I would do that.

testing/CMakeLists.txt
```diff
@@ -258,6 +259,9 @@ add_mock_test(test_xfpga_error_c xfpga-static
     xfpga/test_error_c.cpp
 )
 
+set_tests_properties(test_xfpga_error_c
```
tswhison: nice!

PR #783, opened 2018-10-11
coreidle: add PCIe segment parameter

tswhison:


testing/coreidle/test_main_c.cpp
```diff
@@ -33,6 +33,7 @@ extern "C" {
 
 struct  CoreIdleCommandLine
 {
+	int      segment;
```
ibarraza: Mixing tabs and spaces?! The horror.

tswhison: My bad. :)

PR #785, opened 2018-10-11
Improved support for platform interface ports that are vectors.

michael-adler:
- New preprocessor macro indicating whether a port is a vector.
- Improved error checking for platform/AFU vector mismatch.
- Improved reporting when encountering JSON formatting errors.

### DIFF
--- a/platforms/platform_if/rtl/platform_if_macros.vh
+++ b/platforms/platform_if/rtl/platform_if_macros.vh
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2017, Intel Corporation
+// Copyright (c) 2018, Intel Corporation
 // All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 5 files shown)


michael-adler:
Tests pass:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2731979
http://sw-pert.altera.com/pert/pert.php?test_run_id=2732044

platforms/platform_if/rtl/platform_if_macros.vh
```diff
@@ -0,0 +1,57 @@
...
+// between parameters. These macros maintain state that properly inserts
+// commas between parameters that may or may not be set, based on
+// preprocessor macros.
+//
+// ========================================================================
+
+// Begin a list of parameters with `PLATFORM_ARG_LIST_BEGIN
+`define PLATFORM_ARG_LIST_BEGIN \
+    `undef PLATFORM_ARG_LIST_SEPARATOR \
+    `define PLATFORM_ARG_LIST_SEPARATOR
```
lzhan55: Why did you do undef and define  one after another? Does it make any difference?

michael-adler: Because the pattern may be used multiple times and Verilog emits at least a warning when redefining a macro. C does the same thing.

PR #791, opened 2018-10-11
Fix fpgad initialization in cxx events 

jelonanX:


testing/opae-cxx/test_events_cxx_core.cpp
```diff
@@ -48,13 +48,15 @@ using namespace opae::fpga::types;
...
           handle_(nullptr) {}
 
   virtual void SetUp() override {
-    tmpfpgad_log_ = mkstemp(const_cast<char *>(tmpfpgad_log_.c_str()));
-    tmpfpgad_pid_ = mkstemp(const_cast<char *>(tmpfpgad_pid_.c_str()));
+    strcpy(tmpfpgad_log_, "tmpfpgad-XXXXXX.log");
+    strcpy(tmpfpgad_pid_, "tmpfpgad-XXXXXX.pid");
+
+    tmpfpgad_log_fd_ = mkstemps(tmpfpgad_log_, 4);
+    tmpfpgad_pid_fd_ = mkstemps(tmpfpgad_pid_, 4);
```
tswhison: I think we can just safely use close here and remove the fds..
ie
  close(mkstemps(tmpfpgad_log_, 4));
  close(mkstemps(tmpfpgad_pid_, 4));

testing/opae-cxx/test_events_cxx_core.cpp
```diff
@@ -102,6 +105,7 @@ class events_cxx_core : public ::testing::TestWithParam<std::string> {
     }
   }
 
+
   handle::ptr_t handle_;
```
tswhison: Need to move the declaration of tokens_ to the top to match the initialization order in the constructor.

PR #794, opened 2018-10-12
Fixed userclk tool not destroying token.

ibarraza:


tools/extra/userclk/main.c
```diff
@@ -217,6 +217,10 @@ int main(int argc, char *argv[])
 	result = fpgaClose(accelerator_handle);
 	ON_ERR_GOTO(result, out_destroy_prop, "closing accelerator");
 
+out_destroy_tok:
+	result = fpgaDestroyToken(&accel_token);
+	ON_ERR_GOTO(result, out_destroy_tok, "destroying token object");
```
r-rojo: do we want to go to the same label here? or should we just drop down to the next?

PR #799, opened 2018-10-15
tests: add hello_events unit tests

tswhison:


ibarraza:
hello_events_c_fpgad_p.main2 seems to be failing even though it shows in ctest as passing? It fails with "Expected: (hello_events_main(3, argv)) != (0), actual: 0 vs 0".

tswhison:
Weird.. I just ran it again, and it shows as passing.

ibarraza:
You mean in ctest or in the log? I see ctest reporting it as passing but looking at the log shows it as failing.
test

samples/hello_events.c
```diff
@@ -215,147 +200,138 @@ int find_fpga(fpga_token *fpga, uint32_t *num_matches)
...
+		ON_ERR_GOTO(res1, out_destroy_tok, "setting inject error register");
 
-		res = inject_ras_fatal_error(fpga_device_token, 0);
-		ON_ERR_GOTO(res, out_destroy_tok, "unsetting inject error register");
+		usleep(2000000);
+		res1 = inject_ras_fatal_error(fpga_device_token, 0);
+		ON_ERR_GOTO(res1, out_destroy_tok, "unsetting inject error register");
 
-		goto out_destroy_tok;
+		exit(0);
```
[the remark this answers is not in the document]
tswhison: Oh snap.  No, you're right.  Will fix.

samples/hello_events.c
```diff
@@ -89,59 +86,51 @@ struct ras_inject_error {
...
 
 out_destroy_obj:
-	result = fpgaDestroyObject(&inj_err_object);
-	if (result != FPGA_OK) {
-		OPAE_ERR("Failed to Destroy Object");
-		goto out_close;
-	}
-
+	res2 = fpgaDestroyObject(&inj_err_object);
+	ON_ERR_GOTO(res2, out_close, "Failed to Destroy Object");
```
r-rojo: Shouldn't we just print a message here? We're `goto`'ing to a label that we would fall into next anyway, no?

samples/hello_events.c
```diff
@@ -215,147 +200,140 @@ int find_fpga(fpga_token *fpga, uint32_t *num_matches)
...
 	if (num_matches < 1) {
-		fprintf(stderr, "No matches for bus number provided \n");
-		res = FPGA_NOT_FOUND;
+		fprintf(stderr, "No matches for bus number provided.\n");
+		res1 = FPGA_NOT_FOUND;
 		goto out_exit;
 	}
 
+	res1 = get_bus(fpga_device_token, &bus);
+	ON_ERR_GOTO(res1, out_exit, "getting bus num");
```
r-rojo: We don't destroy token if we `goto out_exit`

samples/hello_events.c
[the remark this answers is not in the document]
r-rojo: Good catch @ibarraza

samples/hello_events.c
tswhison: This does print an error, but there is no harm done is there?  I mean, it's logically correct.

samples/hello_events.c
tswhison: Right, the jump should be to out_destroy_tok.

testing/hello_events/test_hello_events_c.cpp
```diff
@@ -0,0 +1,369 @@
...
+  sprintf(two, "%d", platform_.devices[0].bus);
+
+  char *argv[] = { zero, one, two };
+
+  /* FIXME: hello_events_main will time out here for
+  ** a mock environment, because there is no hardware
+  ** behind the error injection register.
+  EXPECT_EQ(hello_events_main(3, argv), 0);
+  */
+  EXPECT_NE(hello_events_main(3, argv), 0);
```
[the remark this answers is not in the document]
tswhison: 👍

PR #801, opened 2018-10-16
Add a strdup replacement

r-rojo:
Add a replacement to strdup which uses safestring strcpy version


libopae/plugins/xfpga/sysfs.c
```diff
@@ -777,13 +777,24 @@ fpga_result cat_handle_sysfs_path(char *dest, fpga_handle handle,
 	return cat_token_sysfs_path(dest, _handle->token, path);
 }
 
+STATIC char* cstr_dup(const char *str) {
+	size_t s = strlen(str);
```
tswhison: Might get dinging for not using strnlen_s. :(

r-rojo: I think strlen is OK. We use it all over our codebase 😄

PR #805, opened 2018-10-17
fix compilation of ASE gtase because missing sysfs

abelardojarab:
fix compilation of ASE gtase because missing sysfs


nakulkorde:
I second @tswhison . If we merge this change into feature/plugin-api, we would be able to test it out before merging feature/plugin-api into master.

cmake/modules/test_config_ase.cmake
```diff
@@ -32,9 +32,9 @@ set(COMMON_SRC gtmain.cpp jsonParser.cpp
   unit/gtEnumerate.cpp
   unit/gtOptionParser.cpp
   unit/gtAnyValue.cpp
-  unit/gtsysfs.cpp
+  # unit/gtsysfs.cpp
```
r-rojo: remove instead of commenting out

PR #806, opened 2018-10-18
Fix some AFU source file list parsing:

michael-adler:
- VHDL simulation doesn't accept +define in the list of sources.
- QI:foo.tcl was treated as a filename when setting up Qsys simulation.

### DIFF
--- a/ase/scripts/generate_ase_environment.py
+++ b/ase/scripts/generate_ase_environment.py
@@ -222,12 +222,9 @@ def config_sources(fd, filelist):
         if (len(s) == 0):
             None
         elif (s[0] == '+'):
-            # + simulator commands go in both Verilog and VHDL
+            # + simulator commands go only in Verilog.  VHDL simulation
+            # doesn't support these.
             vlog_srcs.append(s)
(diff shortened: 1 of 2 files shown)


michael-adler:
Tests pass, module the usual license problems:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2749774
http://sw-pert.altera.com/pert/pert.php?test_run_id=2749841

ase/scripts/generate_ase_environment.py
```diff
@@ -222,12 +222,9 @@ def config_sources(fd, filelist):
         if (len(s) == 0):
             None
         elif (s[0] == '+'):
-            # + simulator commands go in both Verilog and VHDL
+            # + simulator commands go only in Verilog.  VHDL simulation
+            # doesn't support these.
             vlog_srcs.append(s)
```
lzhan55: Do we still support vhdl_srcs files?

michael-adler: Yes, just not + directives since they don't work/aren't supported for VHDL.

PR #809, opened 2018-10-18
Fixes to support different FME/Port numbers

tswhison:
In certain device topologies, it has become valid to have
an FME/Port device number that differs from the
containing device number, eg

/sys/class/fpga/intel-fpga-dev.1/intel-fpga-fme.0
/sys/class/fpga/intel-fpga-dev.1/intel-fpga-port.0

This fix adds a subdev_instance member to struct _fpga_token and
tracks both the device_instance (1 in the example) and the
subdev_instance (0 in the example) in the struct. 5 of the sysfs
accessor functions are also updated to accomodate both device and
subdev numbers.


libopae/plugins/xfpga/sysfs.c
```diff
@@ -415,7 +414,7 @@ fpga_result sysfs_get_socket_id(int dev, uint8_t *socket_id)
 	snprintf_s_ii(spath, SYSFS_PATH_MAX,
 		      SYSFS_FPGA_CLASS_PATH SYSFS_FME_PATH_FMT
 		      "/" FPGA_SYSFS_SOCKET_ID,
-		      dev, dev);
+		      dev, subdev);
```
r-rojo: Interesting, lines 413 to 417 seem to be duplicated a lot... 🤔

testing/xfpga/test_token_list_c.cpp
```diff
@@ -42,7 +42,27 @@ extern pthread_mutex_t global_lock;
 
 using namespace opae::testing;
 
-TEST(token_list_c, simple_case)
+class token_list_c_p : public ::testing::TestWithParam<std::string> {
```
r-rojo: Why are we parameterizing these tests? I thought they faked the sysfs structures (rather than relying on mock/real driver). Do they use driver related structures?

libopae/plugins/xfpga/token_list.c
```diff
@@ -160,23 +178,69 @@ struct _fpga_token *token_get_parent(struct _fpga_token *_t)
...
+	int i;
+	int found;
 
 	p = strstr(_t->sysfspath, FPGA_SYSFS_AFU);
 	if (!p) // FME objects have no parent.
 		return NULL;
 
-	p = strrchr(_t->sysfspath, '.');
-	if (!p)
+	// Find the parent FME device of the specified Port device.
```
r-rojo: Should we create an internal object that refers to the parent and then stash that in the fme/port object? Or at the very least, stash the path to the top device?

tswhison:
@r-rojo Yes, actually.  I had to update a few of the tests.  In xfpga, the tests for sysfs.c in particular had some cases that needed to be removed. They relied on checks done on the formatting of the sysfs path string and were no longer valid in some cases.  See xfpga/test_sysfs_c.cpp.

testing/xfpga/test_token_list_c.cpp
tswhison: They needed to be parameterized because they use token_get_parent(), and because token_get_parent() now uses opendir, which requires the mock redirection to be in place before it will work.  These tests were failing until I added the system_->prepare_sysfs(platform_), since the mock sysfs wasn't set up.

r-rojo: 👍

testing/xfpga/test_error_c.cpp
```diff
@@ -63,17 +63,25 @@ class error_c_p
...
+    fake_fme_token_.subdev_instance = 0;
     fake_fme_token_.errors = nullptr;
 
+    filter_ = nullptr;
   }
 
   virtual void TearDown() override {
-    EXPECT_EQ(fpgaDestroyProperties(&filter_), FPGA_OK);
+    if (filter_) {
+      EXPECT_EQ(fpgaDestroyProperties(&filter_), FPGA_OK);
```
r-rojo: Should we set filter_ to nullptr here?

tswhison: done

PR #812, opened 2018-10-19
add test_pyopae binary and some tests

r-rojo:
test_pyopae binary uses pybind11 to embed the python interpreter and to
create opae.fpga as an embedded module along with test_system,
test_platform, and test_device.

test_pyopae uses the first argument to determine if it should run unit
tests defined in an external file. If the first command line argument is
`test`, it evaluates the script identified by the next argument.
This script is assumed to contain tests compatible with the Python
unittest module.

If the first argument is not `test`, the script will be simply evaluated within the context of the embedded interpreter.

In order to embed the opae.fpga module, a preprocessor definition was
added to control how the pybind11 module is defnined (embedded or not).
It also exposes fpgaInitailize as a module method, `initialize` when it
is embedded. This allows calling initialize when running on mock
devices.

This commit also adds two Python test scripts, test_properties.py and
test_shared_buffers.py and adds CMake tests for running them with
test_pyopae.

test_system is changed to include an atomic bool variable to indicate if
it has been initialized. If it hasn't, it will pass through to glibc
calls open and close.
TODO: Add similar logic to the other mocked calls.

### DIFF
--- a/testing/mock/test_system.cpp
+++ b/testing/mock/test_system.cpp
@@ -180,7 +180,7 @@ test_device test_device::unknown() {
 
 test_system *test_system::instance_ = nullptr;
 
-test_system::test_system() : root_("") {
+test_system::test_system() : initialized_(false), root_("") {
   open_ = (open_func)dlsym(RTLD_NEXT, "open");
   open_create_ = (open_create_func)open_;
   read_ = (read_func)dlsym(RTLD_NEXT, "read");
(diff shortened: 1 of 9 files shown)


testing/mock/test_system.cpp
```diff
@@ -396,6 +401,9 @@ uint32_t get_device_id(const std::string &sysclass) {
 }
 
 int test_system::open(const std::string &path, int flags) {
+  if (!initialized_) {
```
tswhison: Ah.. good catch.

r-rojo: Discovered this testing the embedded Python 🐍

PR #813, opened 2018-10-19
opae-c: add fpgaFinalize API

tswhison:
Performs OPAE instance destruction. (counterpart to fpgaInitialize)


tswhison:
@r-rojo as with multiple calls to fpgaInitialize() from opae_init(), we safeguard against errant behavior.  In the case of fpgaFinalize(), since the adapter_list is freed and set back to NULL, subsequent calls simply pass through and return success.

PR #818, opened 2018-10-19
Added tests for samples/object-api-c

jelonanX:


testing/object_api/test_object_api_c.cpp
```diff
@@ -0,0 +1,230 @@
...
+  char *argv[] = { zero, one, two };
+
+  /*
+  ** FIXME object_api will glob sysfs for *perf directory,
+  ** current test_system does not mock glob and will not
+  ** find *perf/ files,
+  ** but this test should run to successful completion
+  ** on hardware, and the fn should return FPGA_OK.
+  */
+  EXPECT_NE(object_api_main(3, argv), FPGA_OK);
```
r-rojo: Shouldn't we duplicate this test case for HW but the assertion would be EXPECT_EQ(main, 0)?

testing/object_api/test_object_api_c.cpp
```diff
@@ -0,0 +1,230 @@
...
+  char *argv[] = { zero, one };
+
+  EXPECT_NE(object_api_main(2, argv), 0);
+}
+
+INSTANTIATE_TEST_CASE_P(object_api_c, object_api_c_p,
+                        ::testing::ValuesIn(test_platform::platforms({"skx-p"})));
+
+
+class object_api_c_mock_p : public ::testing::TestWithParam<std::string> {
```
r-rojo: Is this class exactly the same as the previous fixture? Can we just inherit from the `object_api_c_p` instead?

testing/object_api/test_object_api_c.cpp
```diff
@@ -0,0 +1,230 @@
...
+  char one[20];
+  char two[20];
+  char three[20];
+  strcpy(zero, "object_api");
+  strcpy(one, "-B");
+  strcpy(two, "3");
+
+  char *argv[] = { zero, one, two };
+
+  EXPECT_EQ(parse_args(3, argv), FPGA_OK);
```
tswhison: Can we also verify that the global config struct was populated correctly here?

tswhison:
The coverage is only at 49.3%.  Can we bring it up by adding tests?

ibarraza:
I haven't checked the test cases, but @jelonanX if this only works on hw, do you include tests that only work on hw and does the coverage improve significantly when run on hw?

jelonanX:
Ran tests on MCP and it passed.

tswhison:
@jelonanX I'm sure the coverage numbers were increased.  Any idea what we're looking at now with the combined effect?

jelonanX:
I don't have the exact number. With only mock we are getting just below 50%. My guess is at least 80%. 90%> if we are optimistic. Definitely 100% on functions.

jelonanX:
actual coverage number is 89.7%.

tswhison:
awesome!

southerngs:
`cache` performance counter is not valid for DCP.

PR #820, opened 2018-10-19
Update xfpga open close tests to work on hw

jelonanX:


testing/xfpga/test_open_close_c.cpp
```diff
@@ -316,15 +316,16 @@ TEST_P(openclose_c_p, close_03) {
...
+  fpga_handle h2 = NULL;
 
   EXPECT_EQ(FPGA_OK, xfpga_fpgaOpen(tokens_[0], &h1, FPGA_OPEN_SHARED));
   EXPECT_EQ(FPGA_OK, xfpga_fpgaOpen(tokens_[0], &h2, FPGA_OPEN_SHARED));
   EXPECT_EQ(FPGA_OK, xfpga_fpgaClose(h1));
   EXPECT_EQ(FPGA_OK, xfpga_fpgaClose(h2));
 }
 
-INSTANTIATE_TEST_CASE_P(openclose_c, openclose_c_p, ::testing::ValuesIn(test_platform::keys(true)));
+INSTANTIATE_TEST_CASE_P(openclose_c, openclose_c_p, ::testing::ValuesIn(test_platform::platforms({"skx-p"})));
```
ibarraza: Why are these tests exclusive to hw and only on mcp? They should be good for hw and mock for all platforms, right?

testing/xfpga/test_open_close_c.cpp
```diff
@@ -359,3 +381,5 @@ TEST(openclose_c, invalid_open_close) {
 #endif
 }
 
+INSTANTIATE_TEST_CASE_P(openclose_c, openclose_c_mock_p, 
+                        ::testing::ValuesIn(test_platform::mock_platforms({"skx-p"})));
```
ibarraza: Why is this test not running for all mock platforms?

testing/xfpga/test_open_close_c.cpp
ibarraza: Or I guess the only one missing is for dcp.

testing/xfpga/test_open_close_c.cpp
```diff
@@ -316,15 +316,17 @@ TEST_P(openclose_c_p, close_03) {
  *             already opened token returns FPGA_OK.
  */
 TEST_P(openclose_c_p, open_share) {
-  fpga_handle h1, h2;
+  fpga_handle h1 = NULL;
+  fpga_handle h2 = NULL;
```
tswhison: Probably nullptr.  Looks ok otherwise.

testing/xfpga/test_open_close_c.cpp
```diff
@@ -333,7 +335,28 @@ INSTANTIATE_TEST_CASE_P(openclose_c, openclose_c_p, ::testing::ValuesIn(test_pla
...
+//    system_->initialize();
+//    system_->prepare_syfs(platform_);
+//  }
+//
+//  virtual void TearDown() override {
+//    system_->finalize();
+//  }
+//
+//  test_platform platform_;
+//  test_system *system_;
```
ibarraza: Remove instead of comment out.

testing/xfpga/test_open_close_c.cpp
```diff
@@ -333,7 +335,14 @@ INSTANTIATE_TEST_CASE_P(openclose_c, openclose_c_p, ::testing::ValuesIn(test_pla
  *             but driver is not loaded. the function returns FPGA_NO_DRIVER.
  *
  */
-TEST(openclose_c, invalid_open_close) {
+class openclose_c_mock_p
+    : public ::testing::TestWithParam<std::string> {
+ protected:
+  openclose_c_mock_p() {}
+
+};
```
r-rojo: You probably want to derive from the fixture defined above (openclose_c_p)
```c++
class openclose_c_mock_p : public openclose_c_p {
};
```

jelonanX: To my understanding, I don't believe we want to derive from the openclose_c_p class because it will load in the driver and this tests for FPGA_NO_DRIVER on fpgaOpen. I could be wrong.

ibarraza: This is correct. The test expects there to be no driver, so we do not want to setup like in the other fixture. This test is missing a comment header stating it's test case though.

PR #822, opened 2018-10-19
Porting fpgainfo to C: fme, port, errors, and temp subcommands

zhangzhang10:
- Replace fpgainfo Python scripts with C implementation, using
Scott Gibbons' code.
- Use the new Obj API where it is applicable.
- Temporarily disable power subcommand. Will add back later.


tools/base/argsfilter/argsfilter.c
```diff
@@ -0,0 +1,199 @@
...
+	} while (0)
+
+int set_properties_from_args(fpga_properties filter, fpga_result *result,
+			     int *argc, char *argv[])
+{
+	// prefix the short options with '-' so that unrecognized options are
+	// ignored
+	const char *short_opts = "-:B:D:F:S:";
+	struct option longopts[] = {
+		{"bus", required_argument, NULL, 'B'},
```
tswhison: Seems like this should also support the PCIe segment portion of the address.

tools/base/fpgainfo/main.c
```diff
@@ -0,0 +1,264 @@
...
+	       "Usage:\n"
+	       "        fpgainfo [-h] [-B <bus>] [-D <device>] "
+	       "[-F <function>] [-S <socket-id>] {");
+	printf("%s", cmd_array[0].command);
+	for (i = 1; i < sizeof(cmd_array) / sizeof(cmd_array[0]); i++) {
+		printf(",%s", cmd_array[i].command);
+	}
+	printf("}\n\n"
+	       "                -h,--help           Print this help\n"
+	       "                -B,--bus            Set target bus number\n"
```
tswhison: PCIe segment?

tools/base/argsfilter/argsfilter.c
r-rojo: As a short and long option or just as a long option?

tools/base/fpgainfo/errors.c
```diff
@@ -0,0 +1,480 @@
...
+				size = PCIE1_ERROR_COUNT;
+				error_string = PCIE1_ERROR;
+			}
+
+			for (j = 0; (j < size) && (NULL != error_string); j++) {
+				if (FPGA_BIT_IS_SET(error_value, j)) {
+					printf("\t %s \n", error_string[j]);
+				}
+			}
+		}
```
r-rojo: Let's think about how we may refactor lines 355-386 (maybe in another PR?)

tools/base/fpgainfo/errors.c
```diff
@@ -0,0 +1,480 @@
...
+			} else if (!strcmp(errinfos[i].name, "First Error")) {
+				size = 0;
+				error_string = NULL;
+			}
+
+			for (j = 0; (j < size) && (NULL != error_string); j++) {
+				if (FPGA_BIT_IS_SET(error_value, j)) {
+					printf("\t %s \n", error_string[j]);
+				}
+			}
```
r-rojo: Same here, let's think about how this can be refactored into a common function (394-418)

tools/base/argsfilter/argsfilter.c
tswhison: Just as a long opt would be sufficient to match the other apps which have it.  Use --segment.

PR #829, opened 2018-10-22
Seperate xfpga errors mock tests and hw tests

jelonanX:


testing/xfpga/test_error_c.cpp
```diff
@@ -92,22 +92,34 @@ class error_c_p
...
-    result = std::system(cmd.c_str());
-    (void) result;
-    return 1;
+    cmd = "rm -rf " + port_sysfspath;
+    goto remove;
   }
   else {return -1;}
+
+remove:
+  result = std::system(cmd.c_str());
```
ibarraza: Can we use an EXPECT_EQ here to check the return value and remove result altogether?

PR #830, opened 2018-10-22
doc: add plugin developer's guide

tswhison:


lzhan55:
This developer's guide looks very good. Could you also add the explanation of the platform data table for users to add new plugins?

tswhison:
@zhangzhang10 Thanks, could you approve [840](https://github.com/OPAE/opae-sdk/pull/830) as well? :)

PR #836, opened 2018-10-23
Seperate umsg tests to run on hw

jelonanX:
*Hugepages must be allocated prior to running the tests. Else, mmap will fail and return FPGA_NO_MEMORY.


ibarraza:
All of the positive test cases (test_umsg_drv_01, test_umsg_drv_04, and test_umsg_08) will fail on DCP and should be moved to a test fixture that is for only MCP.

PR #837, opened 2018-10-23
tests: update hello_events tests for hw

tswhison:
Fix bug in fpgad. The fme.x/errors/fatal_errors was renamed to
fme.x/errors/catfatal_errors in the driver.

### DIFF
--- a/testing/hello_events/test_hello_events_c.cpp
+++ b/testing/hello_events/test_hello_events_c.cpp
@@ -74,7 +74,7 @@ using namespace opae::testing;
 
 class hello_events_c_p : public ::testing::TestWithParam<std::string> {
  protected:
-  hello_events_c_p() {}
+  hello_events_c_p() : token_(nullptr) {}
 
   virtual void SetUp() override {
     std::string platform_key = GetParam();
@@ -86,8 +86,6 @@ class hello_events_c_p : public ::testing::TestWithParam<std::string> {
 
     EXPECT_EQ(fpgaInitialize(NULL), FPGA_OK);
 
-    token_ = nullptr;
-
     optind = 0;
     events_config_ = events_config;
   }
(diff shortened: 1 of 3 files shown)


testing/hello_events/test_hello_events_c.cpp
```diff
@@ -344,10 +342,39 @@ class hello_events_c_fpgad_p : public ::testing::TestWithParam<std::string> {
...
+
+/**
+ * @test       main2
+ * @brief      Test: hello_events_main
+ * @details    When a suitable accelerator device is found,<br>
+ *             hello_events_main (hw) runs to completion,<br>
  *             and the fn returns zero.<br>
  */
-TEST_P(hello_events_c_fpgad_p, main2) {
+TEST_P(hw_hello_events_c_fpgad_p, main2) {
```
ibarraza: This test seems to be failing for MCP with FPGA_EXCEPTION - poll timeout occurred.

tswhison: Worked for me on cpt-ncf022.  Did you try running from my branch?

tswhison:
@ibarraza I don't see the issue with 224.  Do you mean the bus initialization to 99?  That's just setup for the get_bus call.

ibarraza:
@tswhison Ah, I think the issue is that find_fpga() returns bus 0xbe on my system but the platform_.devices[0].bus is 0x81.

testing/hello_events/test_hello_events_c.cpp
jelonanX: This test is also failing for me on bmp012 with error registering FME event, FPGA_NO_DAEMON.

jelonanX: It passes when running with sudo.

tswhison: Ah yes, sudo is required, since we are trying to open a handle to the FME.  This wouldn't be required on mock, of course, since there is no enforcement of device permissions.  Not sure how to best handle this.  @r-rojo Thought about tests that require sudo?

ibarraza: Might just assume that when running on hw that the test runner has root privileges.

tswhison:
We had a lot of conversation on this one.  Are we good to go, so long as we run this as root on hw?

PR #839, opened 2018-10-24
Updated xfpga reconf tests with expected hw platform behavior.

ibarraza:
Updated the test expectations with the expected hw platform behavior.
Updated the have_powermgmt check to be specific to the device that we are running on.


ibarraza:
Will retarget to master.

PR #840, opened 2018-10-24
Tswhison/plugin to master

tswhison:
This is a request to merge feature/plugin-api to master.  Please review diligently.


.travis.yml
```diff
@@ -129,6 +153,26 @@ matrix:
...
+#         - doxygen
+#   - script: scripts/build-tests-default.sh
+#     cache: ccache
+#     addons:
+#       apt:
+#         packages:
+#         - libjson-c-dev
+#         - libhwloc-dev
+#         - uuid-dev
+#         - doxygen
```
r-rojo: Should we get rid of this?

CMakeLists.txt
```diff
@@ -104,8 +104,9 @@ set(LIBRARY_OUTPUT_PATH ${CMAKE_BINARY_DIR}/lib CACHE PATH "Build directory" FOR
 
 set(COMMON_DIR        ${CMAKE_CURRENT_SOURCE_DIR}/common)
 set(OPAE_INCLUDE_DIR  ${COMMON_DIR}/include CACHE PATH "Path to include directory" FORCE)
-add_custom_target(copy-common-opae-header-files ALL
-  COMMAND ${CMAKE_COMMAND} -E copy_directory ${OPAE_INCLUDE_DIR} ${CMAKE_BINARY_DIR}/include)
+
+#add_custom_target(copy-common-opae-header-files ALL
+#  COMMAND ${CMAKE_COMMAND} -E copy_directory ${OPAE_INCLUDE_DIR} ${CMAKE_BINARY_DIR}/include)
```
r-rojo: remove dead code

CMakeLists.txt
```diff
@@ -184,10 +185,20 @@ mark_as_advanced(BUILD_TESTS)
 option(BUILD_ASE_TESTS "Enable ASE test compilation." OFF)
 mark_as_advanced(BUILD_ASE_TESTS)
 
+## TODO: Remove this once this replaces old school tests
+option(NEW_UNITS "Use new testing framework" OFF)
```
r-rojo: Are we good to remove this option?

CMakeLists.txt
```diff
@@ -184,10 +185,20 @@ mark_as_advanced(BUILD_TESTS)
...
+mark_as_advanced(NEW_UNITS)
+option(ENABLE_MOCK "Enable mock driver for testing." ON)
+mark_as_advanced(ENABLE_MOCK)
+
 if(BUILD_TESTS OR BUILD_ASE_TESTS)
   enable_testing()
   include(CTest)
-  add_subdirectory(tests)
+  if (NEW_UNITS)
```
r-rojo: Maybe we can remove NEW_UNITS option and replace this line with BUILD_TESTS

.travis.yml
tswhison: @nakulkorde?

CMakeLists.txt
tswhison: Agreed

CMakeLists.txt
tswhison: I think so.

CMakeLists.txt
tswhison: I think so.

.travis.yml
nakulkorde: yes I would agree. Also, the test-build script which is supposed to run three different types with different build types is being run with RelWithDebInfo in all three. I will fix that and add to this PR in a few minutes.

ibarraza:
@lzhan55 Based on what those tests have in common, I'm guessing that hugepages have not been allocated.

tswhison:
@lzhan55 
```sh
sudo sh -c 'echo 8 > /sys/kernel/mm/hugepages/hugepages-2048kB/nr_hugepages'
./scripts/cover.sh
```

michael-adler:
@lzhan55 - The diagnosis of no huge pages is unlikely for ASE since ASE is supposed to emulate allocation of huge pages even if they aren't available on the machine. It's possible there is a bug in that emulation code, but it is supposed to work without host huge pages.

tswhison:
@michael-adler she's trying to run the OPAE unit tests, not the ASE ones.

michael-adler:
Sorry. I was confused by "ase development machine".

michael-adler:
@tswhison - The new code has some pthreads compilation failures on older systems (like our standard EC SUSE machines). See branch mcadler/tswhison_plugin_pthread with changes that compile. There is probably a better solution, but the definition of _GNU_SOURCE has to be before other includes.

michael-adler:
@tswhison - fpgaconf fails on BDX, saying no suitable slot. I will investigate.

tswhison:
@michael-adler Thanks for the updates!  I will consume your branch and wait to hear back on fpgaconf.

libopae/api-shell.c
```diff
@@ -803,6 +803,13 @@ fpga_result fpgaPrepareBuffer(fpga_handle handle, uint64_t len, void **buf_addr,
 	opae_wrapped_handle *wrapped_handle =
 		opae_validate_wrapped_handle(handle);
 
+	// A special case: respond FPGA_OK when !buf_addr and !len
+	// as an indication that FPGA_BUF_PREALLOCATED is supported
+	// by the library.
+	if ((flags & FPGA_BUF_PREALLOCATED) && !buf_addr && !len) {
```
r-rojo: Is this documented in the doxygen markup?

r-rojo: It may be worth adding a \@note  to the API docs to reflect this.

tswhison: @r-rojo I added it in 'opae-c: plugins decide FPGA_BUF_PREALLOCATED'

PR #855, opened 2018-10-26
Updated xfpga enum tests to work on hw.

ibarraza:
Updated the xfpga enum tests to get the number of fpgas in the system without using a hardcoded value or relying on fpgaEnumerate(). It will only do this on hw. Might be a better way to do this.
Separated all of the tests into single test case only tests. This also fixed a bunch of memory leakage in the tests.
Added comments for each test case.
Updated the fme num_errors to be consistent with hw. Changed mock platform value as well as the mock sysfs tar.


testing/xfpga/test_enum_c.cpp
```diff
@@ -74,6 +74,58 @@ class enum_c_p : public ::testing::TestWithParam<std::string> {
...
+  // Need a concrete way to determine the number of fpgas on the system
+  // without relying on fpgaEnumerate() since that is the function that
+  // is under test. 
+  int GetNumFpgas() {
+    if (platform_.mock_sysfs != nullptr) {
+      return platform_.devices.size();
+    }
+
+    int value;
+    std::string cmd = "lspci | "
```
tswhison: grep 'intel-fpga-dev' /sys/class/fpga | wc -l
Would also work, and it would be closer to the driver.

testing/xfpga/test_enum_c.cpp
```diff
@@ -198,14 +377,29 @@ TEST_P(enum_c_p, function) {
...
+ *             match that socket_id.
+ */
 TEST_P(enum_c_p, socket_id) {
   auto device = platform_.devices[0];
   ASSERT_EQ(fpgaPropertiesSetSocketID(filter_, device.socket_id), FPGA_OK);
   EXPECT_EQ(
       xfpga_fpgaEnumerate(&filter_, 1, tokens_.data(), tokens_.size(), &num_matches_),
       FPGA_OK);
-  EXPECT_EQ(num_matches_, 2);
+  EXPECT_EQ(num_matches_, GetNumFpgaTypes() * 2);
```
r-rojo: I don't follow what this test is doing. socket_id is 0 on pac systems, and can be 0, 1, ... on integrated systems.

ibarraza: Right, and as the test infrastructure is currently implemented, the test will run one test for each platform detected on the system. The current platform socket will always be 0, MCP or DCP, and we enumerate for that. So we expect that fpgaEnumerate() will return 2 matches for MCP and, if detected, 2 matches for DCP.

PR #857, opened 2018-10-26
tools: update fpga_token in legacy C++ API

tswhison:


michael-adler:
No, it's much worse than this. The new API adds a new level of indirection. I hope you will soon remove the need for this and add ways to probe properties of a token such as sysfspath.

tswhison:
Doh, good catch.  I wasn't seeing the big picture.

PR #877, opened 2018-10-26
pyopae: add APIs to Python bindings

r-rojo:
* Add a memory_barrier call at the module level
  Not sure if it is needed for Python as it is an interpreted language
  but calling this function calls the C++11 atomic_thread_fence with
  release semantics

* Add three polling routines to shared_buffer class. This allows for
  polling in C++ without any Python overhead

* Add a copy API function in the shared_buffer class to copy bytes from
  one shared_buffer to another

* Add 32-bit and 64-bit versions of shared_buffer read/write

* Add a shared_buffer split routine that can split a shared buffer
  in Python into other buffers using the first buffer as a parent
  object. It was necessary to change the C++ shared_buffer d'tor to
  avoid double frees


pyopae/pyshared_buffer.cpp
```diff
@@ -134,3 +134,89 @@ py::list shared_buffer_getslice(shared_buffer::ptr_t buf, py::slice slice) {
...
+  uint8_t *src = const_cast<uint8_t *>(self->c_type());
+  uint8_t *dst = const_cast<uint8_t *>(other->c_type());
+
+  std::copy(src, src + (size ? size : self->size()), dst);
+}
+
+const char *shared_buffer_doc_split() {
+  return R"opaedoc(
+    Split the buffer into other shared_buffer objects.
+    The arguments to this method is a list of sizes to use when splitting the buffer.
```
tswhison: s/is/are/

PR #883, opened 2018-10-30
Updated cxx core object tests to work on hw.

ibarraza:
Updated the cxx core object tests to enumerate based on device id.
Changed the object write tests to target the write to fme/*perf/fabric/freeze instead of errors since on hw it would not be able to write to the file.
Updated the mcp mock sysfs tar with fme/iperf/fabric/freeze.


testing/opae-cxx/test_object_cxx_core.cpp
```diff
@@ -47,21 +47,36 @@ class sysobject_cxx_p : public ::testing::TestWithParam<std::string> {
...
+    props_dev->device_id = platform_.devices[0].device_id;
+
+    tokens_dev_ = token::enumerate({props_dev});
+    ASSERT_GT(tokens_dev_.size(), 0);
+    handle_dev_ = handle::open(tokens_dev_[0], 0);
+    ASSERT_NE(handle_dev_.get(), nullptr);
   }
 
   virtual void TearDown() override {
     system_->finalize();
```
tswhison: For similar reasons as the C API, shouldn't system_finalize(); appear after objects are destroyed?

testing/opae-cxx/test_object_cxx_core.cpp
```diff
@@ -117,9 +132,17 @@ TEST_P(sysobject_cxx_p, handle_object) {
  * Then no exceptions are thrown
  */
 TEST_P(sysobject_cxx_p, handle_object_write) {
-  auto obj = sysobject::get(handle_, "errors/errors");
+  std::string path = "iperf/fabric/freeze";;
+
+  if (platform_.devices[0].device_id == 0x09c4 ||
```
tswhison: Hmm.. do we need an is_dcp() function?

ibarraza: Probably should implement something like this in fpga_hw.cpp at some point.

jelonanX:
These tests fails on 012, token_object, handle_object, read_bytes.

PR #885, opened 2018-10-30
Separate and verify userclk tool tests work on hw.

jelonanX:


testing/userclk/test_userclk_c.cpp
```diff
@@ -322,7 +330,196 @@ TEST_P(userclk_c_p, main2) {
...
+  strcpy(one, "--segment");
+  sprintf(two, "%d", platform_.devices[0].segment);
+  strcpy(three, "-B");
+  sprintf(four, "%d", platform_.devices[0].bus);
+  strcpy(five, "-D");
+  sprintf(six, "%d", platform_.devices[0].device);
+  strcpy(seven, "-F");
+  sprintf(eight, "%d", platform_.devices[0].function);
+  strcpy(nine, "-S");
+  if (platform_.devices[0].bus == 129) { strcpy(ten, "0x0"); }
```
tswhison: Did you touch base with @r-rojo on this one?  Is there a better way we can do this?

tswhison:
@jelonanX test error is
../lib/libfpga_db.so: undefined reference to `strncpy_s'
make[3]: *** [bin/test_fpgad_daemonize_c] Error 1
make[2]: *** [testing/CMakeFiles/test_fpgad_daemonize_c.dir/all] Error 2
make[2]: *** Waiting for unfinished jobs....

testing/platform/fpga_hw.cpp
```diff
@@ -241,6 +244,53 @@ static T parse_file(const std::string &path) {
...
+}
+
+static uint16_t read_socket_id(const std::string devices) {
+  std::string glob_path = PCI_DEVICES + "/" + devices + "/fpga/intel-fpga-dev.*/intel-fpga-fme.*/socket_id";
+  std::string socket_path;
+
+  glob_t glob_buf;
+  glob_buf.gl_pathc = 0;
+  glob_buf.gl_pathv = NULL;
+  int globres = glob(glob_path.c_str(), 0, NULL, &glob_buf);
```
tswhison: Check valgrind on this.  I recall that it's possible to have a negative return value, but for glob to actually allocate data in glob_buf.

testing/opae-c/test_enum_c.cpp
```diff
@@ -368,7 +310,11 @@ TEST_P(enum_c_p, socket_id) {
   EXPECT_EQ(
       fpgaEnumerate(&filter_, 1, tokens_.data(), tokens_.size(), &num_matches_),
       FPGA_OK);
-  EXPECT_EQ(num_matches_, GetNumFpgaTypes() * 2);
+  if (device.socket_id == 0) {
+    EXPECT_EQ(num_matches_, GetNumFpgaTypes() * 2);
+  } else {
+    EXPECT_EQ(num_matches_, GetNumFpgaTypes());
```
ibarraza: These checks will fail depending on the system. We need to check that num_matches_ is equal to the number of devices on the system that match the socket id.

testing/xfpga/test_enum_c.cpp
```diff
@@ -390,7 +390,11 @@ TEST_P(enum_c_p, socket_id) {
   EXPECT_EQ(
       xfpga_fpgaEnumerate(&filter_, 1, tokens_.data(), tokens_.size(), &num_matches_),
       FPGA_OK);
-  EXPECT_EQ(num_matches_, GetNumFpgaTypes() * 2);
+  if (device.socket_id == 0) {
+    EXPECT_EQ(num_matches_, GetNumFpgaTypes() * 2);
+  } else {
+    EXPECT_EQ(num_matches_, GetNumFpgaTypes());
```
ibarraza: These checks will fail depending on the system. We need to check that num_matches_ is equal to the number of devices on the system that match the socket id.

jelonanX:
@tswhison the test main2 in test_hello_events_c fails on 37. The log suggests an invalid fme error revision: 0 in the sysfs. I verify that other systems have it set to 1. Any thoughts on the failure?

tswhison:
I'll bet you're getting that with the DCP token.  I think that fme.x/errors/revision is 0 on DCP.

testing/opae-c/test_enum_c.cpp
```diff
@@ -702,7 +648,7 @@ TEST(wrapper, validate) {
 }
 
 INSTANTIATE_TEST_CASE_P(enum_c, enum_c_p,
-                        ::testing::ValuesIn(test_platform::keys(true)));
+                        ::testing::ValuesIn(test_platform::platforms({"skx-p", "dcp-rc"})));
```
ibarraza: This will need to be an empty list or it will fail on systems with MCP only.

PR #893, opened 2018-11-01
ASE read/write line message sizes were different sizes on 32 bit simulators

michael-adler:
Stop using pointers and enums and align data structures to 64 bits
in messages passed between the RTL simulator and the client.

### DIFF
--- a/ase/sw/app_backend.c
+++ b/ase/sw/app_backend.c
@@ -1152,13 +1152,13 @@ void shm_error(const char *msg)
 /*
  * Note pinned pages. This is used only for logging in the simulator.
  */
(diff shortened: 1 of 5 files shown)


michael-adler:
Found this rather disastrous bug while working on simulator support for Quartus objects. On 32 bit simulators the data structures for the new memory read/write support align differently and read/write requests were messed up.

Tested on 32 bit modelsim, ran the usual tests, new code appears ok.

http://sw-pert.altera.com/pert/pert.php?test_run_id=2793469
NLB failures are unrelated. DMA is the usual licensing problems, so I ran the DMA test by hand on 32 bit modelsim. It passed buffer verification, confirming reads and writes working.

PR #894, opened 2018-11-01
ASE now invokes Quartus to choose megafunction and tech. libraries

michael-adler:
- Pick the right libraries for the target platform automatically.
- Add support for VHDL that needs megafunctions, etc.

### DIFF
--- a/ase/Makefile
+++ b/ase/Makefile
@@ -44,6 +44,14 @@
 
 include ase_sources.mk
 
+ifneq ($(wildcard rtl/ase_platform_config.mk),)
+  # Load platform interface manager's configuration
+  include rtl/ase_platform_config.mk
+else
+  # No platform interface manager. Assume Arria 10.
+  PLATFORM_FPGA_FAMILY_A10 = 1
+endif
+
 ###############################################################
 ##                                                           ##
 ##         ASE Platform value (selection method)             ##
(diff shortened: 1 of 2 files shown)


michael-adler:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2794108
http://sw-pert.altera.com/pert/pert.php?test_run_id=2794166

This fixes about 3 bug entries. VHDL is fully supported, including when it needs megafunctions and atoms. The right atoms are pulled in for the target architecture, using Quartus to choose the simulation libraries. (You can see in the tests that different atoms are chosen in the two runs.)

The ASE Makefile is cleaned up so that these simulation libraries are compiled exactly once, even when make is invoked multiple times.

ase/Makefile
```diff
@@ -447,9 +466,7 @@ help: header
 	@echo "#################################################################"
 
 ## Build ASE Software objects and shared library ##
-sw_build:
-	make header
-	mkdir -p $(WORK)
+sw_build: $(SIM_QUARTUS_VERILOG) $(SIM_QUARTUS_VHDL)
```
lzhan55: Why sw_build depends on $(SIM_QUARTUS_VERILOG) and $(SIM_QUARTUS_VHDL) here?

michael-adler: To avoid a race on creating the $(WORK) directory.

ase/Makefile
```diff
@@ -243,6 +226,55 @@ ASE_LD_SWITCHES+= -lrt -lpthread
...
+  FPGA_FAMILY = stratix10
+endif
+$(info FPGA_FAMILY=$(FPGA_FAMILY))
+$(info )
+
+# Gate level libraries to add to simulation
+GLS_VERILOG_OPT ?=
+QSIM_MODEL_VLOG_OPT ?=
+
+# The Quartus library simulation targets change based on the simulator.
```
lzhan55: Will these Quartus library simulation targets be used by ASE?

michael-adler: Yes, by the user RTL and Qsys-generated code.

lzhan55: Will these only be used some user RTL which needs to generate RTL files from Qsys? Thank you!

michael-adler: It is possible to write RTL that doesn't need them, though most real programs will need Quartus libraries. There's no way to probe the code and figure it out, so we now use Quartus to tell ASE how to compile the base libraries. It only takes a minute or two and it is only done once, even if make is invoked multiple times.

PR #895, opened 2018-11-01
Move valid Mock_errinj test to mock only

jelonanX:
When running fpgaAssignPortToInterface on hw, the test releases a port (/dev/intel-fpga-port1) and re-assigns it. However, the file permission is set to root rw only. This causes the rest of the hw tests to fail/segfaults. This PR moves that test to mock platform only.


jelonanX:
Upon further investigating. It turns out the problem was assigning before releasing. The order should be release then re-assign.

PR #897, opened 2018-11-02
Remove unnecessary "cd -" commands from ASE Makefile

michael-adler:


michael-adler:
Chip pointed out that "cd -" at the end of a line does nothing. Removed them.

http://sw-pert.altera.com/pert/pert.php?test_run_id=2796060

PR #899, opened 2018-11-02
libopae: remove static functions in bmc module

anandaravuri:
Changes:
    - Remove static functions in bmc module c files
    - Add functions in bmc c files


tswhison:
You could also use STATIC.  We define it to STATIC for non-test builds and to the empty string for the tests.

PR #901, opened 2018-11-02
Mask the ASE PA/VA XOR mask one bit smaller

michael-adler:
This solves a regression when simulating on BDX (HARP systems).

### DIFF
--- a/ase/sw/ase_host_memory.c
+++ b/ase/sw/ase_host_memory.c
@@ -107,6 +107,20 @@ int ase_host_memory_unpin(uint64_t iova, uint64_t length)
 
 
 /*
+ * Generate an XOR mask that will be used to map between virtual and physical
+ * addresses.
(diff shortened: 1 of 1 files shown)


michael-adler:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2798454

michael-adler:
It's more complicated than that. Our BDX implementation used a segment of the QPI space, allowing only 38 bits for physical addresses. Later machines demand the full 48 bit VA space. The addresses handed out during simulation by Linux typically have all 1's in the high bits except for bit 48. That bit is always 0 in user space and always one in kernel space. That top zero bit was causing all generated PAs to have the high bit set when inverted, leading to dropped address bits in MPF/VTP in BDX simulations. This hack makes BDX simulation more reliable without major impact to simulation of later systems.

PR #904, opened 2018-11-02
Updated metrics tests to work on hw.

ibarraza:
Updated the check in SetUp() to only ensure that fpgaEnumerate returns at least one match. We are not concerned with the number of matches which may be different on different systems.


ibarraza:
@OPAE/ci Can I get a review please?

PR #915, opened 2018-11-06
Address xfpga reconf tests mem leaks

jelonanX:


testing/xfpga/test_reconf_c.cpp
```diff
@@ -96,13 +94,14 @@ class reconf_c : public ::testing::TestWithParam<std::string> {
 
   virtual void TearDown() override {
     EXPECT_EQ(fpgaDestroyProperties(&filter_), FPGA_OK);
+    if (handle_) { EXPECT_EQ(xfpga_fpgaClose(handle_), FPGA_OK); }
```
tswhison: Also, set handle_ back to nullptr after closing it.

PR #920, opened 2018-11-07
pydiag: Add implementation of cool_cpu_cache opt.

r-rojo:


tools/extra/pyfpgadiag/opae/tools/fpgadiag/nlb3.py
```diff
@@ -26,15 +26,19 @@
...
 from diagtest import diagtest
 from opae.utils import cl_align, CACHELINE_BYTES
+
 # pylint: disable=E0611
 from opae import fpga
 
+
 COOL_CACHE_LINES = 1024
 COOL_CACHE_SIZE = CACHELINE_BYTES*COOL_CACHE_LINES
+MAX_CPU_CACHE_SIZE = 100 * 1024 * 1024
```
tswhison: I thought the idea was to warm/cool the data cache within the fpga device.  The implementations of warm_fpga_cache and cool_fpga_cache look like they are doing that.  I wonder if this should be renamed to MAX_FPGA_CACHE_SIZE and if _cpu_cache_buffer should be _fpga_cache_buffer for clarity?

r-rojo: this is for the --cool-cpu-cache option, this operation shouldn't touch fpga cache

PR #922, opened 2018-11-07
fpgad: close client connections on exit

tswhison:
When run repeatedly in our test environment, server_thread was always
accumulating client sockets onto the end of pollfds. Eventually, we
would hit the limit and fail when given a high iteration count. This
change causes server_thread to close all client sockets on exit, and
resets num_fds to 1 when the thread is started to ensure correctness.


tswhison:
Thanks for pointing this out, @jelonanX!

PR #931, opened 2018-11-09
pydiag: add processing of multi-cl argument

r-rojo:
multi-cl size is either 0x20 (for multi-cl size of 2) or 0x60 (for multi-cl size of 4)


michael-adler:
I was unable to test this due to multiple failures when running, e.g. "fpgadiag -m write -r vh0 --begin=32768 -u 4".

In nlb3:setup_buffers(), src.fill(0xc0cac01a) fails at run time because the value is too large for an int. You should probably change the formal param to unsigned int, or 64 bit unsigned int. I changed the value there to something at accepted and then end_fabric was undefined at diagtest.py:391. This was on BDX, in case that affects the assignment of end_fabric.

michael-adler:
Do you have access to a PAC card? It still fails. First it prints a message that no handler is found for "cache_counters", then it fails with the previously noted end_fabric problem.

BDX fails, too, with the same problem.

michael-adler:
Two bugs remain:

1. On DCP, it prints No handlers could be found for logger "cache_counters".
2. The C++ version has the timeout default value set to 1 second. Specifying just --cont results in a 1 second run. The Python version doesn't have a default time, so just runs one trip with --cont and no time argument. I like the C++ behavior.

PR #933, opened 2018-11-12
Change docs target to build _opae before building docs

nakulkorde:


nakulkorde:
Yes I think so. Otherwise the python api spec will not get populated properly. Currently it won't get populated because _opae is getting built after the docs are built, this change resolves that issue.

PR #935, opened 2018-11-12
ASE: Fixed the ASE build issue after OPAE changes

lzhan55:


cmake/modules/test_config_ase.cmake
```diff
@@ -28,27 +28,16 @@ add_library(commonlib-ase SHARED common_test.h common_test.cpp)
 set(COMMON_SRC gtmain.cpp jsonParser.cpp
   unit/gtOpenClose_base.cpp
   unit/gtProperties_base.cpp
-  unit/gtOpen.cpp
```
nakulkorde: Is there a specific reason that several of these tests are removed?

nakulkorde: ok

PR #937, opened 2018-11-13
pydiag: change log level of perf_counters msg

r-rojo:
Change log level of log message emitted by perf_counters to DEBUG.
These messages will only be seen when loglevel of debug is chosen by the
end user.


r-rojo:
It shouldn't be excluded. It does not use `args.SUPPRESS` in its `help` kwarg.

PR #940, opened 2018-11-14
api: Add path restrictions in object api

r-rojo:
Add checks for object names that start with `.` or `/` and disallow if
found.
Add unit tests for these checks

### DIFF
--- a/libopae/plugins/xfpga/sysobject.c
+++ b/libopae/plugins/xfpga/sysobject.c
@@ -49,7 +49,7 @@
 
 #define VALIDATE_NAME(_N)                                                      \
 	do {                                                                   \
-		if (_N[0] == '.' || _N[0] == '/') {                            \
+		if (_N[0] == '.' || _N[0] == '/' || strstr(_N, "..")) {        \
 			FPGA_MSG("%s is not a valid input", _N);               \
 			return FPGA_INVALID_PARAM;                             \
 		}                                                              \
--- a/libopae/plugins/xfpga/sysobject.c
+++ b/libopae/plugins/xfpga/sysobject.c
@@ -47,6 +47,13 @@
 		}                                                              \
 	} while (false);
 
+#define VALIDATE_NAME(_N)                                                      \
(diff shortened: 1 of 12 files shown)


libopae/plugins/xfpga/sysobject.c
```diff
@@ -47,6 +47,13 @@
 		}                                                              \
 	} while (false);
 
+#define VALIDATE_NAME(_N)                                                      \
+	do {                                                                   \
+		if (_N[0] == '.' || _N[0] == '/') {                            \
```
tswhison: What if .. appears, but not as the first character, eg "errors/../../../../../../../../../../../"?

tools/extra/ras/main.c
```diff
@@ -1064,26 +1064,24 @@ fpga_result mmio_error(fpga_handle afu_handle, struct RASCommandLine *rasCmdLine
 	if ( rasCmdLine->function >0 )
 		function = rasCmdLine->bus;
 
-	result = fpgaHandleGetObject(afu_handle, FPAG_DEVICEID_PATH, &dev_object, 0);
+	result = fpgaGetPropertiesFromHandle(afu_handle, &props);
 	if (result != FPGA_OK) {
-		OPAE_ERR("Failed to get Handle Object");
+		OPAE_ERR("Failed to get properteis from handle");
```
tswhison: s/properteis/properties/

zhangzhang10:
@r-rojo @tswhison Do we need to check if a sysobject name may be a symbolic link? This is a required SDL check, but I'm not sure if we should do it here.

tswhison:
@zhangzhang10 Since we're dealing with sysfs hear, the likelihood of encountering a symlink is very high; however, I think we are ok in that the path is always relative to the root stored in the token.  I believe that it is impossible to create an arbitrary rogue symlink in sysfs (at least without a driver).

zhangzhang10:
@tswhison I agree with what you said. Thanks for the explanation.

PR #945, opened 2018-11-16
tests: re-enable pyopae tests

r-rojo:
* Use TRAVIS_PYTHON_VERSION env. variable when configuring the builld
* Check TRAVIS_PYTHON_VERSION in testing/CMakeLists.txt to determine if
  test_pyopae + tests should be built/executed
* Update sysobject Python tests to reflect latest change in the cxx API


ibarraza:
I'm not seeing the output of the tests in the travis build log. It should be in the "Run tests and generate coverage" step, yes? The one that runs scripts/cover.sh where the python variable is defined?

PR #948, opened 2018-11-16
fpgad: harden the workspace

tswhison:
Deprecate --directory and --umask parameters.
Restrict --logfile and --pidfile parameters to allowing file names only.
Set directory based on the current user:
* If root, then use /var/lib/opae
* Otherwise, use ${HOME}/.opae
Set the umask based on the current user:
* If root, deny world read/write and group write.
* Otherwise, deny world write and group write.

### DIFF
--- a/tools/base/fpgad/fpgad.c
+++ b/tools/base/fpgad/fpgad.c
@@ -1,4 +1,4 @@
-// Copyright(c) 2017, Intel Corporation
+// Copyright(c) 2017-2018, Intel Corporation
 //
 // Redistribution  and  use  in source  and  binary  forms,  with  or  without
 // modification, are permitted provided that the following conditions are met:
@@ -39,7 +39,12 @@
 #include "config_int.h"
 #include "log.h"
 #include "ap_event.h"
+#include <stdlib.h>
 #include <getopt.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <errno.h>
 
 #include "safe_string/safe_string.h"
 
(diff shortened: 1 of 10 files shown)


tools/base/fpgad/fpgad.c
```diff
@@ -101,6 +109,103 @@ void sig_handler(int sig, siginfo_t *info, void *unused)
...
+		} else {
+			char cwd[PATH_MAX];
+			// ${HOME} not found - use current dir.
+			if (getcwd(cwd, sizeof(cwd))) {
+				snprintf_s_s(c->directory, sizeof(c->directory),
+						"%s/.opae", cwd);
+			} else {
+				// Current directory not found - use /
+				strncpy_s(c->directory, sizeof(c->directory),
+						"/.opae", 6);
```
zhangzhang10: Does a non-root user have write permission to "/" ?

tools/base/fpgad/fpgad.c
```diff
@@ -101,6 +109,103 @@ void sig_handler(int sig, siginfo_t *info, void *unused)
...
+{
+	char *sub;
+	bool def;
+	mode_t mode;
+	struct stat stat_buf;
+
+	if (!geteuid()) {
+		// If we're being run as root, then use DEFAULT_DIR_ROOT
+		// as the working directory.
+		strncpy_s(c->directory, sizeof(c->directory),
```
r-rojo: I wonder if this will be flagged by static analysis.

tools/base/fpgad/fpgad.c
tswhison: Probably not, however if the user doesn't have a ${HOME}, chances are it's a privileged / system user.  Worst case, the fopen call will fail.  Would it be better to default to /tmp I wonder?

tools/base/fpgad/fpgad.c
tswhison: How so?

tools/base/fpgad/fpgad.c
zhangzhang10: I guess no need to sweat on this corner case. It should be fine to default to "/".

PR #949, opened 2018-11-16
Added symlink tests to the existing tool tests

ibarraza:
Added tests to verify disallowing symlinks for fpgaconf, fpgad, and coreidle for user input file paths that the applications read from and write to.

Note: These tests will fail until symlink check capability is implemented for these tools.

### DIFF
--- a/testing/fpgad/test_fpgad_c.cpp
+++ b/testing/fpgad/test_fpgad_c.cpp
@@ -266,10 +266,6 @@ TEST_P(fpgad_fpgad_c_p, main_symlink_log) {
     ASSERT_EQ(mkdir(dir.c_str(), 0770), 0);
   }
 
-  if (stat(test_symlink.c_str(), &buffer) == 0) {
-    ASSERT_EQ(unlink(test_symlink.c_str()), 0);
-  }
-
   ASSERT_EQ(symlink("/tmp/fpgad.log", test_symlink.c_str()), 0);
   EXPECT_NE(fpgad_main(4, argv), 0);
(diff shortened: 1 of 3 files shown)


testing/fpgad/test_fpgad_c.cpp
```diff
@@ -191,5 +205,113 @@ TEST_P(fpgad_fpgad_c_p, main_invalid) {
...
+}
+
+/**
+ * @test       main_symlink_log
+ * @brief      Test: fpgad_main
+ * @details    When fpgad_main is called with valid params<br>
+ *             but the output log file is a symlink<br>
+ *             to a file, fpgad returns non-zero.<br>
+ */
+TEST_P(fpgad_fpgad_c_p, main_symlink_log) {
```
tswhison: So for the log file and the pid file, both are now interpreted to be relative to the working directory; so main won't actually flag these as a failure.  The new fn resolve_dirs will catch .. and / and reject based on that, but otherwise the path is formed by concatenating the working dir plus the parameter given for -l.  I think we don't care whether the resulting path is a symlink, unless that symlink points to something **not** in the working directory.

ibarraza: So the log and pid are set to be created in the current working directory. Ok, that's not what I thought your PR said. So to test this, it should create a symlink in the current working directory that is pointing to somewhere else, yes?

tswhison: Yes, that would be the correct check, which will currently fail.  @r-rojo you want to take this one while you're tweaking fpgad?

PR #951, opened 2018-11-19
Removed fpgabist compile flags definition in order to use top level compile flags.

ibarraza:
fpgabist's cmake file would overwrite the c flags set in the top level project. This change enables fpgabist to use the compile flags given at the top level of the project so we can get coverage data.


ibarraza:
We require coverage data from all of our tools and fpgabist was overwriting the compile flags that are required to get coverage data. Additionally, there did not appear to be any reason not to use the compile flags defined by the project at the top level in the first place. Was there a reason for requiring fpgabist to use only these specific compile options?

ibarraza:
Looking closer at it, the coverage depends on the afu given to the fpgabist app. So this change should not be required.

PR #952, opened 2018-11-19
cmake: Remove CXX_STANDARD in call to try_compile

r-rojo:
This can be removed as the default C++ flags in this build chain will
use -std=c++11


ibarraza:
This gets rid of the CXX_STANDARD error but it is still showing:
CMake Error at testing/CMakeLists.txt:669 (add_library):
  Cannot find source file:

    INTERFACE

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx

PR #953, opened 2018-11-19
Lzhan55/klocwork fix 1.2

lzhan55:


michael-adler:
Doesn't ASE_ERR force an exit? Do you need return statements after ASE_ERR?

lzhan55:
@michael-adler ASE_ERR doesn't force an exit. In ASE, there are three levels' logging: ASE_ERR, ASE_MSG and ASE_DEBUG. User can ask for more detailed information if needed. ASE_ERR will always be written in logs.

PR #956, opened 2018-11-19
Fixes for SDL

abelardojarab:
1. Fix mmio pkt structure if statement being flagged as not-properly indented.
2. Add stat code to check if directory is valid before opening.
3. Remove double 'import subprocess' on generate_environment.py and apply Python linter


lzhan55:
You need to fix the checks

PR #957, opened 2018-11-19
fpgaconf: use canonicalize_file_name

r-rojo:
Use canonicalize_file_name when validating the input gbs file.
Add/Update unit tests for this functionality.

### DIFF
--- a/tools/base/fpgaconf/fpgaconf.c
+++ b/tools/base/fpgaconf/fpgaconf.c
@@ -36,7 +36,7 @@
  *   * Auto-discovery of compatible slots for supplied bitstream
  *   * Dry-run mode ("what would happen if...?")
  */
-
+#define _GNU_SOURCE
 #include <errno.h>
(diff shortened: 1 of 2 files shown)


tools/base/fpgaconf/fpgaconf.c
```diff
@@ -372,9 +372,13 @@ int parse_args(int argc, char *argv[])
...
 		return -1;
 	}
-	config.filename = argv[optind];
-
-	return 0;
+	config.filename = canonicalize_file_name(argv[optind]);
+	if (config.filename) {
+		return 0;
+	} else {
+		fprintf(stderr, "Error locating GBS file specified: \"%s\"\n", strerror(errno));
```
zhangzhang10: Consider to use 'print_err' here?

r-rojo: `print_err` is specific to fpga_result codes.

PR #960, opened 2018-11-21
update version to 1.2.0-1

nakulkorde:


nakulkorde:
yea, I think it would make sense to add that to this PR. I will add it.

PR #963, opened 2018-11-26
Register reset in the PIM to help timing closure.

michael-adler:


michael-adler:
Tested with the usual DCP tests and on HW. Eliminated some timing closure failures.

platforms/platform_if/rtl/platform_shims/platform_shim_ccip_std_afu.sv
```diff
@@ -167,6 +167,13 @@ module platform_shim_ccip_std_afu
     //  Instantiate the AFU
     // ====================================================================
 
+    // Add a timing stage to reset
+    logic afu_cp2af_softReset_q = 1'b1;
+    always @(posedge `PLATFORM_PARAM_CCI_P_CLOCK)
```
lzhan55: @michael-adler I am trying to understand the relationship of register reset and timing closure. It seems like adding a delay of afu_cp2af_softReset will help the timing closure, am I right?

michael-adler: An extra register stage gives more flexibility for place and route.

PR #969, opened 2018-11-27
fpgad: expand $HOME directory before validating

r-rojo:
Use canonicalize_file_name to expand homedir.
If the directory does not exist, result will be null.
If it does exist, used the canonical name when checking if it's rooted
at /home.

### DIFF
--- a/tools/base/fpgad/fpgad.c
+++ b/tools/base/fpgad/fpgad.c
@@ -131,7 +131,7 @@ void resolve_dirs(struct config *c)
 
 		// Accept ${HOME} only if it is rooted at /home.
 		if (home) {
-			char *canon_path = canonicalize_file_name(home);
+			canon_path = canonicalize_file_name(home);
 			int len = strnlen_s(canon_path, PATH_MAX);
 			if (len >= 5)
 				strcmp_s(canon_path, 5, "/home/", &res);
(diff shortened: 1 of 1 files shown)


tools/base/fpgad/fpgad.c
```diff
@@ -126,18 +126,20 @@ void resolve_dirs(struct config *c)
 	} else {
 		// We're not root. Try to use ${HOME}/.opae
 		char *home = getenv("HOME");
+		char *canon_path = NULL;
 		int res = 1;
 
 		// Accept ${HOME} only if it is rooted at /home.
 		if (home) {
-			int len = strnlen_s(home, PATH_MAX);
+			char *canon_path = canonicalize_file_name(home);
```
tswhison: You've redefined canon_path here.  The outer one will always be NULL.

tools/base/fpgad/fpgad.c
```diff
@@ -126,18 +126,20 @@ void resolve_dirs(struct config *c)
...
 		int res = 1;
 
 		// Accept ${HOME} only if it is rooted at /home.
 		if (home) {
-			int len = strnlen_s(home, PATH_MAX);
+			canon_path = canonicalize_file_name(home);
+			int len = strnlen_s(canon_path, PATH_MAX);
 			if (len >= 5)
-				strcmp_s(home, 5, "/home/", &res);
+				strcmp_s(canon_path, 5, "/home/", &res);
```
zhangzhang10: dmax = 5 when passed to strcmp_s. So only up to first 5 chars are compared? Would this cause "/homex" to be compared as equal to "/home/"? I'm not sure if this is a valid concern.

tswhison: This is by design.  It is exploiting the behavior of strcmp_s, which will actually compare the trailing / character.  (ran it in the debugger to verify :))

PR #991, opened 2018-12-01
Include afu_platform_info in generated rpm

southerngs:
The afu_platform_info file is needed to generate AFUs.  It
is included when installing from source, but not included
in the generated rpm.


southerngs:
@michael-adler I don't know but there I think there should be.  Also I don't think that this file is used for `deb` package generation, so right now `deb` and `rpm` packages may not be the same.

PR #993, opened 2018-12-04
Tswhison/fpgad

tswhison:
Update fpgad to enumerate for supported devices, rather than assuming
that it is running on a dual-socket MCP. This required changing the
event registration mechanism to move away from using the sysfs file as
the device identifier. The fme/port device's object ID is used instead.
Update test cases to comply with the new design.


libopae/plugins/xfpga/event.c
```diff
@@ -468,32 +477,40 @@ STATIC fpga_result daemon_unregister_event(fpga_handle handle,
...
-	req.device[sizeof(req.device) - 1] = '\0';
+	result = fpgaDestroyProperties(&prop);
+	if (result != FPGA_OK) {
+		FPGA_ERR("failed to destroy properties");
+		goto out_close_conn;
+	}
+
+	req.type = UNREGISTER_EVENT;
+	req.event = event_type;
+	req.object_id = object_id;
```
r-rojo: I wonder if some of this can be refactored to one function to avoid some duplication...

tswhison: Did you mean to say similar to daemon_register_event?  Yeah, they are similar.  register checks for and opens the unix domain socket, whereas unregister assumes that it is already open.  I like having the register/unregister pairs, but if you feel strongly about it..

r-rojo: really just the routine to get the object id... I don't feel too strongly about it

PR #996, opened 2018-12-07
Check if the target path already exists in filecopy

deepakcu84:
Check if the target path already exists in filecopy

Fixes a corner case where the source folder contains
a qsys system and subdirectories contain TCL sources.
copytree recursively copies all dirs under the src.
Later, an attempt to copy subdirectories
with TCL sources fails because files already exist.

### DIFF
--- a/ase/scripts/generate_ase_environment.py
+++ b/ase/scripts/generate_ase_environment.py
@@ -389,8 +389,8 @@ def config_qsys_sources(filelist, vlog_srcs):
             try:
                 shutil.copytree(src_dir, tgt_dir, symlinks=False, ignore=None)
             except Exception as e:
-                errorExit("""Failed to copy tree {0} to
-                          {1}: Exception {2}""".
+                errorExit("Failed to copy tree {0} to "
+                          "{1}: Exception {2}".
                           format(src_dir, tgt_dir, e))
 
         # Point to the copy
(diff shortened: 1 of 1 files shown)


deepakcu84:
http://sw-pert.altera.com/pert/pert.php?test_run_id=2901472

michael-adler:
If it's TCL you are worried about, wouldn't it be sufficient to change the second copytree (the one at line 414) but leave the one at 390 alone? It seems safer to generate an error on a duplicate name at 390 unless you can think of a case where there could be a repeated reference there.

deepakcu84:
Yes, that would work. Typically TCL sources are always found in subdirs under qsys (not the other way around). I added checks everywhere to be consistent. I can remove the check for qsys.

ase/scripts/generate_ase_environment.py
```diff
@@ -387,10 +387,11 @@ def config_qsys_sources(filelist, vlog_srcs):
             print("Preparing {0}:".format(q))
             print("  Copying {0} to {1}...".format(src_dir, tgt_dir))
             try:
-                shutil.copytree(src_dir, tgt_dir)
-            except Exception:
-                errorExit("Failed to copy tree {0} to {1}".format(src_dir,
-                                                                  tgt_dir))
+                shutil.copytree(src_dir, tgt_dir, symlinks=False, ignore=None)
+            except Exception as e:
+                errorExit("""Failed to copy tree {0} to \
```
michael-adler: I don't think the format string here works the way you think it does. I think it will have whitespace before {1}. More likely, what you want is:

```
ErrorExit("Failed to copy tree {0} to"
          "{1}: Exception = {2}".format(src_dir, tgt_dir, e))
```

The Python limit of 80 columns would have been good style in 1980.

deepakcu84:
Yeah.. I'll fix it.

PR #999, opened 2018-12-11
Fix issues identified by checkers.

tswhison:


tools/base/fpgad/ap6.c
```diff
@@ -269,7 +269,7 @@ void *ap6_thread(void *thread_context)
 	int ret;
 	struct timespec ts = { .tv_sec = 0, .tv_nsec = 100000 }; /* 100ms */
 
-	fpga_token fme_token;
+	fpga_token fme_token = NULL;
```
r-rojo: can we use `nullptr` for consistency?

r-rojo: oh, yeah... 😜

PR #1004, opened 2018-12-13
Added tests for log levels and logging to file.

ibarraza:
Also fixed a bug where debug info could not be logged due to the variable only being set in xfpga's cmake instead of at the top level.


testing/opae-c/test_init_c.cpp
```diff
@@ -60,3 +60,66 @@ TEST(init, opae_init_rel) {
...
+ * @brief      When LIBOPAE_LOGFILE is specified, then the logger
+ *             will log to the specified file.
+ */
+TEST(init, log_file) {
+  ASSERT_EQ(0, putenv((char*)"LIBOPAE_LOGFILE=opae_log.log"));
+  opae_init();
+
+  struct stat buf;
+
+  EXPECT_EQ(0, stat("opae_log.log", &buf));
```
r-rojo: You probably want to make sure this file doesn't exist before the test runs.

ibarraza: Done. But I don't think we can remove it from inside the test once we're done.

ibarraza:
@r-rojo For the OPAE_DBG test it requires building with -DCMAKE_BUILD_TYPE=Debug. I suppose we can use some #ifdef to handle that case but I don't think we want to add this to travis CI since it would require another build step.

PR #1000, opened 2018-12-12
nullbyte tests for API tools (SDL)

jelonanX:
Also did some format cleanup in xfpga metrics tests

PR #1013, opened 2018-12-14
Added nullbyte tests for coreidle and fpgaconf

jelonanX:
Broke up commits from PR #1000

### DIFF
--- a/testing/coreidle/test_main_c.cpp
+++ b/testing/coreidle/test_main_c.cpp
@@ -109,8 +109,17 @@ class coreidle_main_c_p : public ::testing::TestWithParam<std::string> {
     }
   }
 
+  void copy_bitstream(std::string path) {
+    copy_gbs_ = path;
+    std::ifstream src(tmp_gbs_, std::ios::binary);
+    std::ofstream dst(copy_gbs_, std::ios::binary);
+  
+    dst << src.rdbuf();
+  }
+
   struct CoreIdleCommandLine cmd_line_;
   char tmp_gbs_[20];
+  std::string copy_gbs_;
   test_platform platform_;
   test_system *system_;
 };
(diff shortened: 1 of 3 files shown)


testing/coreidle/test_main_c.cpp
```diff
@@ -400,6 +409,49 @@ TEST_P(coreidle_main_c_p, parse_err2) {
...
+  copy_bitstream("copy_bitstream.gbs");
+  char zero[32];
+  char one[32];
+  char two[32];
+  char *argv[] = { zero, one, two };
+
+  strcpy(zero, "coreidle");
+  strcpy(one, "-G");
+  strcpy(two, "\0 copy_bitstream.gbs");
+  EXPECT_EQ(coreidle_main(3, argv), 0);
```
tswhison: I wouldn't expect this to return 0, because there is essentially no parameter to -G.  We should verify that the bitstream name is valid and exists, else return an error.

testing/coreidle/test_main_c.cpp
```diff
@@ -400,6 +409,49 @@ TEST_P(coreidle_main_c_p, parse_err2) {
...
+TEST_P(coreidle_main_c_p, mid_null_char) {
+  copy_bitstream("copy_bitstream.gbs");
+  char zero[32];
+  char one[32];
+  char two[32];
+  char *argv[] = { zero, one, two };
+
+  memset(two, 0, 32);
+  strcpy(two, "copy_bit\0stream.gbs");
+  EXPECT_EQ(coreidle_main(3, argv), 0);
```
tswhison: I wouldn't expect this to return 0, since the effective bitstream name "copy_bit" won't exist on the file system.

testing/fpgaconf/test_fpgaconf_c.cpp
```diff
@@ -679,6 +688,54 @@ TEST_P(fpgaconf_c_p, main2) {
...
+  strcpy(zero, "fpgaconf");
+  strcpy(one, "-v");
+  strcpy(two, "-n");
+  strcpy(three, "-B");
+  strcpy(four, "0x5e");
+  strcpy(five, "copy_bitstream\0.gbs");
+
+  char *argv[] = { zero, one, two, three, four,
+                   five };
+  EXPECT_NE(fpgaconf_main(6, argv), 0);
```
tswhison: Right 👍 , I would expect not-equal to 0, hopefully because the app realizes that "copy_bitstream" doesn't exist.

testing/fpgaconf/test_fpgaconf_c.cpp
```diff
@@ -679,6 +688,54 @@ TEST_P(fpgaconf_c_p, main2) {
...
+  strcpy(zero, "fpgaconf");
+  strcpy(one, "-v");
+  strcpy(two, "-n");
+  strcpy(three, "-B");
+  strcpy(four, "0x5e");
+  strcpy(five, "\0copy_bitstream.gbs");
+
+  char *argv[] = { zero, one, two, three, four,
+                   five };
+  EXPECT_NE(fpgaconf_main(6, argv), 0);
```
tswhison: Right, not equal to 0 because no GBS file was given.

PR #1014, opened 2018-12-14
xfpga: fix potential memleak in regex

r-rojo:
Change PARSE_MATCH_INT macro to use goto instead of returning to avoid
not freeing the regex structure in error cases.


libopae/plugins/xfpga/sysfs.c
```diff
@@ -108,11 +108,14 @@ STATIC int parse_pcie_info(sysfs_fpga_region *region, char *buffer)
 		FPGA_ERR("Error executing regex: %s", err);
 		res = FPGA_EXCEPTION;
```
tswhison: res is overwritten with FPGA_OK on line 116.  Probably need a goto out; here.

r-rojo: Oops. You're right

PR #1016, opened 2018-12-17
xfpga: add device_id/vendor_id and iter fn

r-rojo:
* add `parse_device_vendor_id` function that reads device/vendor sysfs
attributes
* change `make_region` to call `parse_device_vendor_id` using region
path as the root, and `device/device` and `device/vendor` to get
device_id and vendor_id.
* add `sysfs_foreach_region` iterator function to  process regions
* add sysfs tests

### DIFF
--- a/libopae/plugins/xfpga/sysfs.c
+++ b/libopae/plugins/xfpga/sysfs.c
@@ -306,13 +306,16 @@ STATIC int sysfs_region_destroy(sysfs_fpga_region *region)
 
 int sysfs_region_count(void)
 {
-	int res = 0;
+	int res = 0, count = 0;
 	if (!opae_mutex_lock(res, &_sysfs_region_lock)) {
-		res = _sysfs_region_count;
+		count = _sysfs_region_count;
 	}
 
-	opae_mutex_unlock(res, &_sysfs_region_lock);
-	return res;
+	if (opae_mutex_unlock(res, &_sysfs_region_lock)) {
+		count = 0;
+	}
+
+	return count;
 }
 
 void sysfs_foreach_region(region_cb cb, void *context)
--- a/libopae/plugins/xfpga/sysfs.c
+++ b/libopae/plugins/xfpga/sysfs.c
@@ -117,9 +117,66 @@ STATIC int parse_pcie_info(sysfs_fpga_region *region, char *buffer)
 	return res;
 }
 
+int syfs_parse_attribute64(const char *root, const char *attr_path, uint64_t *value)
+{
+	uint64_t pg_size = (uint64_t)sysconf(_SC_PAGE_SIZE);
+	char path[SYSFS_PATH_MAX];
+	char buffer[pg_size];
+	int fd = -1;
+	ssize_t bytes_read = 0;
+	int len = snprintf_s_ss(path, SYSFS_PATH_MAX, "%s/%s",
+				root, attr_path);
+	if (len < 0) {
+		FPGA_ERR("error concatenating strings (%s, %s)",
+			 root, attr_path);
+		return FPGA_EXCEPTION;
+	}
+	fd = open(path, O_RDONLY);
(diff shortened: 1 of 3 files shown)


libopae/plugins/xfpga/sysfs.c
```diff
@@ -236,6 +300,14 @@ int sysfs_region_count(void)
 	return _sysfs_region_count;
 }
 
+void sysfs_foreach_region(region_cb cb, void *context)
```
tswhison: Don't you need some locks around this, eg to protect _sysfs_region_count?

r-rojo: well, the idea is that the count is only set at initialization, after discovery. I suppose there is a risk of this being called while it is being initialized, in which case I would rather just return from the function instead of blocking. Maybe this function needs a return code?
Another scenario where a lock would help is if we do dynamic updates of region info (like a hotplug event) but we aren't there yet.
Thoughts?

tswhison: I agree that it's probably indicative of a poor design if we were to have thread contention here, but better safe than sorry. :)

libopae/plugins/xfpga/sysfs.c
```diff
@@ -301,7 +306,17 @@ STATIC int sysfs_region_destroy(sysfs_fpga_region *region)
 
 int sysfs_region_count(void)
 {
-	return _sysfs_region_count;
+	int res = 0;
+	if (pthread_mutex_lock(&_sysfs_region_lock)) {
```
tswhison: Have a look at opae_mutex_lock(res, mtxptr) to see how you like it.  It will save you some time.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -314,16 +329,33 @@ void sysfs_foreach_region(region_cb cb, void *context)
...
-	int res = 0;
+	int res = FPGA_OK;
 	uint32_t i = 0;
 	struct stat st;
 	DIR *dir = NULL;
 	char err[128] = {0};
 	struct dirent *dirent = NULL;
 	regex_t region_re;
 	regmatch_t matches[SYSFS_MAX_REGIONS];
+	if (pthread_mutexattr_init(&mattr)) {
```
tswhison: 343-357 isn't necessary, because you've already init'ed the mutex with PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP.  That's the beauty of the static initializer.  Also, this code is prone to race conditions itself (because the lock is not held.)  Better to remove it.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -396,8 +440,15 @@ int sysfs_initialize(void)
 		find_resources(&_regions[i]);
 	}
 
-
-	return FPGA_OK;
+out_unlock:
+	if (pthread_mutex_unlock(&_sysfs_region_lock)) {
+		FPGA_MSG("error unlocking sysfs region mutex");
```
tswhison: regfree(&region_re); here before returning?

libopae/plugins/xfpga/sysfs.c
```diff
@@ -307,29 +307,29 @@ STATIC int sysfs_region_destroy(sysfs_fpga_region *region)
 int sysfs_region_count(void)
 {
 	int res = 0;
-	if (pthread_mutex_lock(&_sysfs_region_lock)) {
-		FPGA_MSG("Error locking sysfs region lock");
-	} else {
+	if (!opae_mutex_lock(res, &_sysfs_region_lock)) {
 		res = _sysfs_region_count;
```
tswhison: You'll need a separate variable to capture the count, because res will be reset on 314.

r-rojo: Just testing you

PR #1020, opened 2018-12-17
Wait for CCI-P requests to complete before releasing NLB buffers.

michael-adler:
The NLB HW writes to the DSM at the timeout point. There may still
be memory read requests in flight at this point. Update the SW to
probe a CSR that indicates when all CCI-P traffic is complete
before releasing buffers.


michael-adler:
Tested both with ASE and on HW with NLB0 and NLB3.

Fixed ASE support in the process by automatically skipping the perf counter lookup.

This fixes references to unpinned addresses at the end of ASE runs, which were caused by references to buffers that had been deallocated. The behavior is potentially illegal on HW too, since it could result in references to unpinned memory.

michael-adler:
Someone should update the Python version to match. The STATUS1 register is the same on all instances of NLB and could be used in common code. The CPP isn't structured that way, so I replicated it in both nlb0 and nlb3.

michael-adler:
What's up with the error in the coverage tests? I doubt this check-in would cause that.

lzhan55:
@ibarraza Could you re-run the spelling check? It seems like an environment issue.

PR #1030, opened 2018-12-19
Added more tests for tools

jelonanX:
Added new symlink and path validate tests for tools fpgaconf and coreidle.


testing/fpgaconf/test_fpgaconf_c.cpp
```diff
@@ -707,16 +843,34 @@ TEST_P(fpgaconf_c_p, embed_nullchar1) {
...
   strcpy(three, "-B");
   strcpy(four, "0x5e");
-  strcpy(five, "copy_bitstream\0.gbs");
 
   char *argv[] = { zero, one, two, three, four,
                    five };
-  EXPECT_NE(fpgaconf_main(6, argv), 0);
-  unlink(copy_gbs_.c_str());
+
+  if (!symlink(copy_gbs_.c_str(), symlink_gbs.c_str())) {
```
tswhison: Do we want to fail the test if this fails?  Not sure..

testing/fpgaconf/test_fpgaconf_c.cpp
```diff
@@ -728,12 +882,32 @@ TEST_P(fpgaconf_c_p, embed_nullchar2) {
...
   strcpy(four, "0x5e");
-  strcpy(five, "\0copy_bitstream.gbs");
 
   char *argv[] = { zero, one, two, three, four,
                    five };
-  EXPECT_NE(fpgaconf_main(6, argv), 0);
-  unlink(copy_gbs_.c_str());
+
+  // Create link directories
+  if (!(mkdir("./link1", S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH))
```
tswhison: Fail if the mkdir or symlink calls fail?

testing/fpgaconf/test_fpgaconf_c.cpp
```diff
@@ -750,8 +769,8 @@ TEST_P(fpgaconf_c_p, encoding_path) {
  * @test       relative_path
  * @brief      Test: fpgaconf_main
  * @details    When gbs file locates in parent directory and command params<br>
- *             contains path traversal, fpgaconf_main displays file<br>
- *             error and returns non-zero.<br>
+ *             contains path traversal. On success, fpgaconf_main loads bitstream<br>
+ *             and returns zero. Otherwise, it displays file error and returns non-zero.<br>
  */
 TEST_P(fpgaconf_c_p, relative_path) {
```
tswhison: I thought we would expect this one to fail?

testing/fpgaconf/test_fpgaconf_c.cpp
jelonanX: I believe so. But I have yet figure out how to get mkdir and symlink to fail. Any suggestions?

testing/fpgaconf/test_fpgaconf_c.cpp
jelonanX: Previously, it was a false positive on set_userclk. Setup has been modified to load in a valid bitstream so this test succeeds now.

tswhison: Ah, cool.

PR #1033, opened 2018-12-20
xfpga: add opae_ioctl files

r-rojo:
opae_ioctl will contain the ioctl interface methods and defines a common
way of collaing ioctl.
NOTE: While the common ioctl function is variadic, it only unpacks one
argument. IOW, will only work for IOCTL requests that take one argument.

Also, this PR defines one ioctl operation (opae_port_map). 
Other functions will have to be defined and implemented.

### DIFF
--- a/libopae/plugins/xfpga/buffer.c
+++ b/libopae/plugins/xfpga/buffer.c
@@ -31,7 +31,8 @@
 #include "opae/access.h"
 #include "opae/utils.h"
 #include "common_int.h"
-#include "intel-fpga.h"
+
+#include "opae_ioctl.h"
 
 #include <sys/types.h>
 #include <sys/stat.h>
@@ -150,6 +151,7 @@ fpga_result __FPGA_API__ xfpga_fpgaPrepareBuffer(fpga_handle handle, uint64_t le
 {
 	void *addr = NULL;
 	fpga_result result = FPGA_OK;
+	uint64_t io_addr = 0;
 	struct _fpga_handle *_handle = (struct _fpga_handle *) handle;
 	int err;
 
@@ -229,15 +231,7 @@ fpga_result __FPGA_API__ xfpga_fpgaPrepareBuffer(fpga_handle handle, uint64_t le
 		}
 	}
 
-	/* Set ioctl fpga_port_dma_map struct parameters */
-	struct fpga_port_dma_map dma_map = {.argsz = sizeof(dma_map),
-					    .flags = 0,
-					    .user_addr = (__u64) addr,
-					    .length = (__u64) len,
-					    .iova = 0};
-
-	/* Dispatch ioctl command */
-	if (ioctl(_handle->fddev, FPGA_PORT_DMA_MAP, &dma_map) != 0) {
+	if (opae_port_map(_handle->fddev, addr, len, &io_addr)) {
 		if (!preallocated) {
 			buffer_release(addr, len);
 		}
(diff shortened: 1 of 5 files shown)


libopae/plugins/xfpga/buffer.c
```diff
@@ -229,15 +231,7 @@ fpga_result __FPGA_API__ xfpga_fpgaPrepareBuffer(fpga_handle handle, uint64_t le
...
-	/* Set ioctl fpga_port_dma_map struct parameters */
-	struct fpga_port_dma_map dma_map = {.argsz = sizeof(dma_map),
-					    .flags = 0,
-					    .user_addr = (__u64) addr,
-					    .length = (__u64) len,
-					    .iova = 0};
-
-	/* Dispatch ioctl command */
-	if (ioctl(_handle->fddev, FPGA_PORT_DMA_MAP, &dma_map) != 0) {
+	if (opae_port_map(_handle->fddev, addr, len, &io_addr)) {
```
anandaravuri: Don't use wrapper for IOCTL ,  too many wrappers causes code unmaintainable

store driver version enum fpga_token  
enum {
fpga_driver_devp_lateset_version,
fpga_driver_ups_version,
}
if( token -> drv_ver == fpga_driver_devp_lateset_version )
{ 
  ioctl()

}else  {
 ioctl()
}

libopae/plugins/xfpga/opae_ioctl.c
```diff
@@ -0,0 +1,99 @@
...
+
+fpga_result opae_ioctl(int fd, int request, ...)
+{
+	fpga_result res = FPGA_OK;
+	va_list argp;
+	va_start(argp, request);
+	void *msg = va_arg(argp, void *);
+	errno = 0;
+	if (ioctl(fd, request, msg) != 0) {
+		FPGA_MSG("error executing ioctl: %s", strerror(errno));
```
tswhison: Could we use OPAE_MSG, etc, for new code?  I'd like to eventually deprecate the FPGA_ version.

libopae/plugins/xfpga/opae_ioctl.c
```diff
@@ -0,0 +1,99 @@
...
+	return res;
+}
+
+int opae_port_map(int fd, void *addr, uint64_t len, uint64_t *io_addr)
+{
+	int res = 0;
+	int req = 0;
+	void *msg = NULL;
+	/* Set ioctl fpga_port_dma_map struct parameters */
+	struct fpga_port_dma_map dma_map = {.argsz = sizeof(dma_map),
```
tswhison: I think this only supports the original IOCTL interface currently.  You will add upstreamed support in a later pull?

libopae/plugins/xfpga/buffer.c
tswhison: @anandaravuri I see it the other way around.  If the code is sprinkled with this sort of pattern..
```c
if (<this version>) {
  do_this();
} else {
  do_that();
}
```
Then the maintenance headache is increased, especially for handling new versions.  With the abstraction, the code that needs to change can be co-located, which reduces the maintenance overhead.

r-rojo: Yes, I agree with @tswhison . Putting ioctl calls in one file greatly increases maintainability and readability. The approach you're proposing, @anandaravuri , tends to increase function size and code duplication.

libopae/plugins/xfpga/opae_ioctl.c
r-rojo: Yes, that is my plan. Refactor ioctl calls with current driver and gain confidence that it works.
Next, add support for upstreamed driver.

PR #1034, opened 2018-12-20
prog_guide: update Functions table for Umsg

tswhison:
Update the table to reflect that the Umsg API requires a handle of type
FPGA_ACCELERATOR.


testing/opae-c/test_umsg_c.cpp
```diff
@@ -132,7 +132,7 @@ class umsg_c_p : public ::testing::TestWithParam<std::string> {
     filter_ = nullptr;
     ASSERT_EQ(fpgaInitialize(NULL), FPGA_OK);
     ASSERT_EQ(fpgaGetProperties(nullptr, &filter_), FPGA_OK);
-    ASSERT_EQ(fpgaPropertiesSetObjectType(filter_, FPGA_DEVICE), FPGA_OK);
```
r-rojo: Should we have a negative test for this?
Say, when you call this with `FPGA_DEVICE`, it fails?

tswhison: Not a bad idea, but I think that such a test will pass with the current mock.

PR #1037, opened 2018-12-20
Resolve in xfpga-buffer leaks

jelonanX:


testing/xfpga/test_buffer_c.cpp
```diff
@@ -68,7 +68,7 @@ extern "C" {
 #define FPGA_MSG(fmt, ...) \
 	printf("MOCK " fmt "\n", ## __VA_ARGS__)
 
-
+#pragma pack(1)
```
r-rojo: Wouldn't this set this alignment for the rest of the structs in the file?
When setting alignment on structs, it's best to do it explicitly/individually for each struct you want aligned a certain way.
The better approach would be:
```C
#pragma pack(push, 1)
 struct buffer_params {
   fpga_result result;	   fpga_result result;
   size_t size;	   size_t size;
   int flags;	   int flags;
 };
#pragma pack(pop)
```

jelonanX: Learning_list.append("#pragma pack(push, 1)
 struct buffer_params {
   fpga_result result;	   fpga_result result;
   size_t size;	   size_t size;
   int flags;	   int flags;
 };
#pragma pack(pop)")

testing/xfpga/test_buffer_c.cpp
```diff
@@ -68,12 +68,13 @@ extern "C" {
 #define FPGA_MSG(fmt, ...) \
 	printf("MOCK " fmt "\n", ## __VA_ARGS__)
 
-#pragma pack(1)
+#pragma pack(push, 1)
```
r-rojo: 👍

testing/xfpga/test_buffer_c.cpp
```diff
@@ -210,23 +214,23 @@ TEST_P(buffer_prepare, PrepPre2MB01) {
   EXPECT_EQ(FPGA_OK, xfpga_fpgaReleaseBuffer(handle_, wsid));
 
   // buf_addr was preallocated, do not touch it
-  ASSERT_NE(buf_addr, (void*)NULL);
+  ASSERT_NE(buf_addr, (void*)nullptr);
```
r-rojo: I don't think you need the
```C
(void*)
```

PR #1040, opened 2018-12-20
Fixed possible leak in error tests.

ibarraza:


testing/xfpga/test_error_c.cpp
```diff
@@ -87,6 +87,7 @@ class error_c_mock_p
       filter_ = nullptr;
     }
     token_cleanup();
+    tmpsysfs_ = "";
```
tswhison: Interesting.. the error goes away with this change?

ibarraza: Yes. Apparently it doesn't think it gets cleaned up unless we do this.

tswhison: Interesting.  Well, these test fixtures are in global scope, so I suppose that's what the checker is complaining about.

PR #1042, opened 2018-12-21
xfpga: add functions for getting info

r-rojo:
add the folloing functions:
* opae_get_fme_info
* opae_get_port_info
* opae_get_port_region_info

Change code that uses the ioctl interface directly to instead use these
functions.

Because these functions call one common ioctl wrapper that returns error
codes according to errno, some tests had to be updated for to expect the
appropriate error codes.

### DIFF
--- a/libopae/plugins/xfpga/event.c
+++ b/libopae/plugins/xfpga/event.c
@@ -42,6 +42,7 @@
 #include <opae/properties.h>
 #include "xfpga.h"
 #include "common_int.h"
+#include "opae_ioctl.h"
 #include "types_int.h"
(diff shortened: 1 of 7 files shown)


libopae/plugins/xfpga/event.c
```diff
@@ -124,36 +126,35 @@ STATIC fpga_result send_fme_event_request(fpga_handle handle,
 		if (ioctl(_handle->fddev, FPGA_FME_ERR_SET_IRQ, &fme_irq)
```
tswhison: You're going to eventually abstract this one, too?

PR #1044, opened 2018-12-21
Adjust fpgaconf tests for HW

jelonanX:


ibarraza:
These are the same tests, just now they also pass when running on HW. fpgaconf tried to reconfigure the bitstream created in SetUp() before, which only contained metadata and so it would fail. So we could pull out those tests into a mock only class or set fpgaconf to dry run for those tests.

PR #1048, opened 2018-12-22
xfpga: add ioctl wrappers for irq operations

r-rojo:


libopae/plugins/xfpga/opae_ioctl.c
```diff
@@ -233,7 +233,7 @@ int opae_port_set_user_irq(int fd, uint32_t flags, uint32_t start, uint32_t coun
 	irq->flags = 0;
 	irq->start = start;
 	irq->count = count;
-	err = memcpy32_s((uint32_t*)irq->evtfd, count, (uint32_t*)eventfd, count);
+	err = memcpy32_s((uint32_t *)irq->evtfd, count, (uint32_t *)eventfd, count);
```
tswhison: Just curious - why the 32-bit version rather than the generic version?

r-rojo: The generic one specifies the `count` in bytes whereas this one the `count` is the number of 32-bit words which happens to be the size of each element in the eventfd array.

PR #1049, opened 2018-12-28
tools: fix  reconf interface id

anandaravuri:


tools/base/fpgaconf/bitstream-tools.h
```diff
@@ -168,58 +169,65 @@ fpga_result check_bitstream_guid(const uint8_t *bitstream)
 	return FPGA_OK;
 }
 
-fpga_result get_fpga_interface_id(fpga_token token, uint64_t *id_l,
-								uint64_t *id_h)
+int64_t bs_int64_be_to_le(int64_t val)
```
zhangzhang10: Consider using compiler builtin function for byte swapping?

r-rojo: http://man7.org/linux/man-pages/man3/bswap.3.html
```
       #include <byteswap.h>

       bswap_16(x);
       bswap_32(x);
       bswap_64(x);
```

anandaravuri: fixed with bswap_64(x)

tools/base/fpgaconf/bitstream-tools.h
```diff
@@ -168,58 +171,55 @@ fpga_result check_bitstream_guid(const uint8_t *bitstream)
...
+		guid, sizeof(uint64_t));
 	if (EOK != e) {
-		OPAE_ERR("strncpy_s failed (buf_l)\n");
-		return FPGA_EXCEPTION;
+		OPAE_ERR("memcpy_s failed");
+		goto out_destroy;
 	}
 
-	*id_h = strtoull(buf_h, NULL, 16);
+	*id_h = bswap_64(*id_h);
```
tswhison: Wondering if the byte swapping should be done here or internally by the properties layer?  Or is this for compatibility with the bitstream metadata format?

anandaravuri: OPAE properties layer doesn't swap bytes

tswhison: Understood.. I'm wondering whether it should be done at the properties level?

anandaravuri: fixed in tools

tools/base/fpgaconf/bitstream-tools.h
```diff
@@ -168,61 +151,45 @@ fpga_result check_bitstream_guid(const uint8_t *bitstream)
...
 	if (result != FPGA_OK) {
-		OPAE_ERR("Failed to get Token Object \n");
-		return result;
+		OPAE_ERR("Failed to get Token Properties Object \n");
+		goto out;
 	}
-
-	result = fpgaObjectRead(pr_object, buffer, 0,
-		INTFC_ID_LOW_LEN + INTFC_ID_HIGH_LEN, 0);
+	result = fpgaPropertiesGetGUID(filter, &guid);
```
tswhison: Do we want to add a safeguard so that this function fails if the device type is not FPGA_DEVICE?

anandaravuri: fixed

tools/base/fpgaconf/bitstream-tools.h
```diff
@@ -165,6 +166,18 @@ fpga_result get_fpga_interface_id(fpga_token token, fpga_guid interface_id)
...
 	}
+
+	result = fpgaPropertiesGetObjectType(filter, &objtype);
+	if (result != FPGA_OK) {
+		OPAE_ERR("Failed to get Token Properties Object \n");
+		goto out_destroy;
+	}
+
+	if (objtype != FPGA_DEVICE) {
+		OPAE_ERR("Invalid FPGA object type \n");
```
tswhison: Thanks, Ananda.  Probably want to set result to an error code here, too. FPGA_EXCEPTION maybe?

anandaravuri: fixed

PR #1054, opened 2019-01-04
Resolve reconf leak

jelonanX:
Separate out positive test case


jelonanX:
It's leaking in test_system.cpp with an invalid open. My guess is that we are storing fds_ even if fd is -1. We only clear fds_ in teardown and not between assertions . 
https://tc03p-or.devtools.intel.com/repository/download/Aalsdk_DeveloperWorkspace_ItbarrazWorkspace_ValgrindScanOpaeSdkXWeekly/1083889:id/valgrind_reports.zip%21/valgrind_reports/valgrind_xfpga_reconf_c.txt

PR #1056, opened 2019-01-05
xfpga: add dfl ioctl calls

r-rojo:
* Change return type of ioctl functions to fpga_result
* Add a struct that contains function pointers to opae ioctl ops
* Add an array of two elements of this new data structure type
  * The first element refers to fn ptrs related to dfl (upstream) driver
  * The second element refers to fn ptrs for internal (intel) driver
* Add a static pointer to an entry in this table
  * Add initialization routine for resolving this ptr
  * Call opae_ioctl_initialize in plugin initialization

TODO: Look into combining sysfs format operations and ioctl operations
in one interface (perhaps opae_kernel.h|c).

### DIFF
--- a/libopae/plugins/xfpga/opae_ioctl.c
+++ b/libopae/plugins/xfpga/opae_ioctl.c
@@ -476,11 +476,11 @@ static ioctl_ops ioctl_table[MAX_KERNEL_DRIVERS] = {
 	 .fme_port_assign = intel_fme_port_assign,
 	 .fme_port_release = intel_fme_port_release,
 	 .fme_port_pr = intel_fme_port_pr,
-	 .fme_port_reset = intel_fme_port_reset}};
+	 .fme_port_reset = intel_fme_port_reset} };
 
 static ioctl_ops *io_ptr;
 
-int opae_ioctl_initialize()
+int opae_ioctl_initialize(void)
 {
 	struct stat st;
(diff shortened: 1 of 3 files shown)


libopae/plugins/xfpga/opae_ioctl.c
```diff
@@ -286,7 +331,262 @@ int opae_fme_port_pr(int fd, uint32_t flags, uint32_t port_id, uint32_t sz,
...
+	return res;
+}
+
+fpga_result dfl_fme_port_reset(int fd)
+{
+	return opae_ioctl(fd, DFL_FPGA_PORT_RESET, NULL);
+}
+
+#define MAX_KERNEL_DRIVERS 2
+static ioctl_ops ioctl_table[MAX_KERNEL_DRIVERS] = {
```
tswhison: More precisely, ioctl_table[0] is for patch set 0 of the up-streamed driver, and we anticipate extending the table for future upstreamed patchsets?

r-rojo: I think so but only if the changes to future updates are not backwards compatible.
If future patch sets contain new ioctl request codes, the header will be updated and the table will include function pointers to the new codes. If these functions are used on an older patch set, like patch set 0, with the newer codes, we should get an error code that translates to FPGA_NOT_SUPPORTED.

PR #1059, opened 2019-01-07
tests: add support to upstream drviver mock and hw platform

anandaravuri:


testing/platform/fpga_hw.cpp
```diff
@@ -425,9 +451,10 @@ test_device make_device(uint16_t ven_id, uint16_t dev_id, const std::string &pla
 fpga_driver get_driver(const std::string &path)
 {
   char buffer[PATH_MAX];
```
ibarraza: In my tests, this buffer had to be zeroed out first.

anandaravuri: fixed

PR #1060, opened 2019-01-07
Cherry pick: Fixed test platform not removing invalid platforms when running on hw (#1053)

ibarraza:


testing/platform/fpga_hw.cpp
```diff
@@ -202,10 +202,11 @@ std::vector<std::string> test_platform::platforms(
   }
   // from the list of platform names requested, remove the ones not found in
   // the platform db
-  std::remove_if(keys.begin(), keys.end(), [drv](const std::string &n) {
-    auto db = fpga_db::instance();
-    return !db->exists(n) || drv != db->get(n).driver;
```
ibarraza: @r-rojo This change does not work with your change here "|| drv != db->get(n).driver". Can you provide some input here?

PR #1061, opened 2019-01-08
tests: add support for upstream driver mock sysfs

anandaravuri:


testing/opae-c/test_enum_c.cpp
```diff
@@ -670,3 +620,65 @@ TEST_P(enum_c_mock_p, clone_token02) {
...
+		FPGA_OK);
+	EXPECT_EQ(num_matches_, 0);
+}
+
+INSTANTIATE_TEST_CASE_P(enum_c, enum_c_err_p,
+	::testing::ValuesIn(test_platform::platforms({ "skx-p","dcp-rc" })));
+
+class enum_c_socket_p : public enum_c_p {};
+
+TEST_P(enum_c_socket_p, socket_id) {
```
ibarraza: This test class is inheriting the exact same class and is targeting the same platforms as the enum_c_err_p test class. Should probably just have both tests use the same class and name them something generic like enum_c_skx_dcp_p or something.

anandaravuri: test case should be generic , not having any name like skx, dcp etc

PR #1064, opened 2019-01-08
Fixed test platforms removal logic

ibarraza:


ibarraza:
@r-rojo Not yet I think. Still a bit of logic to fix in test_platform::platforms() first.

testing/opae-cxx/test_properties_cxx_core.cpp
```diff
@@ -286,4 +286,4 @@ TEST_P(properties_skx_dcp_cxx_core, get_guid_valid) {
 }
 
 INSTANTIATE_TEST_CASE_P(properties_skx_dcp, properties_skx_dcp_cxx_core,
```
tswhison: Since you're calling out the Intel driver here, should the fixture name reflect that as well?

ibarraza:
The get_guid test that was failing for cxx properties tests was failing correctly. Will leave that test and just use this PR for the platforms removal logic bit.

testing/platform/fpga_hw.cpp
```diff
@@ -205,7 +205,9 @@ std::vector<std::string> test_platform::platforms(
   keys.erase(
     std::remove_if(keys.begin(), keys.end(), [drv](const std::string &n) {
       auto db = fpga_db::instance();
-      return !db->exists(n);
+      return !db->exists(n) || (drv != fpga_driver::linux_any
+                                && drv != db->get(n).driver)
+                            || db->get(n).devices.empty();
```
tswhison: devices can be empty?

ibarraza: Devices *shouldn't* be empty and the db should only be populated with the detected platform when running on hardware. However, due to a bug whose source has yet to be identified, when we run on a system that has only the dfl platform, skx-p and dcp-rc also are in the db although they have no devices. This change is a workaround for now so that we can continue verifying the rest of the tests on hardware. This should only affect tests that are specific to certain platforms and only when running on hardware and with this workaround, even that case will work.

PR #1065, opened 2019-01-08
Separate enum tests to skx and dcp ONLY

jelonanX:


libopae/plugins/xfpga/enum.c
```diff
@@ -386,7 +386,7 @@ STATIC fpga_result enum_fme(const char *sysfspath, const char *name,
 	// Discover the socket id from the FME's sysfs entry.
 	resval = sysfs_parse_attribute64(sysfspath, FPGA_SYSFS_SOCKET_ID, &value);
 	if (resval != 0) {
-		return FPGA_NOT_FOUND;
+		//return FPGA_NOT_FOUND;
```
tswhison: Can we consume the fix made by @anandaravuri for this?  Not sure this should be checked in with commented-out error reporting.

jelonanX: Yup. This is resolved in the merge conflict.

testing/xfpga/test_enum_c.cpp
```diff
@@ -669,7 +626,7 @@ TEST_P(enum_c_p, clone_token) {
       FPGA_OK);
   EXPECT_GT(num_matches_, 0);
   fpga_token src = tokens_[0];
-  fpga_token dst;
+  fpga_token ds
```
tswhison: whoops, need a semicolon after ds

PR #1080, opened 2019-01-11
Compatibility with arch "aarch64"

JensHuthmann:
These changes allow the SDK to be compiled on arch "aarch64" (arm 64)


tools/extra/c++utils/option_parser.cpp
```diff
@@ -172,7 +172,7 @@ bool option_parser::parse_args(int argc, char* argv[], option_map & known_option
     optarray.push_back({nullptr,0,nullptr,0});
 
     int longidx;
-    char c;
+    int c;
```
JensHuthmann: getopt returns an integer

on aarch64 comparision of char c with -1 is throwing a compile error

JensHuthmann:
> 
> 
> Have you run any tests after compilation on aarch64?

We did only run one application up until it tries to communicate with the accelerator card. We did not yet move the card into the arm system. If you need this for approval we try to do it ASAP.

lzhan55:
ASE related changes are good.

PR #1086, opened 2019-01-14
Move event test to mock only

jelonanX:
Both send_port_event_request and send_fme_event_request return FPGA_NOT_SUPPORTED for mock. 
When on hw, it returns FPGA_INVALID_PARAM. This pr moves tests to mock only.


jelonanX:
When running on a machine with dcp, this particular test also returns FPGA_INVALID_PARAM. It seems the return behavior are not hw platform specific. Could it be a bug?

PR #1091, opened 2019-01-15
Resolved xfpga event tests

jelonanX:
Follow up PR #1086.

Separated send_fme_event_request and send_port_event_request tests to mcp and dcp platforms only. It will return FPGA_NOT_SUPPORTED for all mcp platforms. FPGA_OK is returned for dcp on hw only. (FPGA_NOT_SUPPORTED is returned for mock. Maybe this should be a FIX_ME?)

### DIFF
--- a/testing/xfpga/test_events_c.cpp
+++ b/testing/xfpga/test_events_c.cpp
@@ -354,8 +354,8 @@ class events_p : public ::testing::TestWithParam<std::string> {
     EXPECT_EQ(fpgaDestroyProperties(&filter_dev_), FPGA_OK);
     EXPECT_EQ(fpgaDestroyProperties(&filter_accel_), FPGA_OK);
 
-    if (handle_dev_) { EXPECT_EQ(xfpga_fpgaClose(handle_dev_), FPGA_OK); }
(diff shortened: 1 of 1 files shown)


testing/xfpga/test_events_c.cpp
```diff
@@ -777,7 +846,7 @@ TEST_P(events_mock_p, valid_fme_event_request){
 
   gEnableIRQ = true;
   system_->register_ioctl_handler(FPGA_FME_GET_INFO, fme_info);
-  auto res = send_fme_event_request(handle_dev_,eh_,fme_op);
+  auto res = send_fme_event_request(handle_accel_,eh_,fme_op);
```
tswhison: I think this one should be handle_dev, since we're making an fme event request.

jelonanX: good catch.

testing/xfpga/test_events_c.cpp
```diff
@@ -744,6 +735,84 @@ INSTANTIATE_TEST_CASE_P(events, events_p,
                         ::testing::ValuesIn(test_platform::platforms({ "skx-p","dcp-rc" })));
 
 
+class events_dcp_p : public events_p {};
```
r-rojo: Can we use `pac` instead of `dcp`?

PR #1094, opened 2019-01-15
xfpga: fix sysfs region enum error msg

anandaravuri:


tswhison:
Seems like we should print the error when using the "Intel" driver, but suppress it when using the "dfl" driver.

PR #1098, opened 2019-01-17
tests: fix xfpga testcases for release build

anandaravuri:


testing/xfpga/test_events_c.cpp
```diff
@@ -1377,7 +1379,7 @@ class events_handle_p : public ::testing::TestWithParam<std::string> {
     system_->initialize();
     system_->prepare_syfs(platform_);
 
-    ASSERT_EQ(FPGA_OK, fpgaInitialize(NULL));
+    ASSERT_EQ(FPGA_OK, xfpga_plugin_finalize());
```
ibarraza: I think this is supposed to be xfpga_plugin_initialize().

testing/xfpga/test_sysfs_c.cpp
```diff
@@ -325,7 +327,7 @@ class sysfs_c_p : public ::testing::TestWithParam<std::string> {
           t = nullptr;
       }
     }
-    fpgaFinalize();
+    xfpga_plugin_initialize();
```
ibarraza: I think this is supposed to be xfpga_plugin_finalize().

PR #1102, opened 2019-01-18
xfgpa: reset sysfs region ptr to NULL in finalize

r-rojo:


libopae/plugins/xfpga/sysfs.c
```diff
@@ -487,6 +487,7 @@ int sysfs_finalize(void)
 		sysfs_region_destroy(&_regions[i]);
 	}
 	_sysfs_region_count = 0;
+	_sysfs_format_ptr = NULL;
```
tswhison: Are these protected by any kind of lock?

r-rojo: oh, no.. 😞

tswhison: Could you address in another pull? :)

PR #1106, opened 2019-01-22
tools: ras clear all fme errors

anandaravuri:


tools/extra/ras/main.c
```diff
@@ -909,14 +909,50 @@ fpga_result inject_ras_errors(fpga_handle fme_handle,
 	return result;
 }
 
+fpga_result clear_fpga_errors(fpga_handle handle, char *error )
```
r-rojo: You might want to make this STATIC, and the error a `const char *` type
```
STATIC fpga_result clear_fpga_errors(fpga_handle handle, const char *error )
```

tools/extra/ras/main.c
```diff
@@ -909,14 +909,50 @@ fpga_result inject_ras_errors(fpga_handle fme_handle,
...
+	result = fpgaHandleGetObject(handle, error, &error_object, 0);
+	if (result != FPGA_OK) {
+		OPAE_ERR("Failed to get handle Object");
+		return result;
+	}
 
+	result = fpgaObjectRead64(error_object, &value, 0);
+	if (result != FPGA_OK) {
+		OPAE_ERR("Failed to Read Object ");
+		return result;
```
r-rojo: memory leak here, we're not destroying `error_object`

tools/extra/ras/main.c
```diff
@@ -909,14 +909,50 @@ fpga_result inject_ras_errors(fpga_handle fme_handle,
...
+	}
+
+	result = fpgaObjectWrite64(error_object, value, 0);
+	if (result != FPGA_OK) {
+		OPAE_ERR("Failed to Write Object ");
+		return result;
```
r-rojo: memory leak here, we're not destroying `error_object`

tools/extra/ras/main.c
anandaravuri: fixed

tools/extra/ras/main.c
anandaravuri: fixxed

tools/extra/ras/main.c
anandaravuri: fixed

tools/extra/ras/main.c
```diff
@@ -834,21 +834,21 @@ fpga_result clear_port_errors(fpga_handle afu_handle)
 	result = fpgaObjectRead64(port_error_object, &value, 0);
 	if (result != FPGA_OK) {
 		OPAE_ERR("Failed to Read Object ");
-		return result;
+		goto out;
```
r-rojo: You lose the value of `result` when you `goto out`

tools/extra/ras/main.c
```diff
@@ -834,21 +834,21 @@ fpga_result clear_port_errors(fpga_handle afu_handle)
...
 	}
 
 	printf("\n \n Port error CSR : 0x%lx \n", value);
 
 	result = fpgaObjectWrite64(port_error_object, 0x0, 0);
 	if (result != FPGA_OK) {
 		OPAE_ERR("Failed to Read Object ");
-		return result;
+		goto out;
```
r-rojo: You lose the value of `result` when you `goto out`

tools/extra/ras/main.c
```diff
@@ -874,7 +874,7 @@ fpga_result inject_ras_errors(fpga_handle fme_handle,
 	result = fpgaObjectRead64(inj_error_object, &inj_error.csr, 0);
 	if (result != FPGA_OK) {
 		OPAE_ERR("Failed to Read Object ");
-		return result;
+		goto out;
```
r-rojo: You lose the value of `result` when you `goto out`

tools/extra/ras/main.c
```diff
@@ -896,27 +896,63 @@ fpga_result inject_ras_errors(fpga_handle fme_handle,
 	result = fpgaObjectWrite64(inj_error_object, inj_error.csr, 0);
 	if (result != FPGA_OK) {
 		OPAE_ERR("Failed to Read Object ");
-		return result;
+		goto out;
```
r-rojo: Not needed here. We can just fall through to out.

tools/extra/ras/main.c
```diff
@@ -896,27 +896,63 @@ fpga_result inject_ras_errors(fpga_handle fme_handle,
...
+	result = fpgaHandleGetObject(handle, error, &error_object, 0);
+	if (result != FPGA_OK) {
+		OPAE_ERR("Failed to get handle Object");
+		return result;
+	}
+
+	result = fpgaObjectRead64(error_object, &value, 0);
+	if (result != FPGA_OK) {
+		OPAE_ERR("Failed to Read Object ");
+		goto out;
```
r-rojo: You lose the value of `result` when you `goto out`

tools/extra/ras/main.c
```diff
@@ -896,27 +896,63 @@ fpga_result inject_ras_errors(fpga_handle fme_handle,
...
+	}
+
+	result = fpgaObjectWrite64(error_object, value, 0);
+	if (result != FPGA_OK) {
+		OPAE_ERR("Failed to Write Object ");
+		goto out;
```
r-rojo: You lose the value of `result` when you `goto out`. Also not needed here since we can just fall through.

tools/extra/ras/main.c
```diff
@@ -834,6 +835,7 @@ fpga_result clear_port_errors(fpga_handle afu_handle)
 	result = fpgaObjectRead64(port_error_object, &value, 0);
 	if (result != FPGA_OK) {
 		OPAE_ERR("Failed to Read Object ");
+		resval = (result != FPGA_OK) ? result : resval;
```
r-rojo: If we're falling into this block because of the condition above `result != FPGA_OK` then it is not necessary to do the check again in the ?: operator. Because you're not using result in the error message when `fpgaDestroyObject` fails, you can call the destroy function without assigning it to a variable but instead just checking the result directly. This can avoid having two different result variables.

PR #1108, opened 2019-01-23
xfpga: get socketid in bare metal(PF)

anandaravuri:


libopae/plugins/xfpga/properties.c
```diff
@@ -224,15 +224,15 @@ fpga_result __FPGA_API__ xfpga_fpgaUpdateProperties(fpga_token token,
...
+			FPGA_SYSFS_SOCKET_ID, &value);
 
-	if (0 == resval) {
-		_iprop.socket_id = (uint8_t)value;
-		SET_FIELD_VALID(&_iprop, FPGA_PROPERTY_SOCKETID);
+		if (0 == resval) {
+			_iprop.socket_id = (uint8_t)value;
+			SET_FIELD_VALID(&_iprop, FPGA_PROPERTY_SOCKETID);
+		}
 	}
```
tswhison: Should there be..
```c
else
   return result;
```

anandaravuri: else return result works in PF , But fails in VF , VF doesn't get socket  ID

tswhison: ah, good point.

PR #1109, opened 2019-01-23
NLB0: Address failure in virtualization

jelonanX:


tools/extra/fpgadiag/nlb0.cpp
```diff
@@ -331,7 +331,15 @@ bool nlb0::setup()
...
 bool nlb0::run()
 {
-    auto fme_token = get_parent_token(accelerator_);
+    token::ptr_t fme_token = nullptr;
+
+    try {
+        fme_token = get_parent_token(accelerator_);
+    }
+    catch (opae::fpga::types::not_found) {
+        std::cerr << "Parent token not found. " << std::endl;
```
r-rojo: You might want to make this message more meaningful, like:
"Error locating device token. Stats will be disabled."

PR #1110, opened 2019-01-23
opae-c: add plugin cfg parsing

r-rojo:
Add logic in pluginmgr.c to look for and process plugin configuration
files.

The schema is still WIP.

### DIFF
--- a/libopae/pluginmgr.c
+++ b/libopae/pluginmgr.c
@@ -265,7 +265,7 @@ STATIC int process_plugin(const char *name, json_object *j_config)
 	JSON_GET(j_config, "configuration", &j_plugin_cfg);
 	JSON_GET(j_config, "enabled", &j_enabled);
 	if (json_object_get_string_len(j_plugin) > PLUGIN_NAME_MAX) {
-		OPAE_ERR("module name too long");
+		OPAE_ERR("plugin name too long");
 		return 1;
 	}
 
@@ -293,8 +293,8 @@ STATIC int process_plugin(const char *name, json_object *j_config)
 		goto out_err;
 	}
 
-	if (strcpy_s(cfg->module, PLUGIN_NAME_MAX, json_object_get_string(j_plugin))) {
-		OPAE_ERR("error copying module name");
+	if (strcpy_s(cfg->plugin, PLUGIN_NAME_MAX, json_object_get_string(j_plugin))) {
+		OPAE_ERR("error copying plugin name");
 		goto out_err;
 	}
 
@@ -309,6 +309,8 @@ out_err:
 	cfg->cfg_size = 0;
 	return 1;
 }
+
+
 
 STATIC int process_cfg_buffer(const char *buffer, const char *filename)
 {
@@ -550,7 +552,7 @@ STATIC int opae_plugin_mgr_load_cfg_plugin(int i)
 	opae_api_adapter_table *adapter = NULL;
 
 	if (cfg->enabled && cfg->cfg && cfg->cfg_size) {
-		adapter = opae_plugin_mgr_alloc_adapter(cfg->module);
+		adapter = opae_plugin_mgr_alloc_adapter(cfg->plugin);
 		if (!adapter) {
 			OPAE_ERR("malloc failed");
 			return 1;
--- a/libopae/pluginmgr.c
+++ b/libopae/pluginmgr.c
@@ -27,16 +27,14 @@
 #ifdef HAVE_CONFIG_H
 #include <config.h>
 #endif /* HAVE_CONFIG_H */
-
+#define _GNU_SOURCE
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdint.h>
 #include <dlfcn.h>
-#include <sys/stat.h>
 #include <sys/types.h>
 #include <dirent.h>
 #include <linux/limits.h>
-#define __USE_GNU
 #include <pthread.h>
 
 #include <json-c/json.h>
@@ -81,8 +79,8 @@ static int plugin_count;
 #define CFG_PATHS 4
 static const char *_opae_cfg_files[CFG_PATHS] = {
 	"/etc/opae.cfg",
-	"~/.config/opae/config",
-	"~/.local/opae/config",
+	"~/.config/opae/opae.cfg",
+	"~/.local/opae/opae.cfg",
 	"~/.local/opae.cfg"
 };
 
@@ -93,17 +91,18 @@ int opae_plugin_mgr_plugin_count(void)
 
 const plugin_cfg *opae_plugin_mgr_get(int i)
 {
-	return plugin_count ? &plugin_list[i] : NULL;
+	return (plugin_count && i < MAX_PLUGINS) ? &plugin_list[i] : NULL;
 }
 
-STATIC const char *find_cfg()
+STATIC char *find_cfg()
 {
 	int i = 0;
-	struct stat st;
+	char *canon_name = NULL;
 
 	for (; i < CFG_PATHS; ++i) {
-		if (stat(_opae_cfg_files[i], &st)) {
-			return _opae_cfg_files[i];
+		canon_name = canonicalize_file_name(_opae_cfg_files[i]);
+		if (canon_name) {
+			return canon_name;
 		}
 	}
 	return NULL;
@@ -259,13 +258,13 @@ STATIC int process_plugin(const char *name, json_object *j_config)
 {
 	plugin_cfg *cfg = &plugin_list[plugin_count];
 	const char *stringified = NULL;
-	json_object *j_module = NULL;
+	json_object *j_plugin = NULL;
 	json_object *j_plugin_cfg = NULL;
 	json_object *j_enabled = NULL;
-	JSON_GET(j_config, "module", &j_module);
+	JSON_GET(j_config, "plugin", &j_plugin);
 	JSON_GET(j_config, "configuration", &j_plugin_cfg);
 	JSON_GET(j_config, "enabled", &j_enabled);
-	if (json_object_get_string_len(j_module) > PLUGIN_NAME_MAX) {
+	if (json_object_get_string_len(j_plugin) > PLUGIN_NAME_MAX) {
 		OPAE_ERR("module name too long");
 		return 1;
 	}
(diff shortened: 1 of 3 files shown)


libopae/pluginmgr.c
```diff
@@ -71,6 +73,42 @@ STATIC opae_api_adapter_table *adapter_list = (void *)0;
 static pthread_mutex_t adapter_list_lock =
 	PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP;
 
+
+#define MAX_PLUGINS PLUGIN_SUPPORTED_DEVICES_MAX
+static plugin_cfg plugin_list[MAX_PLUGINS];
```
tswhison: I'm hesitant to set such limits..  If we only use a few of the entries, then space is wasted.  If we use a lot, we have to check for and handle the overflow case.
Can we use a linked list so that we allocate space as we go?

libopae/pluginmgr.c
```diff
@@ -71,6 +73,42 @@ STATIC opae_api_adapter_table *adapter_list = (void *)0;
...
+static int plugin_count;
+
+#define CFG_PATHS 4
+static const char *_opae_cfg_files[CFG_PATHS] = {
+	"/etc/opae.cfg",
+	"~/.config/opae/config",
+	"~/.local/opae/config",
+	"~/.local/opae.cfg"
```
tswhison: Why support different filenames?  Can we converge on opae.cfg?

libopae/pluginmgr.c
```diff
@@ -71,6 +73,42 @@ STATIC opae_api_adapter_table *adapter_list = (void *)0;
...
+};
+
+int opae_plugin_mgr_plugin_count(void)
+{
+	return plugin_count;
+}
+
+const plugin_cfg *opae_plugin_mgr_get(int i)
+{
+	return plugin_count ? &plugin_list[i] : NULL;
```
tswhison: Bounds checking on i?

libopae/pluginmgr.c
```diff
@@ -71,6 +73,42 @@ STATIC opae_api_adapter_table *adapter_list = (void *)0;
...
+}
+
+STATIC const char *find_cfg()
+{
+	int i = 0;
+	struct stat st;
+
+	for (; i < CFG_PATHS; ++i) {
+		if (stat(_opae_cfg_files[i], &st)) {
+			return _opae_cfg_files[i];
```
tswhison: In the case that both /etc/opae.cfg and ~/.local/opae.cfg exist, this will return /etc/opae.cfg.  Can we follow a different model?  Can we process the config file in /etc if any first, then let the local one override settings if it exists?  (similar to what git does, I think.)

libopae/pluginmgr.c
```diff
@@ -190,15 +240,166 @@ int opae_plugin_mgr_finalize_all(void)
...
+
+#define MAX_PLUGIN_CFG_SIZE 1024
+STATIC int process_plugin(const char *name, json_object *j_config)
+{
+	plugin_cfg *cfg = &plugin_list[plugin_count];
+	const char *stringified = NULL;
+	json_object *j_module = NULL;
+	json_object *j_plugin_cfg = NULL;
+	json_object *j_enabled = NULL;
+	JSON_GET(j_config, "module", &j_module);
```
tswhison: I think you mean module as in module .so file.  I'm concerned that some users won't understand this convention.  Could we use plugin for clarity?

libopae/pluginmgr.c
```diff
@@ -133,6 +171,18 @@ STATIC int opae_plugin_mgr_configure_plugin(opae_api_adapter_table *adapter,
 	return cfg(adapter, config);
 }
 
+STATIC void opae_plugin_mgr_reset_cfg(void)
+{
+	int i = 0;
+	for ( ; i < plugin_count; ++i) {
+		if (plugin_list[i].cfg) {
```
tswhison: Rather than maintaining a separate list, does it make sense to put this data into the adapter?

libopae/pluginmgr.c
r-rojo: OK. Sounds good. One of the reasons for doing this in a feature branch is to hash out the things like this.

libopae/pluginmgr.c
```diff
@@ -190,15 +240,166 @@ int opae_plugin_mgr_finalize_all(void)
...
+	char buffer[MAX_CFG_SIZE] = { 0 };
+	char *ptr = &buffer[0];
+	size_t bytes_read = 0, total_read = 0;
+	FILE *fp = fopen(filename, "r");
+	if (!fp) {
+		OPAE_ERR("Error opening config file: %s", filename);
+		return 1;
+	}
+
+	while ((bytes_read = fread(ptr + total_read, 1, 1, fp))
```
tswhison: The style checker won't like this.  (It got me for this one earlier today. :))

r-rojo: I _think_ this is OK. I think it depends on the usage (if vs while).

libopae/pluginmgr.c
r-rojo: Can you elaborate? I maintain a separate list because I separate parsing/loading.

libopae/pluginmgr.c
r-rojo: Ok

libopae/pluginmgr.c
tswhison: 👍

libopae/pluginmgr.c
tswhison: I think I see.  It's a chicken and egg.  You need to read the config first before you know which adapter to create.
```c
typedef struct _plugin_cfg {
	char name[PLUGIN_NAME_MAX];
	char module[PLUGIN_NAME_MAX];
	bool enabled;
	char *cfg;
	size_t cfg_size;
	uint32_t supported_devices[PLUGIN_SUPPORTED_DEVICES_MAX];
} plugin_cfg;
```
I guess I was wondering whether it made sense to store "cfg" and "size" in the adapter?

libopae/pluginmgr.c
r-rojo: Longer term, that would be nice. I think let's take small steps, and then iterate.

libopae/pluginmgr.c
r-rojo: can we do that in another PR?

tswhison: Yeah, sure.

libopae/pluginmgr.c
```diff
@@ -190,15 +239,175 @@ int opae_plugin_mgr_finalize_all(void)
...
+STATIC int process_plugin(const char *name, json_object *j_config)
+{
+	plugin_cfg *cfg = &plugin_list[plugin_count];
+	const char *stringified = NULL;
+	json_object *j_plugin = NULL;
+	json_object *j_plugin_cfg = NULL;
+	json_object *j_enabled = NULL;
+	JSON_GET(j_config, "plugin", &j_plugin);
+	JSON_GET(j_config, "configuration", &j_plugin_cfg);
+	JSON_GET(j_config, "enabled", &j_enabled);
```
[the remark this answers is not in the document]
r-rojo: the value for "configuration" is the plugin-specific configuration. "enabled" and "module" or "plugin" are used by the plugin manager for loading plugins.

libopae/pluginmgr.c
```diff
@@ -190,15 +239,175 @@ int opae_plugin_mgr_finalize_all(void)
...
+		goto out_err;
+	}
+
+	if (strcpy_s(cfg->name, PLUGIN_NAME_MAX, name)) {
+		OPAE_ERR("error copying plugin name");
+		goto out_err;
+	}
+
+	if (strcpy_s(cfg->plugin, PLUGIN_NAME_MAX, json_object_get_string(j_plugin))) {
+		OPAE_ERR("error copying plugin name");
```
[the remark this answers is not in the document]
r-rojo: it's the filename, I will update the message

testing/opae-c/test_pluginmgr_c.cpp
```diff
@@ -230,4 +231,179 @@ TEST_P(pluginmgr_c_p, bad_final_all) {
   EXPECT_EQ(2, test_plugin_finalize_called);
 }
 
+const char *plugin_cfg_1 = R"plug(
+{
+    "configurations": {
+        "plugin1": {
+            "configuration": {
```
lzhan55: Why the same key "configuration" was used twice?

r-rojo: The "plugins" list on line 254 is a list of plugins to load. 
Each entry in that list refers to a key in the "configurations" dictionary.
Each value in the "configurations" dictionary refers to the plugin data that the plugin manager uses.
The value for the "configuration" key in this plugin data is passed on to the plugin itself during the "configure" phase of loading plugins. It is also passed on as a raw JSON string so as to not force what kind of JSON parser the plugin uses. It will be up to the plugin to parse the raw JSON string and consume its data as it needs.

lzhan55: Thanks for the clarification. 

One more questions: at line 356 you get the 'j_config' which should be the under the key 'plugin1' from here. Then j_config should refer to the json object containing keys "configuration", "enabled" and "plugin", right?

libopae/pluginmgr.h
```diff
@@ -41,4 +41,17 @@ int opae_plugin_mgr_finalize_all(void);
 int opae_plugin_mgr_for_each_adapter(
 	int (*callback)(const opae_api_adapter_table *, void *), void *context);
 
+#define PLUGIN_SUPPORTED_DEVICES_MAX 256
+#define PLUGIN_NAME_MAX 64
+typedef struct _plugin_cfg {
+	char name[PLUGIN_NAME_MAX];
+	char plugin[PLUGIN_NAME_MAX];
```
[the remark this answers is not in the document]
r-rojo: The name here is just a name of a plugin. Theoretically, you could have one plugin library but included twice with different names (and plugin-specific configurations).

libopae/pluginmgr.c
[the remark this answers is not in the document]
lzhan55: Now the "module" was modified to "plugin" which makes sense. I was confused yesterday.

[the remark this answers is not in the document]
r-rojo: Yes, I changed it after @tswhison suggested it changed to "plugin".

PR #1111, opened 2019-01-24
Fix issues identified by checkers.

tswhison:


testing/opae-c/test_init_c.cpp
```diff
@@ -151,5 +151,6 @@ TEST(init, log_file) {
   opae_init();
 
   EXPECT_EQ(0, stat("opae_log.log", &buf));
+  unlink("opae_log.log");
```
ibarraza: At this point it is still trying to log to opae_log.log, right? Can we yank the file out from under it while it is in this state or is it going to try to do something with the file later upon closing the application or test?

tswhison: Good point.  Maybe we should remove it in TearDown?  I will make the change.

PR #1112, opened 2019-01-24
opae-c: don't apply versions to modules

tswhison:
Attempting to apply VERSION/SOVERSION to module libraries is causing
build errors on distros with newer versions of cmake. This change
removes the code that was doing that, and it changes the library type of
libbmc.so from SHARED to MODULE (it is dynamically loaded).

### DIFF
--- a/libopae/plugins/xfpga/metrics/bmc/CMakeLists.txt
+++ b/libopae/plugins/xfpga/metrics/bmc/CMakeLists.txt
@@ -27,7 +27,7 @@
 include_directories(
 	${OPAE_INCLUDE_DIR} )
 
-add_library(bmc SHARED
+add_library(bmc MODULE
 	bmc.c
 	bmcdata.c
 	bmc_ioctl.c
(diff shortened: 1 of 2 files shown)


libopae/plugins/xfpga/metrics/bmc/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@
 include_directories(
 	${OPAE_INCLUDE_DIR} )
 
-add_library(bmc SHARED
+add_library(bmc MODULE
```
asgibbons: Are you certain of this? libbmc.so is used on the link line, so it should be SHARED instead of MODULE.  MODULEs are loaded with dlopen and can't be specified on a link line (-lbmc).

tswhison: It is dlopen'ed on metrics/metrics_utils.c:857

asgibbons: But it is (I believe) linked with pacd and fpgainfo (doesn't fpgad use this?)

tswhison: I looked at fpgainfo's CMakeLists.txt.  It isn't linking with bmc.  If pacd is specifying -lbmc, is there any way it can be loaded dynamically instead.  fpgad isn't using the metrics api atm.  If we make such updates in the future, I think it will access libbmc.so by way of libxfpga.so only.

tswhison: The bmc CMakeLists.txt has provisions for building the library as libstaticbmc.a.  That might be another option.

r-rojo: I think at this point, we should make it a module and any applications that need it can link to bmcstatic.

PR #1115, opened 2019-01-25
Retry socket select operation if it fails with EINTR

JensHuthmann:
Sometimes the select is interrupted and thus returns EINTR. According to the reference the operation can be retried.

This fixed crashed in the ASE simulation where sockets where the event socket was not working.

I wonder if the use of the gnu macro is ok, or if the same behavior should be included in the opae source. What is the coding style guide for this? 

reference:
https://www.gnu.org/software/libc/manual/html_node/Interrupted-Primitives.html

### DIFF
--- a/ase/sw/protocol_backend.c
+++ b/ase/sw/protocol_backend.c
@@ -705,7 +705,7 @@ static void *start_socket_srv(void *args)
 
 	do {
 		FD_SET(sock_fd, &readfds);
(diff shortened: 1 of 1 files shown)


lzhan55:
@tswhison @r-rojo @zhangzhang10 Could you please review the changes for "Retry socket select operation if it fails with EINTR" ?

PR #1116, opened 2019-01-25
Update platform RTL Quartus IP from 18.1.

michael-adler:
- The only synthesis-visible change is a new timing constraint on Avalon DC FIFO.
- Update the script the pulls sources from a Quartus release due to file name changes.


michael-adler:
The FIFO is used only for Avalon MM clock crossing. Tested with platform-inserted clock crossing on all DCP platforms.

The changed shell script is run only when extracting sources by hand. I use it to generate the other files before checking them in. No other testing required.

PR #1117, opened 2019-01-26
opae-c: add dummy plugin and basic test

r-rojo:
This dummy plugin makes use of the configuration string by parsing it
and then using the values. The first two values are used to print out to
stdout which the tests verifies. The third value is the number of dummy
tokens to create which is then used for the rest of the test for
enumerating and opening/closing tokens.

This also fixes the ASSERT_NOT_NULL_MSG_RESULT macro to include a space
before "is NULL" in the mssage.

### DIFF
--- a/libopae/opae_int.h
+++ b/libopae/opae_int.h
@@ -62,7 +62,7 @@
 #define ASSERT_NOT_NULL(__arg) ASSERT_NOT_NULL_MSG(__arg, #__arg " is NULL")
 
 #define ASSERT_NOT_NULL_RESULT(__arg, __result)                                \
-	ASSERT_NOT_NULL_MSG_RESULT(__arg, #__arg "is NULL", __result)
+	ASSERT_NOT_NULL_MSG_RESULT(__arg, #__arg " is NULL", __result)
 
 #define ASSERT_RESULT(__result)                                                \
 	do {                                                                   \
--- a/testing/opae-c/dummy_plugin.c
+++ b/testing/opae-c/dummy_plugin.c
@@ -76,10 +76,19 @@ fpga_result DUMMY_HIDDEN dummy_plugin_fpgaEnumerate(const fpga_properties *filte
 
 	for ( ; i < min; ++i) {
 		dummy_token *t = (dummy_token*)malloc(sizeof(dummy_token));
+		if (!t) {
+			goto err_enum;
+		}
 		t->number = i;
 		tokens[i] = t;
 	}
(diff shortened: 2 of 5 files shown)


libopae/opae_int.h
```diff
@@ -62,7 +62,7 @@
 #define ASSERT_NOT_NULL(__arg) ASSERT_NOT_NULL_MSG(__arg, #__arg " is NULL")
 
 #define ASSERT_NOT_NULL_RESULT(__arg, __result)                                \
-	ASSERT_NOT_NULL_MSG_RESULT(__arg, #__arg "is NULL", __result)
+	ASSERT_NOT_NULL_MSG_RESULT(__arg, #__arg " is NULL", __result)
```
tswhison: Thanks!

testing/opae-c/dummy_plugin.c
```diff
@@ -0,0 +1,203 @@
...
+	uint32_t i = 0;
+	uint32_t min = MIN(_fake_tokens, max_tokens);
+	UNUSED_PARAM(filters);
+	UNUSED_PARAM(num_filters);
+	*num_matches = _fake_tokens;
+
+	for ( ; i < min; ++i) {
+		dummy_token *t = (dummy_token*)malloc(sizeof(dummy_token));
+		t->number = i;
+		tokens[i] = t;
```
tswhison: This always assumes that tokens is non-NULL.  I assume this is OK for the tests.

testing/opae-c/dummy_plugin.c
```diff
@@ -0,0 +1,203 @@
...
+
+fpga_result DUMMY_HIDDEN dummy_plugin_fpgaOpen(fpga_token t, fpga_handle *h, int flags)
+{
+	UNUSED_PARAM(h);
+	dummy_token *dt = (dummy_token*)t;
+	dummy_handle *dh = (dummy_handle*)malloc(sizeof(dummy_handle));
+	//printf("dummy/fpgaOpen %d %d\n", dh->number, flags);
+	dh->number = dt->number*flags*2;
+	dh->token = dt;
+	h = (fpga_handle)dh;
```
tswhison: Don't you need..
```c
*h = dh;
```

r-rojo: fixed

testing/opae-c/dummy_plugin.c
r-rojo: fixed

PR #1119, opened 2019-01-28
xfpga: fix static code scanner issues for metrics

anandaravuri:


anandaravuri:
> Not sure I feel comfortable about removing the error code checks. At a minimum, shouldn't we detect and respond to FPGA_NO_MEMORY when we fail to allocate mem?


Code scanner issue: Unreachable code
Removed Unreachable code 
fpga_vector_total() function returns FPGA_OK .

tswhison:
In that case, can we mark fpga_vector_total as a void function?  Seems like that would be better.  Thanks

anandaravuri:
> In that case, can we mark fpga_vector_total as a void function? Seems like that would be better. Thanks

Input parameters are always valid ->fpga_vector_total(fpga_enum_metrics_vector, &num_enun_metrics) 
KW shows 

fpga_result fpga_vector_total(fpga_metric_vector *vector, uint64_t *total)
{
	if (vector == NULL ||
		total == NULL) {
		FPGA_ERR("Invalid parm");
		return FPGA_EXCEPTION;
	}
	*total =  vector->total;
	return FPGA_OK;
}

tswhison:
I see.  I would say either mark the issue as "Not a problem" or else change the function type to void, since we have confirmation that no one is calling it with NULL ptrs.

PR #1122, opened 2019-01-28
xfpga: fix static code scanner issues for metrics

anandaravuri:


libopae/plugins/xfpga/metrics/metrics.c
```diff
@@ -273,8 +273,6 @@ fpga_result __FPGA_API__ xfpga_fpgaGetMetricsByIndex(fpga_handle handle,
 			result = FPGA_OK;
 		}
 
-	} else {
```
r-rojo: Why are we removing this?

anandaravuri: Not reachable code

PR #1135, opened 2019-02-04
opae-c: fix finding of cfg files

r-rojo:
* Switch to using $HOME instead of ~ in config file paths
* Add `resolve_file_name` that will resolve paths with environment
variables
* Add test for finding a config file in one of the pre-determined paths

### DIFF
--- a/libopae/pluginmgr.c
+++ b/libopae/pluginmgr.c
@@ -78,41 +78,57 @@ STATIC int opae_plugin_mgr_plugin_count;
 
 #define CFG_PATHS 4
 static const char *_opae_cfg_files[CFG_PATHS] = {
-	"/etc/opae.cfg",
+	"/etc/opae/opae.cfg",
 	"$HOME/.config/opae/opae.cfg",
 	"$HOME/.local/opae/opae.cfg",
 	"$HOME/.local/opae.cfg"
 };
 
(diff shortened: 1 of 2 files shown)


libopae/pluginmgr.c
```diff
@@ -79,20 +79,49 @@ STATIC int opae_plugin_mgr_plugin_count;
...
+	size_t len = strlen(src);
+	size_t len_cpy = len;
+
+	strncpy_s(src_cpy, PATH_MAX, src, len);
+
+	while(ptok && len_cpy) {
+		*pdst++ = '/';
+		ptok = strtok_s(NULL, &len_cpy, "/", &pstr);
+		if (ptok[0] == '$') {
+			strncpy_s(pdst, len, getenv(ptok+1), len);
```
tswhison: Probably should check whether getenv returns NULL and provide an error return code.

libopae/pluginmgr.c
```diff
@@ -79,20 +79,49 @@ STATIC int opae_plugin_mgr_plugin_count;
 #define CFG_PATHS 4
 static const char *_opae_cfg_files[CFG_PATHS] = {
 	"/etc/opae.cfg",
```
tswhison: Maybe this should be /etc/opae/opae.cfg?

PR #1137, opened 2019-02-05
opae-c: add default plugins if no config file

r-rojo:
In the event that no config files are found or plugins loaded, go back
to loading native plugins as the default ones.

### DIFF
--- a/libopae/pluginmgr.c
+++ b/libopae/pluginmgr.c
@@ -652,6 +652,7 @@ STATIC int opae_plugin_mgr_load_dflt_plugins(int *platforms_detected)
 	if (errors)
 		return errors;
 	// Load each of the native plugins that were detected.
+	*platforms_detected = 0;
 
 	for (i = 0 ; platform_data_table[i].native_plugin ; ++i) {
 		const char *native_plugin;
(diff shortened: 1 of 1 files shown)


libopae/pluginmgr.c
```diff
@@ -643,37 +643,14 @@ STATIC int opae_plugin_mgr_load_cfg_plugins(void)
...
-			free(found_cfg);
-		}
-	}
-
-	errors = opae_plugin_mgr_detect_platforms();
+	int i = 0, j = 0;
+	int res = 0;
+	opae_api_adapter_table *adapter = NULL;
+	int errors = opae_plugin_mgr_detect_platforms();
 	if (errors)
```
tswhison: Do we want to initialize *platforms_detected to 0 before the loop?

libopae/pluginmgr.c
```diff
@@ -695,7 +672,7 @@ int opae_plugin_mgr_initialize(const char *cfg_file)
 						platform_data_table[j].native_plugin, &res)) {
 				OPAE_ERR("strcmp_s failed");
 				++errors;
-				goto out_unlock;
+				goto out;
```
tswhison: No mutex to unlock, so maybe just return errors here?

libopae/pluginmgr.c
```diff
@@ -713,7 +690,7 @@ int opae_plugin_mgr_initialize(const char *cfg_file)
 		if (!adapter) {
 			OPAE_ERR("malloc failed");
 			++errors;
-			goto out_unlock;
+			goto out;
```
tswhison: No mutex to unlock, so maybe just return errors here?

PR #1138, opened 2019-02-06
testing: add mock_opae class

r-rojo:
`mock_opae_p` class is a templated class that derives from TestWithParam
and is meant to be a mock class for mock platforms.
The template argument is used to specify how many tokens to use
(if any).

This class implements the SetUp/TearDown that google test uses in the fixture
but also defines test_setup/test_teardown functions that can be overridden by
a sub-class.

This is a first step in refactoring tests to avoid code cloning.

### DIFF
--- a/testing/mock/mock_opae.h
+++ b/testing/mock/mock_opae.h
@@ -1,0 +1,77 @@
+// Copyright(c) 2017-2019, Intel Corporation
+//
+// Redistribution  and  use  in source  and  binary  forms,  with  or  without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of  source code  must retain the  above copyright notice,
+//   this list of conditions and the following disclaimer.
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+// * Neither the name  of Intel Corporation  nor the names of its contributors
+//   may be used to  endorse or promote  products derived  from this  software
+//   without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,  BUT NOT LIMITED TO,  THE
+// IMPLIED WARRANTIES OF  MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED.  IN NO EVENT  SHALL THE COPYRIGHT OWNER  OR CONTRIBUTORS BE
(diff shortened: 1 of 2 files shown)


testing/mock/mock_opae.h
```diff
@@ -0,0 +1,77 @@
...
+      if (t) {
+        EXPECT_EQ(fpgaDestroyToken(&t), FPGA_OK);
+        t = nullptr;
+      }
+    }
+    test_teardown();
+    system_->finalize();
+  }
+
+  virtual void test_setup() {
```
tswhison: Just curious.. is there an advantage of providing these virtual overrides vs using the SetUp and TearDown that are already provided by gtest?  (A derived class could be required to call the parent implementation.)

testing/mock/mock_opae.h
```diff
@@ -0,0 +1,77 @@
...
+#include "gtest/gtest.h"
+#include "test_system.h"
+
+namespace opae {
+namespace testing {
+
+template<int _T = 2>
+class mock_opae_p : public ::testing::TestWithParam<std::string> {
+ protected:
+  mock_opae_p(): tokens_{ {} } {}
```
jelonanX: There might be a leak if it's the array is not initialized?

testing/mock/mock_opae.h
r-rojo: The SetUp and TearDown provided by gtest are the interfaces for gtest meaning they will always be called by gtest. The test_setup/test_teardown are the opae mock interface. The mock base class implements just enough boiler plate in the SetUp/TearDown and then calls test_setup/test_teardown which can be overridden by derived classes. This means the derived classes don't have to worry about mock setup (test_system/mock sysfs) and only concentrate on their specific routines.

testing/mock/mock_opae.h
r-rojo: Not sure, seeing as it depends on the template argument.

r-rojo: So I tried a little experiment with some sample code (pasted below).
I compiled with gcc 5.2 using our compiler configuration. I didn't get any build errors.
When I ran this test through valgrind, I saw that it did complained after running d10j.run().
Also, the contents of the array in d10 are zero whereas the contents of d10j have junk.
So it seems this initialization is OK.

```C++
#include <array>
#include <iostream>
#include <string>
template <int _T = 2> class dummy {
public:
  dummy() : x_("initialized"), a_ { {} } {}
  //dummy() : x_(0) {}
  dummy(const char *x) : x_(x) {}

  void run() {
    std::cout << x_ << ": dummy<" << _T << ">::run()\n";
    for (auto v : a_) {
      std::cout << v << "\n";
    }
  }

private:
  const char *x_;
  std::array<int, _T> a_;
};

int main(int argc, char *argv[]) {
  (void)argc;
  (void)argv;
  std::array<int, 10> a{{0}};
  for (auto v : a) {
    std::cout << v << "\n";
  }

  dummy<0> d0;
  d0.run();
  dummy<10> d10;
  d10.run();
  dummy<10> d10j("junk");
  d10j.run();
  return 0;
}
```

PR #1139, opened 2019-02-06
fpgad: start of config file support

tswhison:
Add config file search and load support by way of --config,-c.
Coming commits will complete the config file parsing and
utilization of the loaded configuration.

### DIFF
--- a/tools/base/fpgad/command_line.c
+++ b/tools/base/fpgad/command_line.c
@@ -30,6 +30,8 @@
 
 #include <getopt.h>
 #include "command_line.h"
+#include "config_file.h"
+#include "monitored_device.h"
 
 #ifdef LOG
 #undef LOG
@@ -37,7 +39,9 @@
 #define LOG(format, ...) \
 log_printf("args: " format, ##__VA_ARGS__)
 
-#define OPT_STR ":hdl:p:s:n:"
+extern fpgad_supported_device default_supported_devices_table[];
+
+#define OPT_STR ":hdl:p:s:n:c:"
 
 STATIC struct option longopts[] = {
 	{ "help",           no_argument,       NULL, 'h' },
@@ -46,6 +50,7 @@ STATIC struct option longopts[] = {
 	{ "pidfile",        required_argument, NULL, 'p' },
 	{ "socket",         required_argument, NULL, 's' },
 	{ "null-bitstream", required_argument, NULL, 'n' },
+	{ "config",         required_argument, NULL, 'c' },
 
 	{ 0, 0, 0, 0 }
 };
@@ -54,6 +59,7 @@ STATIC struct option longopts[] = {
 #define DEFAULT_DIR_ROOT_SIZE 13
 #define DEFAULT_LOG           "fpgad.log"
 #define DEFAULT_PID           "fpgad.pid"
+#define DEFAULT_CFG           "fpgad.cfg"
 
 void cmd_show_help(FILE *fptr)
 {
@@ -65,6 +71,7 @@ void cmd_show_help(FILE *fptr)
 	fprintf(fptr, "\t-s,--socket <sock>          the unix domain socket [/tmp/fpga_event_socket].\n");
 	fprintf(fptr, "\t-n,--null-bitstream <file>  NULL bitstream (for AP6 handling, may be\n"
 		      "\t                            given multiple times).\n");
+	fprintf(fptr, "\t-c,--config <file>          the configuration file [%s].\n", DEFAULT_CFG);
 }
 
 STATIC bool cmd_register_null_gbs(struct fpgad_config *c, char *null_gbs_path)
@@ -167,6 +174,16 @@ int cmd_parse_args(struct fpgad_config *c, int argc, char *argv[])
 			}
 			break;
 
+		case 'c':
+			if (tmp_optarg) {
+				strncpy_s(c->cfgfile, sizeof(c->cfgfile),
+						tmp_optarg, PATH_MAX);
+			} else {
+				LOG("missing cfgfile parameter.\n");
+				return 1;
+			}
+			break;
+
 		case ':':
 			LOG("Missing option argument.\n");
 			return 1;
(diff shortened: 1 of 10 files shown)


tools/base/fpgad/command_line.c
```diff
@@ -294,6 +312,57 @@ void cmd_canonicalize_paths(struct fpgad_config *c)
...
+		def = true;
+
+	if (def || (c->cfgfile[0] == '\0')) {
+		search = true;
+	} else {
+		char *canon_path;
+
+		canon_path = canonicalize_file_name(c->cfgfile);
+		if (canon_path) {
+			if (!stat(canon_path, &stat_buf)) {
```
r-rojo: I think `canonicalize_file_name`  returns NULL if a component of the path doesn't exist.
http://man7.org/linux/man-pages/man3/canonicalize_file_name.3.html
image

tools/base/fpgad/config_file.c
```diff
@@ -30,7 +30,257 @@
...
 
+	e = getenv("HOME");
+	if (e) {
+		// try $HOME/.opae/fpgad.cfg
+		snprintf_s_s(path, sizeof(path),
+			     "%s/.opae/fpgad.cfg", e);
 
+		canon = canonicalize_file_name(path);
+		if (canon) {
+			if (!stat(canon, &stat_buf)) {
```
r-rojo: I think canonicalize_file_name returns NULL if file does not exist

tools/base/fpgad/config_file.c
```diff
@@ -30,7 +30,257 @@
...
+
+			free(canon);
+		}
+
+	}
+
+	// try /etc/opae/fpgad.cfg
+	canon = canonicalize_file_name("/etc/opae/fpgad.cfg");
+	if (canon) {
+		if (!stat(canon, &stat_buf)) {
```
r-rojo: I think canonicalize_file_name returns NULL if file does not exist

tools/base/fpgad/config_file.c
```diff
@@ -30,7 +30,257 @@
...
+
+		}
+
+		free(canon);
+	}
+
+	// try /var/lib/opae/fpgad.cfg
+	canon = canonicalize_file_name("/var/lib/opae/fpgad.cfg");
+	if (canon) {
+		if (!stat(canon, &stat_buf)) {
```
r-rojo: I think canonicalize_file_name returns NULL if file does not exist

tools/base/fpgad/config_file.c
```diff
@@ -30,7 +30,257 @@
...
 
+#include <json-c/json.h>
 
+#ifdef LOG
+#undef LOG
+#endif
+#define LOG(format, ...) \
+log_printf("cfg: " format, ##__VA_ARGS__)
 
+int cfg_find_config_file(struct fpgad_config *c)
```
r-rojo: Would it be worth defining a macro that checks a path and copies it to config structure?

tools/base/fpgad/config_file.c
```diff
@@ -30,7 +30,257 @@
...
+} cfg_vendor_device_id;
+
+typedef struct _cfg_plugin_configuration {
+	bool enabled;
+	char *library;
+	cfg_vendor_device_id *devices;
+	struct _cfg_plugin_configuration *next;
+} cfg_plugin_configuration;
+
+#if 0
```
r-rojo: ?

tools/base/fpgad/command_line.c
tswhison: Ok, will remove the stat.

tools/base/fpgad/config_file.c
tswhison: Used in coming commits.

tools/base/fpgad/command_line.c
```diff
@@ -167,6 +174,16 @@ int cmd_parse_args(struct fpgad_config *c, int argc, char *argv[])
 			}
 			break;
 
+		case 'c':
+			if (tmp_optarg) {
+				strncpy_s(c->cfgfile, sizeof(c->cfgfile),
```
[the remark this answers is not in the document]
tswhison: We know that both c->cfgfile and tmp_optarg are non-NULL, and we know that sizeof(c->cfgfile) and PATH_MAX are within the buffer bounds.

tools/base/fpgad/config_file.c
tswhison: sure

tools/base/fpgad/config_file.c
tswhison: fixed

tools/base/fpgad/config_file.c
tswhison: fixed

tools/base/fpgad/config_file.c
tswhison: fixed

tools/base/fpgad/config_file.c
```diff
@@ -30,7 +30,216 @@
...
+#undef LOG
+#endif
+#define LOG(format, ...) \
+log_printf("cfg: " format, ##__VA_ARGS__)
 
+#define CFG_TRY_FILE(__f) \
+do { \
+	canon = canonicalize_file_name(__f); \
+	if (canon) { \
+		err = strncpy_s(c->cfgfile, \
```
r-rojo: would it make sense to parameterize c->cfgfile in the macro?

tools/base/fpgad/command_line.c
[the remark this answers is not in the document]
r-rojo: We don't want to validate user input here too? Are symbolic links OK?

tools/base/fpgad/config_file.c
tswhison: I'm not sure in this case.  Ideally we would also parameterize canon and err, etc.  I'm not sure it buys anything since the macro will only be used here.

tools/base/fpgad/command_line.c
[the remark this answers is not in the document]
tswhison: Good point.  I'll add it to the canonicalization code.

PR #1140, opened 2019-02-06
Fix compilation of fpgad on older systems.

michael-adler:


michael-adler:
@OPAE/tools please review.

PR #1141, opened 2019-02-07
fpgad: prevent symlinks in config file paths

tswhison:


tools/base/fpgad/command_line.c
```diff
@@ -328,15 +328,23 @@ void cmd_canonicalize_paths(struct fpgad_config *c)
...
-			strncpy_s(c->cfgfile,
-				  sizeof(c->cfgfile),
-				  canon_path,
-				  strnlen_s(canon_path, PATH_MAX));
 
-			if (!cfg_load_config(c)) {
-				LOG("daemon cfg file is %s\n",
-				    c->cfgfile);
-				search = false; // found and loaded it
+			stat(c->cfgfile, &stat_buf);
```
r-rojo: Can components of the path be symlinks?

tswhison: fixed

PR #1145, opened 2019-02-08
Prevent user from hurting himself

dasm:
Python has many ways of executing shell commands. By passing 'shell=True'
we've been exposing program for malicious behavior.
To prevent from doing that use built-in solution in Python [1] which helps
with blocking from shell injection.

[1] https://docs.python.org/2/library/subprocess.html#frequently-used-arguments

Signed-off-by: Dariusz Smigiel dariusz.smigiel@intel.com

### DIFF
--- a/tools/extra/pac/fpgabist/bist_app.py
+++ b/tools/extra/pac/fpgabist/bist_app.py
@@ -41,9 +41,9 @@ class BistMode(bc.BistMode):
         bc.load_gbs(gbs_path, bus_num)
         for func, param in self.executables.items():
             print "Running {} test...\n".format(func)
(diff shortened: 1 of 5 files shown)


dasm:
@asgibbons I see errors not connected to my PR.
Let me fix them, create separate PR and after merging I'll update this PR.

buie1:
@dasm I would try updating your branch before fixing whichever changes you were going to make addressing CI issues.

dasm:
@asgibbons I've noticed that there was missing bracket which caused bunch of errors (good thing to have PEP8 check).
I've fixed that.

dasm:
@buie1 yes, automated regression test was executed.

PR #1146, opened 2019-02-08
fpgad: add config file processing

tswhison:
Parse the given (--config,-c) configuration file, constructing a
supported devices table based on its contents. If parsing fails,
revert to a default supported devices table.

### DIFF
--- a/tools/base/fpgad/config_file.c
+++ b/tools/base/fpgad/config_file.c
@@ -191,7 +191,7 @@ cfg_process_plugin_devices(const char *name,
 	int i;
 	int devs;
 	cfg_vendor_device_id *head = NULL;
-	cfg_vendor_device_id *id;
+	cfg_vendor_device_id *id = NULL;
 	uint16_t vendor_id;
 	uint16_t device_id;
 	const char *s;
--- a/tools/base/fpgad/config_file.c
+++ b/tools/base/fpgad/config_file.c
@@ -29,6 +29,8 @@
 #endif // HAVE_CONFIG_H
 
 #include "config_file.h"
+#include "monitored_device.h"
+#include "api/sysfs.h"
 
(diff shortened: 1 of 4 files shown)


tools/base/fpgad/config_file.c
```diff
@@ -179,7 +183,299 @@ STATIC cfg_plugin_configuration *alloc_configuration(bool enabled,
...
+
+		if (!head) {
+			head = alloc_device(vendor_id, device_id);
+			if (!head) {
+				LOG("malloc failed.\n");
+				goto out_free;
+			}
+
+			id = head;
+		} else {
```
jelonanX: Head is initialized to null, would it ever get into the else?

tswhison: This is within the for loop, so !head will only be true the first time through.  Subsequent iterations will it the else case and append to the end of the list via id.

jelonanX: 👍

tools/base/fpgad/config_file.c
```diff
@@ -183,7 +187,299 @@ STATIC cfg_plugin_configuration *alloc_configuration(bool enabled,
...
+			goto out_free;
+		}
+
+		if (json_object_array_length(j_dev) != 2) {
+			LOG("%s 'devices' entry %d not array[2].\n",
+				name, i);
+			goto out_free;
+		}
+
+		j_vid = json_object_array_get_idx(j_dev, 0);
```
r-rojo: You think this (getting or parsing an int from a jarray at index I) could be refactored into a function?

tswhison: Sure.. but for 2 items, I'm not sure it's worth the trouble.

PR #1148, opened 2019-02-11
Add support for Qsys blocks that generate VHDL.

michael-adler:
- Detect generated VHDL files and add them to the simulation.
- Detect HEX memory images and put them in the proper location.

### DIFF
--- a/ase/Makefile
+++ b/ase/Makefile
@@ -517,8 +517,18 @@ ifdef DUT_VLOG_SRC_LIST
 	cd $(WORK) ; vlog $(MENT_VLOG_OPT) $(ASE_PLATFORM_INC) -F $(DUT_VLOG_SRC_LIST) -l vlog-afu.log
 endif
 
+$(WORK):
+	mkdir -p $(WORK)
+	@# Link to HEX memory images generated and required by Qsys
+	@if [ -f qsys_hex_sim_files.list ]; then \
+	  for h in `cat qsys_hex_sim_files.list`; do \
+	    echo Linking to $${h} ; \
+	    ln -s ../$${h} work ; \
+	  done ; \
+	fi
+
 ## VCS base Quartus Verilog libraries
-$(WORK)/synopsys_sim_quartus_verilog.setup:
+$(WORK)/synopsys_sim_quartus_verilog.setup: $(WORK)
 	mkdir -p $(WORK)/verilog_libs
 ifeq ($(GLS_SIM), 1)
 	@# Generate the Quartus family-dependent simulation library list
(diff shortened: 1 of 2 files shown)


michael-adler:
Tests ok: http://sw-pert.altera.com/pert/pert.php?test_run_id=3075307

michael-adler:
Tests still ok: http://sw-pert.altera.com/pert/pert.php?test_run_id=3077576

ase/scripts/generate_ase_environment.py
```diff
@@ -509,6 +509,26 @@ def config_qsys_sources(filelist, vlog_srcs, vhdl_srcs):
...
+    root, ext = os.path.splitext(fn)
+    if (root.endswith('package') or root.endswith('pkg')):
+        return '0' + fn
+    if (root.endswith('library') or root.endswith('lib')):
+        return '1' + fn
+    if (root.endswith('functions')):
+        return '2' + fn
+    if (root.endswith('components')):
+        return '3' + fn
+    return '4' + fn
```
lzhan55: Do qsys generated files follow the naming convention of pkg, lib, functions and components?

michael-adler: The key files do. It's the best I could come up with.

ase/Makefile
```diff
@@ -517,8 +517,18 @@ ifdef DUT_VLOG_SRC_LIST
 	cd $(WORK) ; vlog $(MENT_VLOG_OPT) $(ASE_PLATFORM_INC) -F $(DUT_VLOG_SRC_LIST) -l vlog-afu.log
 endif
 
+$(WORK):
+	mkdir -p $(WORK)
+	@# Link to HEX memory images generated and required by Qsys
+	@if [ -f qsys_hex_sim_files.list ]; then \
+	  for h in `cat qsys_hex_sim_files.list`; do \
+	    echo Linking to $${h} ; \
+	    ln -s ../$${h} work ; \
```
lzhan55: Is it possible that the "work" lib could be linked by multiple hex memory image files here?

michael-adler: I assume you mean multiple hex files sharing the same name. No. The sim_files_found set guarantees that all file names are unique. This is a Qsys property -- that files with the same name anywhere in the tree have the same contents. Qsys adds hashes to generated file names when necessary to preserve this constraint.

lzhan55: Isn't the "work" pointing to different hex image files here? Not sure how Qsys works.

michael-adler: No. Links are created in the ASE work directory to hex files in the Qsys-generated tree. You'd think that Qsys would generate names with absolute paths, or do something clever to load hex files relative to the source files that open them. It does neither, instead opening .hex files in the work directory.

Note that the local memory emulation already does the same thing in the Makefile.

lzhan55: One more generic questions regarding the Quartus sim library: all AFU will linked to all the simulation lib in Quartus no matter it needs it or not?

michael-adler: Correct. This has not changed. There is no good way to predict when it will be needed.

lzhan55: Thanks.

PR #1149, opened 2019-02-11
fpgad: update unit tests for command line

tswhison:
Update command_line.c unit tests after adding configuration file
support.

### DIFF
--- a/testing/fpgad/test_command_line_c.cpp
+++ b/testing/fpgad/test_command_line_c.cpp
@@ -314,6 +314,7 @@ TEST_P(fpgad_command_line_c_p, canonicalize4) {
   cmd_canonicalize_paths(&config_);
 
   char *d = get_current_dir_name();
+  ASSERT_NE(d, nullptr);
   std::string cfg_file = std::string(d) + std::string("/") + std::string(cfg_file_);
 
   EXPECT_STREQ(config_.cfgfile, cfg_file.c_str());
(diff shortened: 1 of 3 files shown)


testing/fpgad/test_command_line_c.cpp
```diff
@@ -234,5 +244,179 @@ TEST_P(fpgad_command_line_c_p, canonicalize1) {
...
+
+  strcpy(config_.cfgfile, cfg_file_);
+  std::ofstream cfg;
+  cfg.open(cfg_file_, std::ios::out);
+  cfg.write(valid_cfg, strlen(valid_cfg));
+  cfg.close();
+
+  cmd_canonicalize_paths(&config_);
+
+  char *d = get_current_dir_name();
```
jelonanX: Would it be a good idea to do an assert on d != NULL before continuing?

tswhison: :+1:

testing/fpgad/test_command_line_c.cpp
```diff
@@ -234,5 +244,181 @@ TEST_P(fpgad_command_line_c_p, canonicalize1) {
...
+ * @test       symlink4
+ * @brief      Test: cmd_path_is_symlink
+ * @details    If the given file name exists,<br>
+ *             and it contains a / character in position 0,<br>
+ *             and there is a symlink in any of the path components,<br>
+ *             then the fn returns true.<br>
+ */
+TEST_P(fpgad_command_line_c_p, symlink4) {
+  std::string s;
+  char *d = get_current_dir_name();
```
jelonanX: found another one. 👍

PR #1150, opened 2019-02-12
fpgad: add unit tests for config_file.c

tswhison:


tswhison:
I don't check the content of the messages explicitly, though I have seen the correct output. I rely on the return codes from the fn under test and the collected coverage data to verify the code paths as I develop the tests.

PR #1156, opened 2019-02-13
xfpga: fix parsing of BBS ID

tswhison:
Provide function to resolve a given vendor/device ID pair to the
corresponding enum fpga_hw_type, which abstracts PF/VF differences.
Move metrics code away from explicit reference to device ID.


libopae/plugins/xfpga/sysfs.c
```diff
@@ -1173,22 +1173,66 @@ fpga_result get_port_sysfs(fpga_handle handle, char *sysfs_port)
...
+			hw_type = FPGA_HW_DCP_VC;
+		break;
+
+		default:
+			FPGA_ERR("unknown device id: 0x%04x", device_id);
+		}
+
+	} else {
+		FPGA_ERR("unknown vendor id: 0x%04x", vendor_id);
+	}
```
r-rojo: Should we define constants/macros for these ids?

tswhison: I'd prefer to only track them here, and use fpga_hw_type otherwise.  I was hoping to remove the notion of referring directly to device id's in the code.

libopae/plugins/xfpga/enum.c
```diff
@@ -406,13 +407,23 @@ STATIC fpga_result enum_fme(const char *sysfspath, const char *name,
 		return FPGA_NOT_FOUND;
 	}
 
-	pdev->fpga_bbs_version.major =
-		FPGA_BBS_VER_MAJOR(pdev->fpga_bitstream_id);
-	pdev->fpga_bbs_version.minor =
-		FPGA_BBS_VER_MINOR(pdev->fpga_bitstream_id);
-	pdev->fpga_bbs_version.patch =
-		FPGA_BBS_VER_PATCH(pdev->fpga_bitstream_id);
+	hw_type = opae_id_to_hw_type(pdev->vendor_id, pdev->device_id);
```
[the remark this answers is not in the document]
tswhison: I really liked how you abstracted the notion of hardware type into the fpga_hw_type enum, so I'm trying to hide as many details as I can inside that abstraction.  The way that this register is coded, it lends itself nicely to an MCP/DCP pattern.

libopae/plugins/xfpga/properties.c
```diff
@@ -196,12 +220,21 @@ fpga_result __FPGA_API__ xfpga_fpgaUpdateProperties(fpga_token token,
 		}
 		SET_FIELD_VALID(&_iprop, FPGA_PROPERTY_BBSID);
 
-		_iprop.u.fpga.bbs_version.major =
-			FPGA_BBS_VER_MAJOR(_iprop.u.fpga.bbs_id);
-		_iprop.u.fpga.bbs_version.minor =
-			FPGA_BBS_VER_MINOR(_iprop.u.fpga.bbs_id);
-		_iprop.u.fpga.bbs_version.patch =
-			FPGA_BBS_VER_PATCH(_iprop.u.fpga.bbs_id);
+		if (hw_type == FPGA_HW_MCP) {
```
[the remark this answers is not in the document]
tswhison: The register encoding naturally lends itself to an MCP vs DCP abstraction.  I'd prefer to keep it that way for now.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -1173,22 +1173,66 @@ fpga_result get_port_sysfs(fpga_handle handle, char *sysfs_port)
...
-// get fpga device id
-fpga_result get_fpga_deviceid(fpga_handle handle, uint64_t *deviceid)
+enum fpga_hw_type opae_id_to_hw_type(uint16_t vendor_id, uint16_t device_id)
+{
+	enum fpga_hw_type hw_type = FPGA_HW_UNKNOWN;
+
+	if (vendor_id == 0x8086) {
+
+		switch (device_id) {
+		case 0xbcbc:
```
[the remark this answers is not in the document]
tswhison: I'd prefer to keep it localized here..  the idea is to prevent their direct usage in the code, because a single device ID doesn't capture both the PF/VF aspect.  eg, there was a check for 0xbcc0, but what about 0xbcc1?  So long as we use fpga_hw_type, we're ok for both.

libopae/plugins/xfpga/types_int.h
```diff
@@ -73,15 +73,13 @@
...
 
-// Discrete FPGA Device ID
-#define FPGA_DISCRETE_DEVICEID 0x09c4
-
-#define FPGA_BBS_VER_MAJOR(i) (((i) >> 56) & 0xf)
-#define FPGA_BBS_VER_MINOR(i) (((i) >> 52) & 0xf)
-#define FPGA_BBS_VER_PATCH(i) (((i) >> 48) & 0xf)
+#define DCP_FPGA_BBS_VER_MAJOR(i) (((i) >> 60) & 0xf)
+#define DCP_FPGA_BBS_VER_MINOR(i) (((i) >> 56) & 0xf)
+#define DCP_FPGA_BBS_VER_PATCH(i) (((i) >> 52) & 0xf)
```
asgibbons: I thought the patch number was 16 bits?  Change 0xf to 0xff?

tswhison: I checked with each of the CSR spreadsheets.  Each say that patch is bits [55:52] for DCP and bits [51:48] for MCP.

PR #1165, opened 2019-02-20
Lzhan55/new ase test1

lzhan55:
This pull request is the start of porting ASE tests from folder tests to folder testing in GitHub repo.

The main goal of the OPAE's testing framework is to enable (Google test) tests that don't require the presence of either an OPAE kernel driver or an OPAE compatible FPGA hardware. However, ASE doesn't need a  kernel driver, nor does it need a mock driver. But ASE tests can still utilize the malloc negative test capabilities in "test_system".


tswhison:
Hi @lzhan55, looks like this review pulls in some extra commits.  Does it make sense to rework it so that only your tests are up for review?

lzhan55:
@tswhison it is clean now with only my changes.

lzhan55:
The coverage number dropped because ASE source code is now added to the coverage statistics. With the ASE tests fully ported, this number will increase.

testing/CMakeLists.txt
```diff
@@ -96,6 +96,41 @@ target_include_directories(xfpga-static PUBLIC
   PRIVATE ${OPAE_SDK_SOURCE}/libopae/plugins/xfpga
   PRIVATE ${OPAE_SDK_SOURCE}/libopae)
 
+add_library(opae-c-ase-static STATIC
```
tswhison: Do you plan to continue calling the ase library opae-c-ase?  I think this made sense when ASE mirrored the opae-c API; but now that it's going to be a plugin, this naming may cause some confusion.  Maybe libase.so would make more sense now?

testing/CMakeLists.txt
```diff
@@ -96,6 +96,41 @@ target_include_directories(xfpga-static PUBLIC
...
+        ${OPAE_SDK_SOURCE}/ase/api/src/version.c
+        ${OPAE_SDK_SOURCE}/ase/api/src/wsid_list.c
+        ${OPAE_SDK_SOURCE}/ase/api/src/error.c)
+
+target_include_directories(opae-c-ase-static PUBLIC
+    $<BUILD_INTERFACE:${OPAE_INCLUDE_DIR}>
+    $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>
+    $<INSTALL_INTERFACE:include>
+    PRIVATE ${OPAE_SDK_SOURCE}/ase/api/src
+    PRIVATE ${OPAE_SDK_SOURCE}/ase/sw)
```
tswhison: Could we consider moving the ASE source directory under libopae/plugins in the repo?

testing/CMakeLists.txt
```diff
@@ -309,6 +344,7 @@ macro(add_unit_test test_name static_lib)
         ${ARGN})
     target_include_directories(${test_name} PRIVATE
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+        PRIVATE ${OPAE_SDK_SOURCE}/libopae/plugins/xfpga
```
tswhison: This makes xfpga include files visible to all tests.  While it may make sense for your tests in the short term, won't this dependency go away eventually?  Wondering if it makes more sense to only add this to the ASE tests?

testing/ase/test_open_c_ase.cpp
```diff
@@ -0,0 +1,127 @@
...
+        tok = &_tok;
+        token_for_afu0(&_tok);
+        accel_ = nullptr;
+    }
+
+    virtual void TearDown() override {
+        system_->finalize();
+    }
+
+    struct _fpga_token _tok;
```
tswhison: Should be tok_ to match the naming convention.

testing/ase/test_open_c_ase.cpp
```diff
@@ -0,0 +1,127 @@
...
+        system_ = test_system::instance();
+        system_->initialize();
+
+        tok = &_tok;
+        token_for_afu0(&_tok);
+        accel_ = nullptr;
+    }
+
+    virtual void TearDown() override {
+        system_->finalize();
```
tswhison: Any future test cases which successfully open accel_ will leak it here.  Add logic here to close a non-NULL accel_ as a safeguard.

testing/CMakeLists.txt
lzhan55: Agree and will rename the ASE library.

testing/CMakeLists.txt
lzhan55: I would move it for the JIRA task of adding ASE plugin. This task is only for porting the ASE tests to new test framework.

tswhison: sure

testing/CMakeLists.txt
lzhan55: ASE doesn't need plugins/xfpga. I moved this include folder from the global include_directories() here since it caused ASE tests compilation errors.

testing/ase/test_open_c_ase.cpp
lzhan55: Will fix it.

testing/ase/test_open_c_ase.cpp
lzhan55: Agree.

testing/CMakeLists.txt
r-rojo: Wouldn't it make more sense to move this to the new `add_ase_test` macro then?

testing/ase/test_open_c_ase.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include "test_system.h"
+
+using namespace opae::testing;
+
+// ASE ID
+#define ASE_TOKEN_MAGIC    0x46504741544f4b40
+static const fpga_guid ASE_GUID = {
+			0xd8, 0x42, 0x4d, 0xc4, 0xa4,  0xa3, 0xc4, 0x13, 0xf8,0x9e,
+						0x43, 0x36, 0x83, 0xf9, 0x04, 0x0b
+};
```
r-rojo: have you run this through clang-format? formatting looks weird

PR #1167, opened 2019-02-21
xfpga: allow enum. to account for device removal

r-rojo:
* Change xfpa_fpgaEnumerate to use sysfs_foreach_device
* Change sysfs_foreach_device to call
  * sysfs_finalize and sysfs_initialize
  * This in essense resets the discovered devices so that when this fn
    is called, it will start from scratch (and not allow for stale
    internal sysf data structures)
  * This is also protected by a mutex so it will be close to a guarantee
    that none of the devices discovered in one call will be stale -
    there's still a small race condition if a device goes away while
    this function is executing.
* Add remove_sysfs_dir to test_system
  * This will recursively remove a directory under the mock sysfs tree
* Add a unit test to verify that subsequent calls for xfpga_fpgaEnumerate
  will reflect devices as represented under /sys/class/fpga(_region)
* Extend mock_opae_p template class to include a prefix template
  argument to specify if certain function it needs should be a plugin or
  an API function. This is needed to call fpgaDestroyToken or
  xfpga_fpgaDestroyToken.

### DIFF
--- a/libopae/plugins/xfpga/enum.c
+++ b/libopae/plugins/xfpga/enum.c
@@ -492,7 +492,6 @@ STATIC fpga_result enum_afu(const char *sysfspath, const char *name,
 	return FPGA_OK;
 }
 
-#define MAX_ENUM_ERRORS 16
 typedef struct _enum_region_ctx{
 	struct dev_list *list;
 	bool include_port;
--- a/libopae/plugins/xfpga/enum.c
+++ b/libopae/plugins/xfpga/enum.c
@@ -492,69 +492,58 @@ STATIC fpga_result enum_afu(const char *sysfspath, const char *name,
 	return FPGA_OK;
 }
 
+#define MAX_ENUM_ERRORS 16
+typedef struct _enum_region_ctx{
+	struct dev_list *list;
+	bool include_port;
+} enum_region_ctx;
+
+STATIC fpga_result enum_regions(const sysfs_fpga_device *device, void *context)
+{
+	enum_region_ctx *ctx = (enum_region_ctx*)context;
+	fpga_result result = FPGA_OK;
+	struct dev_list *pdev = add_dev(device->sysfs_path, "", ctx->list);
+	if (!pdev) {
+		FPGA_MSG("Failed to allocate device");
+		return FPGA_NO_MEMORY;
+	}
+	// Assign bus, function, device
+	// segment,device_id ,vendor_id
+	pdev->function = device->function;
+	pdev->segment = device->segment;
+	pdev->bus = device->bus;
+	pdev->device = device->device;
+	pdev->device_id = device->device_id;
+	pdev->vendor_id = device->vendor_id;
+
+	// Enum fme
+	if (device->fme) {
+		result = enum_fme(device->fme->sysfs_path,
+				  device->fme->sysfs_name, pdev);
+		if (result != FPGA_OK) {
+			FPGA_ERR("Failed to enum FME");
+			return FPGA_EXCEPTION;
+		}
+	}
+
+	// Enum port
+	if (device->port && ctx->include_port) {
+		result = enum_afu(device->port->sysfs_path,
+				  device->port->sysfs_name, pdev);
+		if (result != FPGA_OK) {
+			FPGA_ERR("Failed to enum PORT");
+			return FPGA_EXCEPTION;
+		}
+	}
+	return FPGA_OK;
+}
 
 STATIC fpga_result enum_fpga_region_resources(struct dev_list *list,
 				bool include_port)
 {
-	fpga_result result                 = FPGA_NOT_FOUND;
-	struct dev_list *pdev              = NULL;
-	int i                              = 0;
-	int device_count                   = 0 ;
-	const sysfs_fpga_device *device    = NULL;
+	enum_region_ctx ctx = {.list = list, .include_port = include_port};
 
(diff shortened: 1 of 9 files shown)


testing/mock/mock_opae.h
```diff
@@ -31,33 +31,46 @@
...
     system_->initialize();
     system_->prepare_syfs(platform_);
+    invalid_device_ = test_device::unknown();
     test_setup();
   }
 
-  virtual void TearDown() override {
+  virtual void DestroyTokens() {
+    std::string fn_name = plugin_prefix_ + "fpgaDestroyToken";
+    auto fn = reinterpret_cast<fpga_result (*)(fpga_token *)>(
```
tswhison: neat!

testing/mock/test_system.cpp
```diff
@@ -250,6 +252,22 @@ void test_system::remove_sysfs() {
...
+  (void)sb;
+  (void)ftw;
+  if (flag & FTW_DP) {
+    return rmdir(fpath);
+  }
+  return unlink(fpath);
+}
+}
+
+int test_system::remove_sysfs_dir(const char *path) {
```
tswhison: nice!

testing/xfpga/test_sysfs_c.cpp
```diff
@@ -179,6 +179,7 @@ TEST_P(sysfsinit_c_p, sysfs_initialize) {
         devs.erase(id);
       }
     }
+    return FPGA_OK;
```
jelonanX: What are the advantages of having a return for this CB func?

testing/xfpga/test_enum_c.cpp
```diff
@@ -994,6 +972,9 @@ TEST_P(enum_c_p, get_guid) {
   EXPECT_EQ(fpgaDestroyProperties(&prop), FPGA_OK);
 }
 
+
+
+
 INSTANTIATE_TEST_CASE_P(enum_c, enum_c_p, 
                         ::testing::ValuesIn(test_platform::platforms({})));
```
tswhison: mock_platorms() I think?

testing/xfpga/test_sysfs_c.cpp
```diff
@@ -179,6 +179,7 @@ TEST_P(sysfsinit_c_p, sysfs_initialize) {
...
   };
```
tswhison: extraneous ;

testing/xfpga/test_sysfs_c.cpp
r-rojo: This CB is a trivial one but I changed the signature of the callback to return an fpga_result as an indication in the loop of the `sysfs_foreach_device` function to abort the loop and return the result code.

testing/xfpga/test_enum_c.cpp
```diff
@@ -1058,3 +1039,36 @@ TEST_P(enum_socket_c_p, socket_id) {
...
+ *             and I get one token for that accelerator
+ *             When I remove the port device from the system
+ *             And I enumerate again with the same filter
+ *             Then I get zero tokens as the result.
+ *
+ */
+TEST_P(enum_mock_only, remove_port) {
+  fpga_properties filterp = NULL;
+
+  ASSERT_EQ(xfpga_fpgaGetProperties(NULL, &filterp), FPGA_OK);
```
ibarraza: Does filterp get cleaned up somewheres?

r-rojo: Good call. I've added it

testing/xfpga/test_sysfs_c.cpp
r-rojo: That's actually the end of assignment of the the lambda function to `cb` variable on line 172.

testing/xfpga/test_enum_c.cpp
r-rojo: I didn't change these tests so I'm not sure that they are mock only. I only added a new test that should be mock only.
This is instantiated on line 1074.

testing/xfpga/test_sysfs_c.cpp
tswhison: ah, my bad.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -374,17 +374,34 @@ int sysfs_device_count(void)
 	return count;
 }
 
-void sysfs_foreach_device(device_cb cb, void *context)
+fpga_result sysfs_foreach_device(device_cb cb, void *context)
 {
 	uint32_t i = 0;
 	int res = 0;
+	fpga_result result = FPGA_OK;
```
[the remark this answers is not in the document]
r-rojo: True, true

libopae/plugins/xfpga/enum.c
```diff
@@ -492,69 +492,58 @@ STATIC fpga_result enum_afu(const char *sysfspath, const char *name,
...
-				FPGA_ERR("Failed to enum PORT");
-				break;
-			}
+	// Enum port
+	if (device->port && ctx->include_port) {
+		result = enum_afu(device->port->sysfs_path,
+				  device->port->sysfs_name, pdev);
+		if (result != FPGA_OK) {
+			FPGA_ERR("Failed to enum PORT");
+			return FPGA_EXCEPTION;
```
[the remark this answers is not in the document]
r-rojo: fixed

PR #1176, opened 2019-02-28
dma.h - remove HW related entries from fpga_dma_properties

HananiaAmir:
Removing fields from fpga dma properties structure that are HW related and user does not care about

### DIFF
--- a/common/include/opae/dma.h
+++ b/common/include/opae/dma.h
@@ -52,12 +52,7 @@ extern "C" {
  * Output of fpgaDMAPropertiesGet.
  */
 typedef struct {
-	uint64_t max_channel_num; /**< Max number of channels that the DMA engine supports */
-	uint64_t max_ring_size;   /**< Max number of buffers that the DMA can hold */
-	uint64_t max_buffer_size; /**< Max size of one buffer */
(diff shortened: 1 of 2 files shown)


common/include/opae/dma.h
```diff
@@ -52,12 +52,7 @@ extern "C" {
  * Output of fpgaDMAPropertiesGet.
  */
 typedef struct {
-	uint64_t max_channel_num; /**< Max number of channels that the DMA engine supports */
-	uint64_t max_ring_size;   /**< Max number of buffers that the DMA can hold */
-	uint64_t max_buffer_size; /**< Max size of one buffer */
-	uint64_t addr_alignment_for_dma;    /**< Address alignment requirement for DMA */
-	uint64_t minimum_xfer_size_for_dma; /**< DMA tranfer size should be multiples of this size */
-	uint64_t capabilities_mask;         /**< Bit mask of fpga_dma_transfer_type */
+	uint64_t channel_num; /**< Number of channels that the DMA engine supports */
```
lzhan55: Why was the change?

HananiaAmir: I got the feedback that the user does not care / should not get HW related information.
it is really to keep things simple.

PR #1177, opened 2019-03-01
libbitstream: refactor

tswhison:
* make a clear distinction between legacy and current metadata formats.
* enable a framework for supporting multiple metadata versions using the same API.
New metadata formats can be supported without changing existing call signatures.
* enable metadata parse for files that contain multiple accelerator clusters.


testing/bitstream/test_bits_utils_c.cpp
```diff
@@ -0,0 +1,204 @@
...
+    system_->initialize();
+    system_->prepare_syfs(platform_);
+
+    j_root_ = nullptr;
+  }
+
+  virtual void TearDown() override {
+
+    if (j_root_)
+      json_object_put(j_root_);
```
jelonanX: nice func to know

libbitstream/bitstream.c
```diff
@@ -36,306 +36,279 @@
...
-		result = FPGA_OK;
-		goto out_free;
+	json_object *j_version = NULL;
+	enum json_tokener_error j_err = json_tokener_success;
+	void *parsed = NULL;
+
+        root = json_tokener_parse_verbose(metadata, &j_err);
+	if (!root) {
+		OPAE_ERR("invalid JSON metadata: %s",
+			 json_tokener_error_desc(j_err));
```
jelonanX: is json_tokener_success the correct enumerator here?

libbitstream/bitstream.c
```diff
@@ -36,306 +36,279 @@
...
-	json_len = *((uint32_t *)(bitstream + METADATA_GUID_LEN));
-	if (json_len == 0) {
-		OPAE_MSG("Bitstream has no metadata.");
-		result = FPGA_OK;
-		goto out_free;
+	json_object *j_version = NULL;
+	enum json_tokener_error j_err = json_tokener_success;
+	void *parsed = NULL;
+
+        root = json_tokener_parse_verbose(metadata, &j_err);
```
jelonanX: extra space

libbitstream/bitstream.c
tswhison: I think so, because the check is on !root.  So, if json_tokener_parse_verbose fails, then root will be NULL and j_err should be populated.

libbitstream/bitstream.c
tswhison: fixed

testing/bitstream/test_bitstream_c.cpp
```diff
@@ -50,34 +54,43 @@ fpga_result read_bitstream(const char *filename, opae_bitstream_info *info);
...
+  0x0c, 0x87,
+  0x53, 0x4b,
+  0x56, 0x7d, 0x4a, 0xf6, 0x93, 0xe9
+};
+const fpga_guid guid_reversed = {
+  0xe9, 0x93, 0xf6, 0x4a,
+  0x7d, 0x56,
+  0x4b, 0x53,
+  0x87, 0x0c,
+   0x3b, 0xcb, 0x1a, 0x3a, 0x7f, 0x02
```
jelonanX: extra leading space.

testing/bitstream/test_metadatav1_c.cpp
```diff
@@ -0,0 +1,885 @@
...
+                                         fpga_guid pr_interface_id);
+
+}
+
+#include <config.h>
+#include <opae/fpga.h>
+
+#include <fstream>
+#include <vector>
+
```
jelonanX: Probably don't need these headers.

testing/bitstream/test_bits_utils_c.cpp
```diff
@@ -0,0 +1,204 @@
...
+// POSSIBILITY OF SUCH DAMAGE.
+
+#include "bits_utils.h"
+
+#include <config.h>
+#include <opae/fpga.h>
+
+#include <fstream>
+#include <vector>
+
```
jelonanX: Probably don't need these vector and fstream headers.

libbitstream/bitstream.c
```diff
@@ -36,306 +36,279 @@
...
-	}
+	hdr = (opae_bitstream_header *)info->data;
 
-	/* open file */
-	f = fopen(filename, "rb");
-	if (!f) {
-		OPAE_ERR("fopen failed for \"%s\"", filename);
-		return res;
+	if (uuid_compare(hdr->valid_gbs_guid, valid_GBS_guid) != 0) {
+		OPAE_ERR("valid GBS guid mismatch: \"%s\"", info->filename);
```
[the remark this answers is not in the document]
tswhison: done

PR #1178, opened 2019-03-01
Lzhan55/ase enum mmio test

lzhan55:
Added enumeration and mmio tests. 

These tests are mostly negative tests. ASE's coverage tests need to run some sample tests.

### DIFF
--- a/ase/api/src/enum.c
+++ b/ase/api/src/enum.c
@@ -200,11 +200,8 @@ fpgaEnumerate(const fpga_properties *filters, uint32_t num_filters,
 		ase_token[i] = &aseToken[i];
 	*num_matches = 0;
 	for (i = 0; i < 2; i++) {
-		//if (matches_filters(filters, num_filters, (void**)&(&aseToken[i]), &i)) {
 		if (matches_filters(filters, num_filters, &ase_token[i], &i)) {
 			if (*num_matches < max_tokens)	{
-
-
 				if (FPGA_OK != fpgaCloneToken(ase_token[i], &tokens[*num_matches]))
 					FPGA_MSG("Error cloning token");
 			}
--- a/ase/api/src/enum.c
+++ b/ase/api/src/enum.c
@@ -196,7 +196,7 @@ fpgaEnumerate(const fpga_properties *filters, uint32_t num_filters,
 		ase_memcpy(&aseToken[1].accelerator_id, readback_afuid, sizeof(fpga_guid));
 	}
 
-	for (i=0; i<2; i++)
+	for (i = 0; i < 2; i++)
 		ase_token[i] = &aseToken[i];
 	*num_matches = 0;
 	for (i = 0; i < 2; i++) {
--- a/ase/api/src/enum.c
+++ b/ase/api/src/enum.c
@@ -95,7 +95,7 @@ struct dev_list {
 	struct dev_list *fme;
 };
 
-static bool
+/*STATIC*/ bool
 matches_filters(const fpga_properties *filter, uint32_t num_filter,
 		fpga_token *token, uint64_t *j)
 {
@@ -103,7 +103,7 @@ matches_filters(const fpga_properties *filter, uint32_t num_filter,
 	if (filter == NULL)
 		return true;
 	struct _fpga_properties *_filter = (struct _fpga_properties *)*filter;
-	struct _fpga_token *_tok = (struct _fpga_token *)token;
+	struct _fpga_token *_tok = (struct _fpga_token *)*token;
 	if (!num_filter)	// no filter == match everything
 		return true;
 
@@ -139,6 +139,7 @@ matches_filters(const fpga_properties *filter, uint32_t num_filter,
 		}
 		if (_filter->objtype != _tok->ase_objtype)
 			return false;
+
 		if (FIELD_VALID(_filter, FPGA_PROPERTY_GUID)) {
 			if (0 != memcmp(_tok->accelerator_id, _filter->guid,
 					sizeof(fpga_guid))) {
@@ -161,8 +162,10 @@ fpgaEnumerate(const fpga_properties *filters, uint32_t num_filters,
 	      uint32_t *num_matches)
 {
 	uint64_t i;
+	fpga_token ase_token[2];
 	aseToken[0].ase_objtype = FPGA_DEVICE;
 	aseToken[1].ase_objtype = FPGA_ACCELERATOR;
+
 	if ((num_filters > 0) && (NULL == (filters))) {
 		return FPGA_INVALID_PARAM;
 	}
@@ -194,13 +197,16 @@ fpgaEnumerate(const fpga_properties *filters, uint32_t num_filters,
 		ase_memcpy(&aseToken[1].accelerator_id, readback_afuid, sizeof(fpga_guid));
 	}
 
+	for (i=0; i<2; i++)
+		ase_token[i] = &aseToken[i];
 	*num_matches = 0;
 	for (i = 0; i < 2; i++) {
-		if (matches_filters(filters, num_filters, (void **)&aseToken[i], &i)) {
+		//if (matches_filters(filters, num_filters, (void**)&(&aseToken[i]), &i)) {
+		if (matches_filters(filters, num_filters, &ase_token[i], &i)) {
 			if (*num_matches < max_tokens)	{
 
 
-				if (FPGA_OK != fpgaCloneToken(&aseToken[i], &tokens[*num_matches]))
+				if (FPGA_OK != fpgaCloneToken(ase_token[i], &tokens[*num_matches]))
 					FPGA_MSG("Error cloning token");
 			}
 			++*num_matches;
--- a/ase/api/src/enum.c
+++ b/ase/api/src/enum.c
@@ -95,8 +95,7 @@ struct dev_list {
 	struct dev_list *fme;
 };
 
-/*STATIC*/ bool
-matches_filters(const fpga_properties *filter, uint32_t num_filter,
+bool matches_filters(const fpga_properties *filter, uint32_t num_filter,
 		fpga_token *token, uint64_t *j)
 {
 	uint32_t i;
--- a/testing/ase/test_enum_c_ase.cpp
+++ b/testing/ase/test_enum_c_ase.cpp
@@ -90,7 +90,8 @@ class enum_c_ase_p : public testing::Test {
 /**
  * @test       nullfilter
  *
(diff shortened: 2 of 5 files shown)


ase/api/src/enum.c
```diff
@@ -194,13 +196,16 @@ fpgaEnumerate(const fpga_properties *filters, uint32_t num_filters,
 		ase_memcpy(&aseToken[1].accelerator_id, readback_afuid, sizeof(fpga_guid));
 	}
 
+	for (i=0; i<2; i++)
+		ase_token[i] = &aseToken[i];
```
michael-adler: What's the point of this separate variable?

ase/api/src/enum.c
```diff
@@ -194,13 +196,16 @@ fpgaEnumerate(const fpga_properties *filters, uint32_t num_filters,
...
 	*num_matches = 0;
 	for (i = 0; i < 2; i++) {
-		if (matches_filters(filters, num_filters, (void **)&aseToken[i], &i)) {
+		//if (matches_filters(filters, num_filters, (void**)&(&aseToken[i]), &i)) {
```
michael-adler: Dead code.

ase/api/src/enum.c
lzhan55: ase_token's type is fpga_token which is the opaque type of void*. 
aseToken's type is the internal data type of struct _fpag_token. In matches_filters, the parameter passed in is "fpga_token *", so the type is void**. In previous code at line 105, the code should dereference the pointer and I found this bug during the negative unit tests for match_filter function. When I fix this bug, I need to change the code here.

testing/ase/test_enum_c_ase.cpp
```diff
@@ -0,0 +1,380 @@
...
+ *
+ * @brief      When the uint32_t *num_matches parameter to fpgaEnumerate is nullptr, the
+ *             function returns FPGA_INVALID_PARAM.
+ */
+TEST_F(enum_c_ase_p, nullmatches) {
+  uint32_t matches = 0;
+  EXPECT_EQ(fpgaEnumerate(&filter_, 1, &tok, 1, NULL),
+            FPGA_INVALID_PARAM);
+  EXPECT_EQ(
+      fpgaEnumerate(&filter_, 0, &tok, 1, &matches),
```
jelonanX: could be wrong but this looks like it should return FPGA_OK?

testing/ase/test_enum_c_ase.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+    fpga_properties filter_;
+    struct _fpga_token tok_;
+    fpga_token tok;
+};
+
+/**
+ * @test       nullfilter
+ *
+ * @brief      When the fpga_properties * parameter to fpgaEnumerate is nullptr, the
```
tswhison: NULL filters is acceptable, it's just that num_filters must also be 0.  The test content is correct, but the description could use an update.

testing/ase/test_enum_c_ase.cpp
tswhison: Agree.. why is this failing?

testing/ase/test_enum_c_ase.cpp
```diff
@@ -0,0 +1,380 @@
...
+      fpgaEnumerate(&filter_, 0, &tok, 1, &matches),
+      FPGA_INVALID_PARAM);
+}
+
+/**
+ * @test       nulltokens
+ *
+ * @brief      When the fpga_token* parameter to fpgaEnumerate is nullptr, the
+ *             function returns FPGA_INVALID_PARAM.
+ *
```
tswhison: NULL tokens parameter is a valid case.  That's how we query the number of tokens.  It's just that num_tokens must also be 0.  Test content is ok, but the description could be updated.

testing/ase/test_enum_c_ase.cpp
```diff
@@ -0,0 +1,380 @@
...
+ * @brief      Test internal function matches_filters(), the function returns
+ *             true.
+ */
+TEST_F(enum_c_ase_p, matches_filters_1) {
+  uint64_t j = 0;
+  struct _fpga_properties* _prop = (struct _fpga_properties*)filter_;
+
+  // Set the property valid field
+  SET_FIELD_VALID(_prop, FPGA_PROPERTY_FUNCTION);
+  _prop->function = 0xAE;
```
tswhison: Does this pass because the "bus" number for ASE is 0xAE?  Probably could use some comment re: that.

testing/ase/test_enum_c_ase.cpp
```diff
@@ -0,0 +1,380 @@
...
+ *
+ * @brief      Test internal function matches_filters(), the function returns
+ *             true.
+ */
+TEST_F(enum_c_ase_p, matches_filters_2) {
+  uint64_t j = 0;
+  struct _fpga_properties* _prop = (struct _fpga_properties*)filter_;
+
+  SET_FIELD_VALID(_prop, FPGA_PROPERTY_DEVICE);
+  _prop->device = 0xAE;
```
tswhison: Same as above - update comments.

testing/ase/test_enum_c_ase.cpp
```diff
@@ -0,0 +1,380 @@
...
+ *
+ * @brief      Test internal function matches_filters(), the function returns
+ *             true.
+ */
+TEST_F(enum_c_ase_p, matches_filters_3) {
+  uint64_t j = 0;
+  struct _fpga_properties* _prop = (struct _fpga_properties*)filter_;
+
+  _prop->valid_fields = 0;
+  EXPECT_EQ(matches_filters(&filter_, 1, &tok,  &j), true);
```
tswhison: I think you're testing the case where the given filter is empty, matching all devices.  Should update the comments..

testing/ase/test_enum_c_ase.cpp
```diff
@@ -0,0 +1,380 @@
...
+ *
+ * @brief      Test internal function matches_filters(), the function returns
+ *             false.
+ */
+TEST_F(enum_c_ase_p, matches_filters_4) {
+  uint64_t j = 0;
+  struct _fpga_properties* _prop = (struct _fpga_properties*)filter_;
+
+  SET_FIELD_VALID(_prop, FPGA_PROPERTY_PARENT);
+  _prop->parent = NULL;
```
tswhison: Update comments..

testing/ase/test_enum_c_ase.cpp
```diff
@@ -0,0 +1,380 @@
...
+  fpga_token tok_dev_;
+  tok_dev_ = &tok1_;
+  token_for_afu0(&tok1_);
+
+  struct _fpga_properties* _prop = (struct _fpga_properties*)filter_;
+
+  SET_FIELD_VALID(_prop, FPGA_PROPERTY_PARENT);
+  _prop->parent = tok_dev_;
+  
+  EXPECT_EQ(matches_filters(&filter_, 1, &tok,  &j), false);
```
tswhison: Not sure why this should fail?

testing/ase/test_enum_c_ase.cpp
```diff
@@ -0,0 +1,380 @@
...
+  SET_FIELD_VALID(_prop, FPGA_PROPERTY_PARENT);
+  _prop->parent = tok_dev_;
+  
+  EXPECT_EQ(matches_filters(&filter_, 1, &tok,  &j), false);  
+}
+
+/**
+ * @test       matches_filters_6
+ *
+ * @brief      Test internal function matches_filters(), the function returns
```
tswhison: Update test description to reflect specifics of test.  Why is it expected to return false?

testing/ase/test_enum_c_ase.cpp
lzhan55: Will fix the comments

testing/ase/test_enum_c_ase.cpp
lzhan55: When the filter is not NULL but the number of filter is zero, it will return FPGA_INVALID_PARAM.

testing/ase/test_enum_c_ase.cpp
lzhan55: Will fix the comment.

testing/ase/test_enum_c_ase.cpp
lzhan55: Will update the comments. ASE doesn't have the bus, so basically it is ignored.

testing/ase/test_enum_c_ase.cpp
lzhan55: OK

testing/ase/test_enum_c_ase.cpp
lzhan55: sure.

testing/ase/test_enum_c_ase.cpp
lzhan55: sure

testing/ase/test_enum_c_ase.cpp
lzhan55: This test set the wrong parent token. It should be a fme token not an accelerator token.

testing/ase/test_enum_c_ase.cpp
jelonanX: doesn't 0 mean match all FPGA resources?

jelonanX: nvm, 0 to match all FPGA resources when filters is NULL. But filter isn't.

lzhan55: At line 574 of libopae/api-shell.c, it said 	
        if ((num_filters > 0) && !filters) {
		OPAE_ERR("num_filters > 0 with NULL filters");
		return FPGA_INVALID_PARAM;
	}

tswhison: You're passing filter_ here, which I think is non-NULL.

lzhan55:
@michael-adler Can you see it now?

testing/ase/test_enum_c_ase.cpp
tswhison: > When the filter is not NULL but the number of filter is zero, it will return FPGA_INVALID_PARAM.
Ok, I see it now.  Yes, agree.  Please update the test comments to reflect this point.

lzhan55: Will update doc

testing/ase/test_enum_c_ase.cpp
lzhan55: done

PR #1180, opened 2019-03-01
add 'realpath' to mock

r-rojo:
I accidentally included a mock tarfile of a a system with a port device migrated to a VF.
This shouldn't affect the build. This will be used in a future PR that has tests for this.


testing/mock/test_system.cpp
```diff
@@ -736,6 +738,30 @@ int test_system::glob(const char *pattern, int flags,
   return res;
 }
 
+char *test_system::realpath(const char *inp, char *dst)
```
jelonanX: dst will need to be freed if it's NULL when mock test calls this. 👍

r-rojo: It's up to the caller what dst is. Typically it will be an array allocated on the stack.

PR #1181, opened 2019-03-02
xfpga: fix VF enumeration bug

r-rojo:
## Description of Bug ##

The current enumeration flow assumes that both fme and port nodes in
sysfs have the same parent (device) node. With this assumption, some
code related to either finding an fme node from a port node, or vice
versa, looks for the target node by looking in the current node's parent
directory.

For example, say you have a device with the following fme and port
nodes (using conceptual names instead of names specific to a specific
version of the OPAE kernel driver):
```
/sys/class/fpga/dev.0/fme.0
/sys/class/fpga/dev.0/port.0
```
Given a path to an fme node, the port node can be found by the following
glob pattern:
`/sys/class/fpga/dev.0/fme.0/../*port.*`

or looking for an fme node using the path to the port node:
`/sys/class/fpga/dev.0/port.0/../*fme.*`

### OPAE and SR-IOV ###

The paths change, however, when SR-IOV is enabled (a port device is
released) and a VF is created. The steps to do this would look something
like (again omitting driver specific node names like `intel-fpga`):
```bash
fpgaport release /dev/fme.0 0
echo 1 > /sys/class/fpga/dev.0/device/sriov_numvfs
```
The first command tells the fme to enable SR-IOV and release the port device.
When this is done, the sysfs nodes related to port.0 are removed.
The second command tells the device so create a VF (now that SR-IOV has been enabled).

The result sysfs paths for the fme and port nodes would now look
something like:
```
/sys/class/fpga/dev.0/fme.0
/sys/class/fpga/dev.1/port.1
```

The char device nodes now look like:
```
/dev/fme.0
/dev/port.1
```

## Description of Fix ##

This change fixes the assumption mentioned above and uses symlinks
(physfn or virtfn?) present when a port has been migrated to a VF.

Now, when looking for a port node given a path to an fme node, the code
looks at the following glob patterns:

```
/sys/class/fpga/dev.0/fme.0/../device/virtfn*/fpga*/dev.*/*port.*
/sys/class/fpga/dev.0/fme.0/../device/fpga*/dev.*/*port.*
```
_**Note**: Only one of the preceding glob patterns will exist at a time
depending on whether or not a port is on a VF (the first pattern) or on
a PF (the second pattern)._

Similarly, when looking for an fme node given a path to a port node, the
following patterns will be tried:

```
/sys/class/fpga/dev.0/port.0/../device/physfn/fpga*/dev.*/*fme.*
/sys/class/fpga/dev.0/port.0/../device/fpga*/dev.*/*fme.*
```

These changes were implemented in 
* sysfs_get_fme_path
  This function was changed to take in a path to a port device
(instead of the device number and the port number) and use the algorithm described
above to find an fme node from a port node.

* get_port_sysfs
  This function was also changed to use the algorithm described above to find
an port node from an fme node.

_**Note**: The functions, `sysfs_get_fme_path` and `get_port_sysfs` now return
a realpath (a path to /sys/devices/pci*/...) with no symlinks._Any paths compared
to the return values of these functions must apply realpath to the path before
doing the comparison.

## Testing ##

Finally, for testing, a mock sysfs file structure was added to to
represent a scenario when a port is on a VF. The `test_device` class and
the mock platform table had to be modified to include the number of VFs in
a given device. Several tests had to be modified to use this information
in some of their assertions.


libopae/plugins/xfpga/sysfs.c
```diff
@@ -1137,11 +1190,15 @@ fpga_result get_port_sysfs(fpga_handle handle, char *sysfs_port)
```
tswhison: Gets the port sysfs path, given a handle to the corresponding FME?  Maybe add a comment that describes this.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -1158,27 +1215,60 @@ fpga_result get_port_sysfs(fpga_handle handle, char *sysfs_port)
...
+
+	if (!fpga_globs[i]) {
+		FPGA_ERR("Could not find path to port device/fpga");
+		return FPGA_EXCEPTION;
+	}
+
+	// format a string to look for in the subdirectory of the "fpga*" node
+	// this subdirectory should include glob patterns for the current
+	// driver
+	// -- inte-fgga-dev.*/intel-fpga-port.*
```
tswhison: s/inte-fgga-dev/intel-fpga-dev/

testing/opae-c/test_buffer_c.cpp
```diff
@@ -54,8 +54,12 @@ class buffer_c_p : public mock_opae_p<2> {
     ASSERT_EQ(fpgaInitialize(NULL), FPGA_OK);
     ASSERT_EQ(fpgaGetProperties(nullptr, &filter_), FPGA_OK);
     ASSERT_EQ(fpgaPropertiesSetObjectType(filter_, FPGA_ACCELERATOR), FPGA_OK);
-    ASSERT_EQ(fpgaPropertiesSetDeviceID(filter_,
-              platform_.devices[0].device_id), FPGA_OK);
+    auto device_id = platform_.devices[0].device_id;
+    if (platform_.devices[0].num_vfs) {
+      device_id++;
```
tswhison: ie, 0xbcc0 become 0xbcc1.. but this behavior is not guaranteed I think.  I'm not sure this will always be the case.

testing/opae-c/test_enum_c.cpp
```diff
@@ -277,9 +277,19 @@ TEST_P(enum_c_p, function) {
   EXPECT_EQ(
       fpgaEnumerate(&filter_, 1, tokens_.data(), tokens_.size(), &num_matches_),
       FPGA_OK);
-  EXPECT_EQ(num_matches_, GetNumFpgas() * 2);
+  EXPECT_EQ(num_matches_, GetNumFpgas() * (device.num_vfs == 0 ? 2 : 1));
+  num_matches_ = 0;
+  for (int i = 1; i < device.num_vfs+1; ++i) {
+    ASSERT_EQ(fpgaPropertiesSetFunction(filter_, i), FPGA_OK);
+    EXPECT_EQ(
+        fpgaEnumerate(&filter_, 1, tokens_.data(), tokens_.size(), &num_matches_),
```
tswhison: Isn't this loop bleeding tokens?  ie, each time through tokens_.data() is set to a cloned token.  Each iteration after the first will leak the token.

testing/opae-c/test_enum_c.cpp
```diff
@@ -316,9 +329,19 @@ TEST_P(enum_c_p, device_id) {
...
       FPGA_OK);
-  EXPECT_EQ(num_matches_, GetNumDeviceID() * 2);
+  EXPECT_EQ(num_matches_, GetNumDeviceID() * (device.num_vfs == 0 ? 2 : 1));
 
-  DestroyTokens();
+  num_matches_ = 0;
+  for (int i = 1; i < device.num_vfs+1; ++i) {
+    ASSERT_EQ(fpgaPropertiesSetDeviceID(filter_, device.device_id+i), FPGA_OK);
+    EXPECT_EQ(
+        fpgaEnumerate(&filter_, 1, tokens_.data(), tokens_.size(), &num_matches_),
```
tswhison: Isn't this loop bleeding tokens?  ie, each time through tokens_.data() is set to a cloned token.  Each iteration after the first will leak the token.

testing/xfpga/test_enum_c.cpp
```diff
@@ -355,7 +356,15 @@ TEST_P(enum_c_p, function) {
   EXPECT_EQ(
       xfpga_fpgaEnumerate(&filter_, 1, tokens_.data(), tokens_.size(), &num_matches_),
       FPGA_OK);
-  EXPECT_EQ(num_matches_, GetNumFpgas() * 2);
+  EXPECT_EQ(num_matches_, GetNumFpgas() * 2 - device.num_vfs);
+  for (int i = 0; i < device.num_vfs; ++i) {
+    num_matches_ = 0;
+    ASSERT_EQ(fpgaPropertiesSetFunction(filter_, device.function+i), FPGA_OK);
+    EXPECT_EQ(xfpga_fpgaEnumerate(&filter_, 1, tokens_.data(), tokens_.size(),
```
tswhison: Isn't this loop bleeding tokens?  ie, each time through tokens_.data() is set to a cloned token.  Each iteration after the first will leak the token.

testing/xfpga/test_enum_c.cpp
```diff
@@ -433,7 +442,16 @@ TEST_P(enum_c_p, device_id) {
   EXPECT_EQ(
       xfpga_fpgaEnumerate(&filter_, 1, tokens_.data(), tokens_.size(), &num_matches_),
       FPGA_OK);
-  EXPECT_EQ(num_matches_, platform_.devices.size() * 2);
+  EXPECT_EQ(num_matches_, platform_.devices.size() * 2 - device.num_vfs);
+
+  for (int i = 0; i < device.num_vfs; ++i) {
+    num_matches_ = 0;
+    ASSERT_EQ(fpgaPropertiesSetDeviceID(filter_, device.device_id+i), FPGA_OK);
+    EXPECT_EQ(xfpga_fpgaEnumerate(&filter_, 1, tokens_.data(), tokens_.size(),
```
tswhison: Isn't this loop bleeding tokens?  ie, each time through tokens_.data() is set to a cloned token.  Each iteration after the first will leak the token.

testing/xfpga/test_properties_c.cpp
```diff
@@ -53,8 +58,8 @@ class properties_c_p : public ::testing::TestWithParam<std::string> {
     system_->prepare_syfs(platform_);
     ASSERT_EQ(xfpga_plugin_initialize(), FPGA_OK);
     ASSERT_EQ(xfpga_fpgaGetProperties(nullptr, &filter_accel_), FPGA_OK);
-    ASSERT_EQ(fpgaPropertiesSetDeviceID(filter_accel_, 
-                                        platform_.devices[0].device_id), FPGA_OK);
+    auto accel_dev_id = platform_.devices[0].device_id + platform_.devices[0].num_vfs;
```
tswhison: Not convinced this will always be true.

testing/xfpga/test_sysfs_c.cpp
```diff
@@ -186,11 +193,17 @@ TEST_P(sysfsinit_c_p, sysfs_initialize) {
   for (const auto &d : platform_.devices) {
     devices[to_uint32(d.segment, d.bus, d.device, d.function)] = d;
   }
-
+  if (platform_.devices[0].num_vfs) {
+    auto d = platform_.devices[0];
+    d.function = 1;
+    d.device_id++;
```
tswhison: Not convinced this will always be true.

testing/xfpga/test_usrclk_c.cpp
```diff
@@ -80,8 +80,8 @@ class usrclk_c
     ASSERT_GT(num_matches_, 0);
 
     ASSERT_EQ(xfpga_fpgaGetProperties(nullptr, &filter_accel_), FPGA_OK);
-    ASSERT_EQ(fpgaPropertiesSetDeviceID(filter_accel_,
-                                        platform_.devices[0].device_id), FPGA_OK);
+    auto devid = platform_.devices[0].device_id + platform_.devices[0].num_vfs;
```
tswhison: Not convinced this will always be the case.

jelonanX:
test_xfpga_token_list_c and test_xfpga_sysfs_c are failing on HW ncf37 and bmp12.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -596,35 +612,72 @@ int sysfs_filter(const struct dirent *de)
...
+		} else if (result != FPGA_NOT_FOUND) {
+			return result;
+		}
+	}
 
-	int len = snprintf_s_ss(sysfs_path, SYSFS_PATH_MAX, "%s/%s",
-		SYSFS_FORMAT(sysfs_class_path), SYSFS_FME_PATH);
-	if (len < 0) {
+	if (!fpga_globs[i]) {
+		FPGA_ERR("Could not find path to port device/fpga*");
```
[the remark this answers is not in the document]
r-rojo: Yes, we are looking for device/fpga* under the port path

libopae/plugins/xfpga/sysfs.c
```diff
@@ -596,35 +612,72 @@ int sysfs_filter(const struct dirent *de)
 	return de->d_name[0] != '.';
 }
 
-fpga_result sysfs_get_fme_path(int dev, int subdev, char *path)
+fpga_result sysfs_get_fme_path(const char *rpath, char *sysfs_fme)
```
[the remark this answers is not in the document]
r-rojo: Will do

libopae/plugins/xfpga/enum.c
```diff
@@ -430,14 +412,15 @@ STATIC fpga_result enum_fme(const char *sysfspath, const char *name,
...
+	int resval = 0;
 	struct stat stats;
 	struct dev_list *pdev;
 	char spath[SYSFS_PATH_MAX];
 	char dpath[DEV_PATH_MAX];
-
+	uint64_t value = 0;
 	// Make sure it's a directory.
 	if (stat(sysfspath, &stats) != 0) {
 		FPGA_MSG("stat failed: %s", strerror(errno));
```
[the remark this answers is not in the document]
r-rojo: You're probably right (but it wasn't my original code to begin with). I'll change it anyway.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -596,35 +612,72 @@ int sysfs_filter(const struct dirent *de)
...
+	if (!fpga_globs[i]) {
+		FPGA_ERR("Could not find path to port device/fpga*");
+		return FPGA_EXCEPTION;
+	}
+
+
+	// format a string to look for in the subdirectory of the "fpga*" node
+	// this subdirectory should include glob patterns for the current
+	// driver
+	// -- inte-fgga-dev.*/intel-fpga-fme.*
```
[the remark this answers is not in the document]
r-rojo: will fix

libopae/plugins/xfpga/sysfs.c
r-rojo: 👍

libopae/plugins/xfpga/sysfs.c
r-rojo: 👍

testing/opae-c/test_buffer_c.cpp
r-rojo: You're right. This will work for now but fixing this the right way would require more work to the testing hw tables and data structures.

testing/opae-c/test_enum_c.cpp
r-rojo: good catch 😄

testing/xfpga/test_enum_c.cpp
r-rojo: good catch 😄

testing/xfpga/test_properties_c.cpp
r-rojo: You're right. This will work for now but fixing this the right way would require more work to the testing hw tables and data structures.

testing/xfpga/test_sysfs_c.cpp
r-rojo: You're right. This will work for now but fixing this the right way would require more work to the testing hw tables and data structures.

testing/xfpga/test_usrclk_c.cpp
r-rojo: You're right. This will work for now but fixing this the right way would require more work to the testing hw tables and data structures.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -596,35 +612,84 @@ int sysfs_filter(const struct dirent *de)
...
+	if (!fpga_globs[i]) {
+		FPGA_ERR("Could not find path to port device/fpga*");
+		return FPGA_NOT_FOUND;
+	}
+
+
+	// format a string to look for in the subdirectory of the "fpga*" node
+	// this subdirectory should include glob patterns for the current
+	// driver
+	// -- intel-fgga-dev.*/intel-fpga-fme.*
```
tswhison: s/fgga/fpga/

testing/opae-c/test_enum_c.cpp
tswhison: miss this one?

testing/xfpga/test_enum_c.cpp
tswhison: miss this one?

PR #1184, opened 2019-03-05
Update for ASE

kevinyuan:
doc/src/ase_userguide/ase_userguide.md
Changed: Update / fix description to run client, use with_ase cmd instead of LD_PRELOAD, which error out with the latest OPAE SDK.

ase/Makefile
I am not sure if older version of VCS has the same problem, but
VCS 2018.09-SP1-1 failed with the compilation, complaining about missing ase_top.
Added -top in the command line to resolve this issue.


doc/src/ase_userguide/ase_userguide.md
```diff
@@ -670,6 +670,10 @@ is set to ```/tmp/opae/ase/work```.  Then, start the software application.
     $ export LD_LIBRARY_PATH=/tmp/opae/myinst/lib/
     $ LD_PRELOAD=libopae-c-ase.so ./hello_fpga
 
+    # Run the application through with ```with_ase``` instead of ```LD_PRELOAD```
```
lzhan55: Both "LD_PRELOAD=libopae-c-ase.so ./hello_fpga"  and  "with_ase ./hello_fpga" should work. Could you change the comment here?

PR #1189, opened 2019-03-05
Fix pluginmgr test

jelonanX:
The previous fix in PR #1175 did not consider the case of building mock on HW. This PR turns dummy plugin and no cfg tests to parameter tests and address filter leak. 
Looking at the test case definition, they should be parameterized, right? @r-rojo


r-rojo:
dummy plugin should really be independent of any target platform.

PR #1192, opened 2019-03-06
Add speculative loads to CCI-P

michael-adler:
Speculative loads return an error to the AFU when the load fails, typically due to the page not being mapped. These speculative loads currently are used only by the MPF/VTP (virtual to physical) building block, though could be used by any address translation module in the future.


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=3145955

Changes reviewed by all relevant projects.

platforms/platform_if/rtl/device_if/ccip_if_pkg.sv
```diff
@@ -27,13 +27,13 @@
 //
 // ***************************************************************************
 
-// Date: 02/2/2016
-// Compliant with CCI-P spec v0.7
+// Date: 03/2019
+// Compliant with CCI-P spec v0.8
```
lzhan55: The CCI-P spec v0.8 is for S10?

michael-adler: No, I bumped the number up because I made a user-visible change. The first use is in MPF, which is platform-independent, but the change is potentially usable on future BBSs, so we put it in CCI-P.

PR #1193, opened 2019-03-07
fix bugs related to use of `realpath`

r-rojo:
Recent changes use `realpath` to get the realpath (without symlinks) to
sysfs nodes and uses that for comparing/identifying certain kinds of
nodes in the OPAE sysfs tree. According to the man page of [realpath](http://man7.org/linux/man-pages/man3/realpath.3.html),
there are bugs and it states that "it is impossible to determine a
suitable size for the output buffer".

This change assures that the buffers that are ultimately used as the
output variable to the call to realpath are of size PATH_MAX
(4096 as defined limits.h).

It has also been observed that calling realpath with a buffer smaller
than PATH_MAX can overflow the buffer even though the lengths of the
resolved paths are ~128.

### DIFF
--- a/libopae/plugins/xfpga/reconf.c
+++ b/libopae/plugins/xfpga/reconf.c
@@ -32,6 +32,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <ctype.h>
+#include <limits.h>
 #include <sys/types.h>
 
 #include "safe_string/safe_string.h"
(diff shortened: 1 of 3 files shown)


libopae/plugins/xfpga/reconf.c
```diff
@@ -32,6 +32,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <ctype.h>
+#include <limits.h>
```
jelonanX: Is this needed?

PR #1194, opened 2019-03-07
fix samples and add a note to doxygen of c_type fn

r-rojo:
Fix the C++ sample for hello_fpga-1.cpp to use explicit types instead of
the `auto` keyword. This is done to make it clearer as to the types of
objects being created.

Add a note in shared_buffer::c_type doxygen markup to warn developers of
the dangers of _stealing_ a raw pointer from the shared_buffer object.

### DIFF
--- a/common/include/opae/cxx/core/shared_buffer.h
+++ b/common/include/opae/cxx/core/shared_buffer.h
@@ -87,6 +87,10 @@ class shared_buffer {
   void release();
 
   /** Retrieve the virtual address of the buffer base.
+   *
+   *  @note assiging this to a variable should be done in limited scopes
+   *  as the variable can outlive the shared_buffer object creating the danger
+   *  of a dangling pointer
    */
   volatile uint8_t *c_type() const { return virt_; }
 
--- a/common/include/opae/cxx/core/shared_buffer.h
+++ b/common/include/opae/cxx/core/shared_buffer.h
@@ -88,9 +88,17 @@ class shared_buffer {
 
   /** Retrieve the virtual address of the buffer base.
    *
-   *  @note assiging this to a variable should be done in limited scopes
-   *  as the variable can outlive the shared_buffer object creating the danger
-   *  of a dangling pointer
+   *  @note Instances of a shared buffer can only be created using either
+   *  'allocate' or 'attach' static factory function. Because these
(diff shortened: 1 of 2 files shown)


common/include/opae/cxx/core/shared_buffer.h
```diff
@@ -87,6 +87,10 @@ class shared_buffer {
   void release();
 
   /** Retrieve the virtual address of the buffer base.
+   *
+   *  @note assiging this to a variable should be done in limited scopes
```
tswhison: s/assiging/assigning/

common/include/opae/cxx/core/shared_buffer.h
```diff
@@ -87,6 +87,10 @@ class shared_buffer {
...
+   *  as the variable can outlive the shared_buffer object creating the danger
+   *  of a dangling pointer
```
tswhison: period at end?

michael-adler:
Your note will be interpreted well by C++ experts, but I'm less convinced that casual users will understand the implications. How about explaining that the the buffer pointer is a smart pointer, that the destructor is called automatically when the last reference is deleted (e.g. when leaving the current scope) and that the pointers that result from reinterpret_cast are not smart and not part of the reference count?

common/include/opae/cxx/core/shared_buffer.h
```diff
@@ -88,9 +88,17 @@ class shared_buffer {
 
   /** Retrieve the virtual address of the buffer base.
    *
-   *  @note assiging this to a variable should be done in limited scopes
-   *  as the variable can outlive the shared_buffer object creating the danger
-   *  of a dangling pointer
+   *  @note Instances of a shared buffer can only be created using either
```
tswhison: maybe an example to clarify/drive home the point?

r-rojo: Example here? The cxx hello fpga isn't enough?

tswhison: Maybe offer up the issue submitter's code as an example of de-referencing the invalid pointer after the scope of the shared_ptr..  Having seen the code, it was clear to me what you were describing.  I'm concerned that others might not get the point by only reading the prose.

PR #1195, opened 2019-03-08
Add capability masks for CCI-P speculative reads.

michael-adler:


michael-adler:
I forgot to add this with the commit yesterday. They aren't used anywhere yet but are needed for MPF.

Confirmed that it compiles.

PR #1198, opened 2019-03-12
Lzhan55/ase unit tests3

lzhan55:
1. Removed deprecated 'tests' directory. OPAE unit tests are under 'testing' directory.
2. Removed some dead code in ASE simulation code
3. Added some unit tests to ASE api and sw code
4. Added ASE coverage scripts


ase/api/src/wsid_list.c
```diff
@@ -209,31 +209,3 @@ struct wsid_map *wsid_find(struct wsid_tracker *root, uint64_t wsid)
...
 
-/**
- * @ brief Find entry in linked list
- *
- * @param root
- * @param index
- *
- * @return
- */
-struct wsid_map *wsid_find_by_index(struct wsid_tracker *root, uint32_t index)
```
michael-adler: You're sort of going the opposite direction we want here. I know this function isn't used in ASE, but the point was that the file is as close as possible to to the OPAE source. At one point, I think it _was_ identical. Aren't we trying, ultimately, to share sources with OPAE instead of making ASE an OPAE reimplementation?

CMakeLists.txt
```diff
@@ -180,20 +180,13 @@ add_subdirectory(samples)
 option(BUILD_TESTS "Enable tests compilation" OFF)
 mark_as_advanced(BUILD_TESTS)
 
-option(BUILD_ASE_TESTS "Enable ASE test compilation." OFF)
-mark_as_advanced(BUILD_ASE_TESTS)
-
 option(ENABLE_MOCK "Enable mock driver for testing." ON)
 mark_as_advanced(ENABLE_MOCK)
 
-if(BUILD_TESTS OR BUILD_ASE_TESTS)
```
zhangzhang10: You removed BUILD_ASE_TESTS option. Does this mean ASE tests will always be built whenever BUILD_TESTS is ON? Should we check if BUILD_ASE is ON before building ASE tests?

ase/api/src/wsid_list.c
lzhan55: This function is dead code in ASE and I don't know whether it is used in OPAE. Will check.

ASE is going to be implemented as a plugin of OPAE like xfpga. ASE can have its own internal implementation, but OPAE API will be the same.

michael-adler: It's used in OPAE for MMIO.

Wouldn't it be a better goal to share OPAE sources whenever possible instead of just acting as a separate plugin? The fewer opportunities for divergence, the better.

CMakeLists.txt
lzhan55: BUILD_ASE_TESTS option was to build ASE gtest from tests folder. Actually I am going to remove the tests folder from OPAE repo as well. You can build ASE tests with build target test_ase in the updated scripts/cover.sh.

ase/api/src/wsid_list.c
lzhan55: Currently it is not used in ASE so we should remove it. If we need it in the future, we can easily add it back. 

ASE will be added as a plugin after I complete the ASE unit tests. So there will be three plugins for OPAE: xfpga, remote fpga and ASE. I think it is good to integrate ASE as part of OPAE since it will share the same libopae-c (FPGA API) with other plugins while it still keeps its own implementation since RTL simulation is totally different from the actual FPGA hardware implementation.

testing/ase/test_api_c_ase.cpp
```diff
@@ -0,0 +1,214 @@
...
+	prop.magic = 0xFFFFFFFF;
+	EXPECT_EQ(FPGA_INVALID_PARAM, prop_check_and_lock(&prop));
+
+
+}
+
+/**
+* @test       ase_common_02
+*
+* @brief      When the property's magic is invalid and libopae-ase-c is loaded:
```
tswhison: s/property/handle/

testing/ase/test_api_c_ase.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+	handle.magic = 0xFFFFFFFF;
+	handle.lock = PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP;
+	EXPECT_EQ(FPGA_INVALID_PARAM, handle_check_and_lock(&handle));
+}
+
+/**
+* @test       ase_common_03
+*
+* @brief      When the property's magic is invalid and libopae-ase-c is loaded:
```
tswhison: s/property/event_handle/

testing/ase/test_api_c_ase.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * @test       ase_wsid_1
+ *
+ * @brief      When the parameter n_hash_buckets is zero or greater than 16384, 
+ *             the function returns NULL. 
+ */
+TEST(open_c_ase, ase_wsid_1) {
+	EXPECT_EQ(NULL, wsid_tracker_init(0));
+	EXPECT_EQ(NULL, wsid_tracker_init(16385));
```
tswhison: Is the limit implemented as a #define somewhere that we can use, ie
```c
EXPECT_EQ(NULL, wsid_tracker_init(MAX_WSID_TRACKER_BUCKETS+1));
```

testing/ase/test_api_c_ase.cpp
```diff
@@ -0,0 +1,214 @@
...
+	ase_capability.umsg_feature = 0;
+	EXPECT_EQ(FPGA_NOT_SUPPORTED, fpgaGetNumUmsg(accel_, &value));
+}
+
+/**
+ * @test       ase_umsg_2
+ *
+ * @brief      When ase_capability's umsg feature is zero, 
+ *             fpgaSetUmsgAttributes function returns FPGA_NOT_SUPPORTED. 
+ *             When ase_capability's umsg feature is nonzero,
```
tswhison: This scenario isn't being tested.  Should it be added?

testing/ase/test_open_c_ase.cpp
```diff
@@ -129,6 +129,19 @@ TEST_F(open_c_ase_p, ase_open_04) {
     EXPECT_EQ(FPGA_INVALID_PARAM, fpgaOpen(tok, &accel_, 42));
 }
 
+/**
+* @test    open_05
```
tswhison: ase_open_05

testing/ase/test_open_c_ase.cpp
```diff
@@ -142,3 +155,14 @@ TEST_F(open_c_ase_p, mallocfails) {
     EXPECT_EQ(accel_, nullptr);
 }
 
+/**
+ * @test       close_nullhandle
+ *
+ * @brief      When the malloc function called by fpgaOpen() failed, the function returns
```
tswhison: Description doesn't match scenario under test.

testing/ase/test_sw_c_ase.cpp
```diff
@@ -106,54 +104,19 @@ TEST(LibopaecAseOps, ase_ops_01) {
 * @test       ase_ops_02
 *
 * @brief      When the parameters are valid and libopae-ase-c is loaded:
-*             ase_malloc(size_t) function should be called
+*             ase_malloc(size_t) function should return NULL
```
tswhison: Not sure what this is testing.. malloc is failing inside ase_malloc because you've invalidated it with the mock hook.  If it's supposed to fail due to the size being too large, then we shouldn't use invalidate_malloc.

testing/ase/test_sw_c_ase.cpp
```diff
@@ -542,3 +490,34 @@ TEST(LibopaecAseOps, ase_app_01) {
...
+*
+* @brief      When the parameters are nullptr,
+*             is_directory function returns 0;
+*
+*/
+TEST(sim_sw_ase, ase_app_02) {
+	EXPECT_EQ(0, is_directory(nullptr));
+}
+
+/**
```
tswhison: Move lines 504-510 to line 515, or else doxygen will not render correctly for this test.

testing/ase/test_sw_c_ase.cpp
```diff
@@ -542,3 +490,34 @@ TEST(LibopaecAseOps, ase_app_01) {
...
+* @test       ase_app_03
+*
+* @brief      When the parameters of va is bigger than 48
```
tswhison: Can you expand a bit more.  Why does passing 48 cause it to fail?  Also, the description currently says "bigger than 48", but you're passing 48.

testing/ase/test_api_c_ase.cpp
lzhan55: Added #define

testing/ase/test_api_c_ase.cpp
lzhan55: Will remove these unsupported feature tests.

testing/ase/test_open_c_ase.cpp
lzhan55: Will update.

testing/ase/test_open_c_ase.cpp
lzhan55: Will update

testing/ase/test_sw_c_ase.cpp
lzhan55: Agree. Will change it.

testing/ase/test_sw_c_ase.cpp
lzhan55: Will update comment

PR #1199, opened 2019-03-12
coreidle: switch coreidle app to libbitstream

tswhison:


tools/extra/coreidle/CMakeLists.txt
```diff
@@ -24,18 +24,24 @@
...
 
-target_link_libraries(coreidle opae-c ${libjson-c_LIBRARIES} ${libuuid_LIBRARIES} pthread m)
+target_link_libraries(coreidle
+  safestr
+  bitstream
+  opae-c
+  ${libjson-c_LIBRARIES}
+  ${libuuid_LIBRARIES}
+  ${CMAKE_THREAD_LIBS_INIT}
+  m)
```
jelonanX: m?

tswhison: libm.so, the standard C math library.

PR #1200, opened 2019-03-12
Remove deceptive indentation

chipb:
Signed-off-by: Chip Bilbrey <chip.bilbrey@intel.com>


chipb:
@netsweng FYI

chipb:
Thanks, folks. Hopefully this means less svosbld emails to us.

PR #1201, opened 2019-03-13
opae-c: extend sysobject API, allow container

r-rojo:
* Add a new API level enumeration, fpga_sysobject_type, to indicate if
an object is a system attribute (one that can be read or written) or a
container (one that contains other objects)
* Add APIs to access the sysobject type and to get a contained object by
index
* Add opae_glob_paths internal function to help glob patterns that
yield multiple results
* Add make_sysfs_array internal function that creates a container object
of objects from a glob pattern that yields multiple results
* Add sysfs unit tests
* TODO: Add opae-c plugin implementation

### DIFF
--- a/common/include/opae/types_enum.h
+++ b/common/include/opae/types_enum.h
@@ -159,9 +159,9 @@ enum fpga_sysobject_flags {
 
 enum fpga_sysobject_type {
 	FPGA_OBJECT_CONTAINER =
-		(1u << 0), /** represents a group of objects */
+		(1u << 0), /**< Represents a group of objects */
 	FPGA_OBJECT_ATTRIBUTE =
-		(1u << 1) /** an object with an attribute value that can be
+		(1u << 1) /**< An object with an attribute value that can be
 			       read/written */
 };
 
--- a/common/include/opae/types_enum.h
+++ b/common/include/opae/types_enum.h
@@ -145,13 +145,25 @@ enum fpga_reconf_flags {
 };
 
 enum fpga_sysobject_flags {
-  FPGA_OBJECT_SYNC = (1u << 0), /**< Synchronize data from driver */
-  FPGA_OBJECT_GLOB = (1u << 1), /**< Treat names as glob expressions */
-  FPGA_OBJECT_RAW  = (1u << 2), /**< Read or write object data as raw bytes */
-  FPGA_OBJECT_RECURSE_ONE = (1u << 3), /**< Create subobjects one level down from containers */
-  FPGA_OBJECT_RECURSE_ALL = (1u << 4) /**< Create subobjects all levels from from containers */
+	FPGA_OBJECT_SYNC = (1u << 0), /**< Synchronize data from driver */
+	FPGA_OBJECT_GLOB = (1u << 1), /**< Treat names as glob expressions */
+	FPGA_OBJECT_RAW =
+		(1u << 2), /**< Read or write object data as raw bytes */
+	FPGA_OBJECT_RECURSE_ONE =
+		(1u
+		 << 3), /**< Create subobjects one level down from containers */
+	FPGA_OBJECT_RECURSE_ALL =
+		(1u
+		 << 4) /**< Create subobjects all levels from from containers */
 };
 
+enum fpga_sysobject_type {
+	FPGA_OBJECT_CONTAINER =
+		(1u << 0), /** represents a group of objects */
+	FPGA_OBJECT_ATTRIBUTE =
+		(1u << 1) /** an object with an attribute value that can be
+			       read/written */
+};
 
 /** fpga metrics types
 * opae defines power,thermal, performance counter
--- a/libopae/plugins/xfpga/sysfs.c
+++ b/libopae/plugins/xfpga/sysfs.c
@@ -140,6 +140,14 @@ static sysfs_fpga_device _devices[SYSFS_MAX_DEVICES];
 		}                                                              \
 	} while (0);
 
+#define FREE_IF(var)                                                           \
+	do {                                                                   \
+		if (var) {                                                     \
+			free(var);                                             \
+			var = NULL;                                            \
+		}                                                              \
+	} while (false);
+
 STATIC int parse_pcie_info(sysfs_fpga_device *device, char *buffer)
 {
 	char err[128] = {0};
@@ -1711,6 +1719,22 @@ struct _fpga_object *alloc_fpga_object(const char *sysfspath, const char *name)
 {
 	struct _fpga_object *obj = calloc(1, sizeof(struct _fpga_object));
 	if (obj) {
+		pthread_mutexattr_t mattr;
+		if (pthread_mutexattr_init(&mattr)) {
+			FPGA_ERR("pthread_mutexattr_init() failed");
+			goto out_err;
+		}
+		if (pthread_mutexattr_settype(&mattr,
+					      PTHREAD_MUTEX_RECURSIVE)) {
+			FPGA_ERR("pthread_mutexattr_settype() failed");
+			goto out_err;
+		}
+		if (pthread_mutex_init(&obj->lock, &mattr)) {
+			FPGA_ERR("pthread_mutex_init() failed");
+			goto out_err;
+		}
+
+		pthread_mutexattr_destroy(&mattr);
 		obj->handle = NULL;
 		obj->path = cstr_dup(sysfspath);
 		obj->name = cstr_dup(name);
@@ -1721,6 +1745,31 @@ struct _fpga_object *alloc_fpga_object(const char *sysfspath, const char *name)
 		obj->objects = NULL;
 	}
 	return obj;
+out_err:
+	if (obj) {
+		free(obj);
+		obj = NULL;
+	}
+	return obj;
+}
+
+fpga_result destroy_fpga_object(struct _fpga_object *obj)
+{
+	FREE_IF(obj->path);
+	FREE_IF(obj->name);
+	FREE_IF(obj->buffer);
+	while (obj->size && obj->objects) {
+		if (destroy_fpga_object((struct _fpga_object *)
+						obj->objects[--obj->size])) {
+			FPGA_ERR("Error freeing subobject");
+		}
+	}
+	FREE_IF(obj->objects);
+	if (pthread_mutex_destroy(&obj->lock)) {
+		FPGA_ERR("Error destroying mutex");
+	}
+	free(obj);
+	return FPGA_OK;
 }
 
 fpga_result opae_glob_path(char *path)
@@ -1904,11 +1953,9 @@ fpga_result make_sysfs_group(char *sysfspath, const char *name,
 	return FPGA_OK;
 
 out_free_group:
-	if (group->path)
-		free(group->path);
-	if (group->name)
-		free(group->name);
-	free(group);
+	if (destroy_fpga_object(group)) {
+		FPGA_ERR("Error destroying object");
+	}
 
 out_free_namelist:
 	while (n--)
@@ -1935,6 +1982,7 @@ fpga_result make_sysfs_array(char *sysfspath, const char *name,
 	array->objects = calloc(num_objects, sizeof(fpga_object));
 	if (!array->objects) {
 		FPGA_ERR("Error allocating memory for array of fpga_objects");
+		free(array);
 		return FPGA_NO_MEMORY;
 	}
 
@@ -1957,9 +2005,8 @@ fpga_result make_sysfs_array(char *sysfspath, const char *name,
 	*object = (fpga_object)array;
 	return res;
 out_err:
-	if (array->objects) {
-		free(array->objects);
-		array->objects = NULL;
+	if (destroy_fpga_object(array)) {
+		FPGA_ERR("Error destroying object");
 	}
 	return res;
 }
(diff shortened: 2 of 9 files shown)


common/include/opae/types_enum.h
```diff
@@ -145,13 +145,25 @@ enum fpga_reconf_flags {
...
+		(1u
+		 << 3), /**< Create subobjects one level down from containers */
+	FPGA_OBJECT_RECURSE_ALL =
+		(1u
+		 << 4) /**< Create subobjects all levels from from containers */
 };
 
+enum fpga_sysobject_type {
+	FPGA_OBJECT_CONTAINER =
+		(1u << 0), /** represents a group of objects */
```
tswhison: Doesn't the doxygen comment need a < ? ie
```c
/**<
```
Also applies to 164.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -1874,6 +1918,54 @@ fpga_result make_sysfs_group(char *sysfspath, const char *name,
...
+	char *oname = NULL;
+	if (!array) {
+		FPGA_ERR(
+			"Error allocating memory for container of fpga_objects");
+		return FPGA_NO_MEMORY;
+	}
+	array->objects = calloc(num_objects, sizeof(fpga_object));
+	if (!array->objects) {
+		FPGA_ERR("Error allocating memory for array of fpga_objects");
+		return FPGA_NO_MEMORY;
```
tswhison: Need to free up array here, or else it is leaked.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -1874,6 +1918,54 @@ fpga_result make_sysfs_group(char *sysfspath, const char *name,
...
+		}
+	}
+	*object = (fpga_object)array;
+	return res;
+out_err:
+	if (array->objects) {
+		free(array->objects);
+		array->objects = NULL;
+	}
+	return res;
```
tswhison: Need to free up array here, else it is leaked.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -1874,6 +1918,54 @@ fpga_result make_sysfs_group(char *sysfspath, const char *name,
...
+					flags & ~FPGA_OBJECT_GLOB, handle);
+		if (res) {
+			goto out_err;
+		}
+	}
+	*object = (fpga_object)array;
+	return res;
+out_err:
+	if (array->objects) {
+		free(array->objects);
```
tswhison: Is there any memory allocation in each array->objects[i] that needs to be freed?

libopae/plugins/xfpga/sysobject.c
```diff
@@ -119,6 +120,23 @@ fpga_result __FPGA_API__ xfpga_fpgaObjectGetObject(fpga_object parent, const cha
...
 }
 
+fpga_result __FPGA_API__ xfpga_fpgaObjectGetObjectAt(fpga_object parent,
+						     size_t idx,
+						     fpga_object *object)
+{
+	ASSERT_NOT_NULL(parent);
+	ASSERT_NOT_NULL(object);
+	struct _fpga_object *_obj = (struct _fpga_object *)parent;
+	if (_obj->type == FPGA_SYSFS_FILE) {
```
tswhison: I think we should grab the lock inside the fpga_object when we are introspecting or modifying the object.  This would apply to each of the xfpga_ api's.

r-rojo: I'm not sure about getting child elements from a container type object since the number of items in the container won't change after the container object has been created. I agree with reading/writing from the driver though.

tswhison: Don't we need to prevent the object from being destroyed while we introspect..  I guess in that scenario, we would be accessing invalid pointers after the destruction occurred.  I think this is what the handles, etc, do though.  I'd feel safer about it if we followed that precedent throughout.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -139,6 +140,14 @@ static sysfs_fpga_device _devices[SYSFS_MAX_DEVICES];
 		}                                                              \
 	} while (0);
```
tswhison: nit - removing the ; at the end forces it to appear where the macro is expanded.  As written, it could be omitted.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -139,6 +140,14 @@ static sysfs_fpga_device _devices[SYSFS_MAX_DEVICES];
...
 
+#define FREE_IF(var)                                                           \
+	do {                                                                   \
+		if (var) {                                                     \
+			free(var);                                             \
+			var = NULL;                                            \
+		}                                                              \
+	} while (false);
```
tswhison: Use
```c
while (0)
```
to match the above.  Consider removing ; at end.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -1710,6 +1719,22 @@ struct _fpga_object *alloc_fpga_object(const char *sysfspath, const char *name)
...
 	struct _fpga_object *obj = calloc(1, sizeof(struct _fpga_object));
 	if (obj) {
+		pthread_mutexattr_t mattr;
+		if (pthread_mutexattr_init(&mattr)) {
+			FPGA_ERR("pthread_mutexattr_init() failed");
+			goto out_err;
+		}
+		if (pthread_mutexattr_settype(&mattr,
+					      PTHREAD_MUTEX_RECURSIVE)) {
+			FPGA_ERR("pthread_mutexattr_settype() failed");
```
tswhison: This path needs a pthread_mutexaddr_destroy.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -1710,6 +1719,22 @@ struct _fpga_object *alloc_fpga_object(const char *sysfspath, const char *name)
...
+			goto out_err;
+		}
+		if (pthread_mutex_init(&obj->lock, &mattr)) {
+			FPGA_ERR("pthread_mutex_init() failed");
```
tswhison: This path needs a pthread_mutexaddr_destroy.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -1720,6 +1745,31 @@ struct _fpga_object *alloc_fpga_object(const char *sysfspath, const char *name)
...
+	FREE_IF(obj->name);
+	FREE_IF(obj->buffer);
+	while (obj->size && obj->objects) {
+		if (destroy_fpga_object((struct _fpga_object *)
+						obj->objects[--obj->size])) {
+			FPGA_ERR("Error freeing subobject");
+		}
+	}
+	FREE_IF(obj->objects);
+	if (pthread_mutex_destroy(&obj->lock)) {
```
tswhison: Can the mutex be locked when this is called / does it need to be unlocked?

libopae/plugins/xfpga/sysfs.c
```diff
@@ -1720,6 +1745,31 @@ struct _fpga_object *alloc_fpga_object(const char *sysfspath, const char *name)
...
+
+fpga_result destroy_fpga_object(struct _fpga_object *obj)
+{
+	FREE_IF(obj->path);
+	FREE_IF(obj->name);
+	FREE_IF(obj->buffer);
+	while (obj->size && obj->objects) {
+		if (destroy_fpga_object((struct _fpga_object *)
+						obj->objects[--obj->size])) {
+			FPGA_ERR("Error freeing subobject");
```
tswhison: Capture the returned result here / propagate it upwards.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -1720,6 +1745,31 @@ struct _fpga_object *alloc_fpga_object(const char *sysfspath, const char *name)
...
+		}
+	}
+	FREE_IF(obj->objects);
+	if (pthread_mutex_destroy(&obj->lock)) {
+		FPGA_ERR("Error destroying mutex");
+	}
```
tswhison: Propagate error upwards?

libopae/plugins/xfpga/sysfs.c
```diff
@@ -1860,11 +1953,9 @@ fpga_result make_sysfs_group(char *sysfspath, const char *name,
 	return FPGA_OK;
 
 out_free_group:
-	if (group->path)
-		free(group->path);
-	if (group->name)
-		free(group->name);
-	free(group);
+	if (destroy_fpga_object(group)) {
+		FPGA_ERR("Error destroying object");
```
tswhison: Catch error code / propagate it upwards?

libopae/plugins/xfpga/sysfs.c
tswhison: Shouldn't the call be to destroy_fpga_object?

libopae/plugins/xfpga/sysfs.c
tswhison: Catch res / propagate it upward?

libopae/plugins/xfpga/sysfs.c
```diff
@@ -1883,8 +2022,27 @@ fpga_result make_sysfs_object(char *sysfspath, const char *name,
 	struct stat objstat;
 	int statres;
 	fpga_result res = FPGA_OK;
+	max_path_t object_paths[MAX_SYSOBJECT_GLOB];
```
tswhison: consumes quite a lot of stack space.  Is this necessary?

libopae/plugins/xfpga/sysobject.c
```diff
@@ -119,26 +113,71 @@ fpga_result __FPGA_API__ xfpga_fpgaObjectGetObject(fpga_object parent, const cha
...
+	_dst->handle = _src->handle;
+	_dst->perm = _src->perm;
+	_dst->size = _src->size;
+	_dst->max_size = _src->max_size;
+	if (_src->type == FPGA_SYSFS_FILE) {
+		memcpy_s(_dst->buffer, _dst->max_size, _src->buffer,
+			 _src->max_size);
+	} else {
+		_dst->buffer = NULL;
+		_dst->objects = calloc(_src->size, sizeof(fpga_object));
```
tswhison: NULL check for failed calloc?

libopae/plugins/xfpga/sysfs.c
r-rojo: If we're in here, it was because of an error, I don't want to overwrite the error with another one.

libopae/plugins/xfpga/sysfs.c
r-rojo: If we're here it's because of an error. I don't want to overwrite or mask this error.

libopae/plugins/xfpga/sysfs.c
r-rojo: let me think about it...

libopae/plugins/xfpga/sysfs.c
tswhison: I think we should go on to free stuff and just remember the failed result to return at the end.

libopae/plugins/xfpga/sysfs.c
```diff
@@ -1756,6 +1812,56 @@ fpga_result opae_glob_path(char *path)
...
+
+	if (!globres) {
+		*num_found = pglob.gl_pathc;
+		to_copy = *num_found < found_max ? *num_found : found_max;
+		while (found && i < to_copy) {
+			found[i] = cstr_dup(pglob.gl_pathv[i]);
+			if (!found[i]) {
+				// we had an error duplicating the string
+				// undo what we've duplicated so far
+				while (i) {
```
tswhison: 👍

libopae/plugins/xfpga/sysobject.c
```diff
@@ -119,26 +113,86 @@ fpga_result __FPGA_API__ xfpga_fpgaObjectGetObject(fpga_object parent, const cha
...
+	}
+	if (idx >= _obj->size) {
+		return FPGA_INVALID_PARAM;
+	}
+	res = xfpga_fpgaCloneObject(_obj->objects[idx], object);
+	if (pthread_mutex_unlock(&_obj->lock)) {
+		FPGA_ERR("pthread_mutex_unlock() failed");
+	}
+	return res;
+
```
tswhison: Remove blank line.

libopae/plugins/xfpga/sysobject.c
```diff
@@ -266,3 +320,38 @@ fpga_result __FPGA_API__ xfpga_fpgaObjectWrite64(fpga_object obj,
...
+		res = FPGA_INVALID_PARAM;
+	}
+	return res;
+}
+
+fpga_result __FPGA_API__ xfpga_fpgaObjectGetName(fpga_object obj, char *name,
+						 size_t max_len)
+{
+	fpga_result res = FPGA_OK;
+	struct _fpga_object *_obj = (struct _fpga_object *)obj;
```
tswhison: Mutex?

libopae/plugins/xfpga/sysobject.c
```diff
@@ -266,3 +320,38 @@ fpga_result __FPGA_API__ xfpga_fpgaObjectWrite64(fpga_object obj,
 	}
 	return res;
 }
+
+fpga_result __FPGA_API__ xfpga_fpgaObjectGetType(fpga_object obj,
+						 enum fpga_sysobject_type *type)
+{
+	fpga_result res = FPGA_OK;
+	struct _fpga_object *_obj = (struct _fpga_object *)obj;
```
tswhison: Mutex?

PR #1212, opened 2019-03-20
Split previous commits 8a25f3936 into two parts: one for the changes …

lzhan55:
…files and the other for deleting tests folder

This PR is about the changed files


ase/api/src/open.c
```diff
@@ -92,7 +92,7 @@ fpga_result __FPGA_API__ fpgaOpen(fpga_token token, fpga_handle *handle, int fla
 	_handle->mmio_root = NULL;
 
 	// Init workspace table
-	_handle->wsid_root = wsid_tracker_init(16384);
+	_handle->wsid_root = wsid_tracker_init(MAX_WSID_TRACKER_BUCKETS);
```
michael-adler: MAX is the wrong prefix. This is just a static number of buckets to allocate in the hash table. A better name would be "NUM_WSID_TRACKER_BUCKETS".

lzhan55: Will fix it

PR #1215, opened 2019-03-21
Feature/ase tests2

lzhan55:
- Merge ASE tests into OPAE new test framework in 'testing' folder, removing OPAE old 'tests' folder.
- All the changes in this feature branches were reviewed


ase/api/src/common_int.h
```diff
@@ -126,4 +126,7 @@ fpga_result event_handle_check_and_lock(struct _fpga_event_handle *eh);
 void fpga_print(int loglevel, char *fmt, ...);
 struct _fpga_token *token_get_parent(struct _fpga_token *);
 fpga_result objectid_for_ase(uint64_t *object_id);
+void api_guid_to_fpga(uint64_t guidh, uint64_t guidl, uint8_t *guid);
+bool matches_filters(const fpga_properties *filter, uint32_t num_filter,
```
michael-adler: If you are going to start introducing new methods into the global namespace, they shouldn't be generic names that might clash with an application's methods. If an application happens to have a function named "matches_filters" there will now be a link-time error.

lzhan55: Good catch! I was planning to change this as well.

lzhan55: @michael-adler The fix was checked into feature/ase_tests2 branch.

lzhan55:
The changes were merged into master already.

PR #1219, opened 2019-03-26
libbitstream: allow some keys to be optional

tswhison:
Allow the following keys to be optional. Some GBS in current DCP
releases omit them:
"clock-frequency-high"
"clock-frequency-low"
"power"
"platform-name"


tswhison:
Hey @ibarraza, could you have a look at this one?

PR #1220, opened 2019-03-27
Improve ASE simulation of fpgaPrepareBuffer() failures.

michael-adler:
- When fpgaPrepareBuffer() is called in preallocated mode it should
  fail when there is no page mapped at the buffer address.
- MPF's new VTP speculative mode depends on this behavior.

### DIFF
--- a/ase/api/src/buffer.c
+++ b/ase/api/src/buffer.c
@@ -194,6 +194,74 @@ static fpga_result buffer_release(void *addr, uint64_t len)
 	return FPGA_OK;
 }
 
+
+#define MAPS_BUF_SZ 4096
+/*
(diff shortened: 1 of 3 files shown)


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=3215605

michael-adler:
The basic problem this solves:

When calling fpgaPrepareBuffer() in preallocated mode on a real FPGA, the driver verifies that the buffer is mapped. It returns an error when pages are not mapped at the requested address. ASE was not doing this. When passed an address it never failed, even when no memory was mapped at the address. The new code was repurposed from similar code in MPF. It confirms that a valid mapping of the appropriate size is present.

PR #1222, opened 2019-03-28
libbitstream: allow fields to be int or double

tswhison:
Change the type of the following fields to double and allow them
to appear in the JSON metadata as either int or double:
afu_image.clock_frequency_high
afu_image.clock_frequency_low
afu_image.power

### DIFF
--- a/testing/bitstream/test_bits_utils_c.cpp
+++ b/testing/bitstream/test_bits_utils_c.cpp
@@ -230,6 +230,31 @@ TEST_P(bits_utils_c_p, double_err1) {
 }
 
 /**
+ * @test       double_err2
+ * @brief      Test: opae_bitstream_get_json_double
+ * @details    If the given name exists,<br>
+ *             but isn't of type double,<br>
+ *             the fn returns FPGA_EXCEPTION.<br>
+ */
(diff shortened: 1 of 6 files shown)


testing/bitstream/test_bits_utils_c.cpp
```diff
@@ -180,6 +180,55 @@ TEST_P(bits_utils_c_p, int_err1) {
...
+  root = parse(mdata);
+  ASSERT_NE(root, nullptr);
+
+  EXPECT_EQ(opae_bitstream_get_json_double(root,
+					   "a",
+					   &value),
+            FPGA_EXCEPTION);
+  EXPECT_EQ(value, 0.0);
+}
+
```
r-rojo: Can we add a test for parsing something like:
```JSON
{
  "a": 42
}
```

I wonder if it is even necessary to call `opae_bitstream_get_json_int` if the get double fn fails.

tswhison: It appears to be.  Makes sense, because there is a separate json_type_X for double, ie both json_type_int and json_type_double exist.

PR #1229, opened 2019-04-02
xfpga: revert difference in BBS version

tswhison:
A recent change introduced a difference in the bitfield encoding
for the bitstream_id CSR for MCP vs. DCP. It has been discovered
that this difference was a bug in the CSR spreadsheets for DCP.
With this change, both MCP and DCP assume the same encoding for
the bitstream_id and determine the version in the same way.


tswhison:
Thanks all.  Let's keep this unmerged for now.  I'm still waiting back on confirmation from the design team.

PR #1237, opened 2019-04-08
fpgad-vc: adjust sensor trip points

tswhison:
Pull the high/low temperature trip points down/up by 5 degrees.
Pull the high/low power trip points down/up by 5%.
This is done to enable catching overages before the BMC takes action.

### DIFF
--- a/tools/base/fpgad/plugins/fpgad-vc/fpgad-vc.c
+++ b/tools/base/fpgad/plugins/fpgad-vc/fpgad-vc.c
@@ -228,9 +228,18 @@ STATIC fpga_result vc_sensor_get_u64(vc_sensor *sensor,
 	return res;
 }
 
+// The percentage by which we adjust the power trip
+// points so that we catch anomolies before the hw does.
+#define VC_PERCENT_ADJUST_PWR 5
+// The number of degrees by which we adjust the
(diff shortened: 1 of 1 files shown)


tools/base/fpgad/plugins/fpgad-vc/fpgad-vc.c
```diff
@@ -228,9 +228,18 @@ STATIC fpga_result vc_sensor_get_u64(vc_sensor *sensor,
 	return res;
 }
 
+// The percentage by which we adjust the power trip
+// points so that we catch anomolies before the hw does.
+#define VC_PERCENT_ADJUST_PWR 5
```
r-rojo: Why not express this as 0.05?
Or treat is as the not how much we subtract but as a percentage of the value?
```C
#define VC_ADJUSTED_PERCENT_PWR 0.95

//...

s->high_fatal *= VC_ADJUSTED_PERCENT_PWR;

```

tswhison: Both the values and the trip points are uint64_t's.  I just didn't want to go the floating point route.

tswhison:
@r-rojo I thought about the separate function approach, but it seemed like overkill to abstract two lines of code.

PR #1240, opened 2019-04-11
Enable pinning of pages that are mapped read-only.

michael-adler:
Without this, pages that are mapped read-only can not be pinned. This change, along with a kernel update, adds the ability to pin read-only pages.

Default behavior is unchanged. This patch, when run against an old kernel, will return an error from the pinning ioctl -- the same as before.


libopae/plugins/xfpga/buffer.c
```diff
@@ -158,6 +159,9 @@ fpga_result __FPGA_API__ xfpga_fpgaPrepareBuffer(fpga_handle handle, uint64_t le
 	bool preallocated = (flags & FPGA_BUF_PREALLOCATED);
 	bool quiet = (flags & FPGA_BUF_QUIET);
 
+	bool read_only = (flags & FPGA_BUF_READ_ONLY);
+	uint32_t map_flags = (read_only ? FPGA_DMA_TO_DEV : 0);
+
```
tswhison: I wonder whether we should prevent FPGA_BUF_PREALLOCATED | FPGA_BUF_READ_ONLY ?  Or, is there checking being done in the kernel that would prevent this sort of error?  I'm concerned about someone attempting to pass a pre-allocated buffer that is actually r/w.

michael-adler: I don't think so. A preallocated page (one managed by the application) can certainly be read-only. Imagine I've mapped a file read-only and want the FPGA to scan it.

Even a page that is writeable by the CPU may reasonably be mapped read-only by the FPGA if you want to protect it from FPGA-side writes. The flag causes the driver to configure the IOMMU protection.

PR #1243, opened 2019-04-17
VC: Address fpgad leaks in tests

tswhison:
* fpgad: fix mem leak for invalid GBS test
* Fixed uninitialized value in fpgad test.


tswhison:

PR #1244, opened 2019-04-18
libbitstream: remove fall-through switch case

tswhison:
Newer versions of gcc flag fall-through switch cases as warnings,
which become build errors, because we're using -Werror.


asgibbons:
How about we *not* duplicate code and use gcc's __attribute__ described [here](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough)?

tswhison:
There was an email chain on that, but it was decided against.. I'll forward to you.

PR #1245, opened 2019-04-19
libbitstream: enable fallthrough switch case

tswhison:
Add cmake C compiler check (C++ TBD) for -Werror=implicit-fallthrough=3.
With this option, placing a /* FALLTHROUGH */ comment in the switch
case in question avoids what would otherwise be a warning.

### DIFF
--- a/cmake/modules/compiler_config.cmake
+++ b/cmake/modules/compiler_config.cmake
@@ -50,6 +50,7 @@ endfunction(SET_CACHED_VARIABLE)
 # Default flags to compiler when build user-space programs.
 # Should come before enabling language.
 
+CHECK_C_COMPILER_FLAG(-Werror=implicit-fallthrough=3 COMPILER_SUPPORTS_IMPLICIT_FALLTHROUGH3)
 
 set(CMAKE_C_FLAGS_DEBUG            "-g -O0 -Wall -Wextra -Werror")
 set(CMAKE_CXX_FLAGS_DEBUG          "-g -O0 -Wall -Wextra -Werror")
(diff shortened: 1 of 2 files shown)


libbitstream/bitstream.c
```diff
@@ -177,17 +177,8 @@ STATIC void *opae_bitstream_parse_metadata(const char *metadata,
 	// version to 640/650 respectively.
 	// Allow 640/650 to serve as an alias for 1.
 	case 650:
```
asgibbons: Don't need a fallthrough here?

tswhison: It didn't seem to require it.  This was with gcc 8.x on Fedora 29.

cmake/modules/compiler_config.cmake
```diff
@@ -63,6 +64,13 @@ set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-g -O2 -Wall -Wextra -Werror")
 set(CMAKE_C_FLAGS_MINSIZEREL       "-Os -Wall -Wextra -Werror")
 set(CMAKE_CXX_FLAGS_MINSIZEREL     "-Os -Wall -Wextra -Werror")
 
+if(COMPILER_SUPPORTS_IMPLICIT_FALLTHROUGH3)
```
chipb: The GCC documentation indicates `-Wimplicit-fallthrough=3` is enabled when `-Wextra` is enabled (as it already seems to be on e.g. line 55 and 64). Can you drop this extra check?

tswhison: I tried it without, but it didn't seem to work.  The error message indicated that -Werror=implicit-fallthrough= (without the 3) was active.

chipb: Whoops. Sorry I missed your PR opening comment, but thanks for double-checking!

PR #1246, opened 2019-04-22
Make ASE memory management more like real FPGA pinned buffers.

michael-adler:
This can be valuable when managing pinning automatically, such as in MPF. For
normal simulation there is no visible difference.


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=3284712

PR #1248, opened 2019-04-26
Add C++ support for read-only pinned buffers.

michael-adler:


michael-adler:
@tswhison - now it's ready. Finally got the formatting right.

PR #1249, opened 2019-04-26
fpgad-vc: graceful shutdown models super-rsu flow

tswhison:
Update the method used to remove the device driver to that of super-rsu:
* Write 1 to the device's reset node.
* Write 1 to the remove node of the root bridge's child. (removes
all VC devices)

When rescanning,
* Write "on" to the power/control node of the root bridge's child.
* Write 1 to the rescan node of the root bridge's child.


tswhison:

PR #1254, opened 2019-05-07
fpgad-vc: don't write device reset node

tswhison:
Remove the code that was writing 1 to the root bridge device's
reset sysfsnode.
Fix the tripped sensor counter so that it correctly measures for
each device.
Remove extraneous cool down lock.


tswhison:

PR #1263, opened 2019-05-07
cherry-pick changes from feature/ase_plugin to feature/ase_plugin2 (#…

lzhan55:
…1252)

* ASE: adding ASE as a plugin.
* Lzhan55/ase plugin (#1236)
* Lzhan55/ase plugin2 (#1247)
* ASE: Added a libopae-c-ase.so to make ASE application backward compat… (#1250)

PR #1252, opened 2019-05-06
cherry-pick changes from feature/ase_plugin to feature/ase_plugin2

lzhan55:
* feature/ase_plugin2 was just branched off from master today
* cmake .. -DCPACK_GENERATOR=DEB && make package_deb passed

PR #1236, opened 2019-04-08
Lzhan55/ase plugin

lzhan55:
* Move the ase folder to libopae/plugin

PR #1265, opened 2019-05-07
Merge ASE feature branch feature/ase_plugin3 into master 

lzhan55:
…… (#1263)

* cherry-pick changes from feature/ase_plugin to feature/ase_plugin2 (#1252)

* ASE: adding ASE as a plugin.
* Lzhan55/ase plugin (#1236)
* Lzhan55/ase plugin2 (#1247)
* ASE: Added a libopae-c-ase.so to make ASE application backward compat… (#1250)

* Added back the missing ASE Makefile


libopae/plugins/ase/CMakeLists.txt
```diff
@@ -75,12 +75,17 @@ add_subdirectory(rtl)
...
-  ase_regress.sh)
+  ase_regress.sh
+  opae_ase.cfg)
 
 install(FILES ${EXTRA_ASE_FILES}
   DESTINATION ${ASE_INST_SHARE_DIR}
   COMPONENT aseextra)
 
+install(FILES opae_ase.cfg
+  DESTINATION $ENV{HOME}/.local/opae
```
nakulkorde: can this file not be installed in the home directory? is there a specific need for this file to be in the home directory? it creates problems when specifying file paths in the rpm specfile. @tswhison what do you think?

nakulkorde: I have updated the rpm specfile to incorporate the new changes but I wasn't able to get the home directory to be identified by the specfile.

tswhison: Ooh, I can see how the home directory would be problematic for the rpm spec file.  It's probably better that it go into /etc/opae for that reason alone.

michael-adler: It's installed in <install root>/share/opae/ase/ too. Is that sufficient and you can skip both .local and /etc?

michael-adler:
Sigh. It appears that gcc 4.5 and later don't work with our new libopae-c-ase strategy. They changed the default in the linker to -no-copy-dt-needed-entries, which causes links against just libopae-c-ase to fail because the functions now in libopae-c are not found. You can try this yourself using a simple example, like samples/tutorial/01_hello_world/sw in intel-fpga-bbb. With gcc 4.5 and later, the cci_hello_ase link fails. If you change the link step, instead, to:

`gcc -o cci_hello_ase obj/cci_hello.o -z noexecstack -z relro -z now -pie -luuid -Wl,-copy-dt-needed-entries -lopae-c-ase`

it works fine. Unfortunately, the whole point of this effort was to be transparent, so forcing a change in linker flags or library lists would defeat the purpose.

The only solution I can think if, so far, is to build libopae-c-ase from the same sources as libopae-c, adding the thunk you have now that puts the run in ASE mode. You would then remove the dependence of libopae-c-ase on libopae-c. It isn't ideal, but it would be easy to do and it would work.

Any better ideas?

libopae/plugins/ase/CMakeLists.txt
lzhan55: Sure, I will remove it from $HOME.

lzhan55:
@michael-adler Agree with you.

michael-adler:
@deepakcu84 - Have you tried this branch on DCP regressions?

deepakcu84:
well, sorry no. I'll try it.

michael-adler:
You should wait until tracy fixes the problem I noted this morning, described above.

deepakcu84:
Got it. Sorry, I wasn't paying attention.

tswhison:
The coverage number decreased a bit.  Could you add some unit tests for find_ase_cfg( ) and the new init code?

lzhan55:
@tswhison Will add unit tests for that.

michael-adler:
@deepakcu84 - I think it's a good time to test now.

michael-adler:
What’s the plan for this merge? It is starting to get in the way of other changes.

lzhan55:
@michael-adler Amir will take over the ASE work. It could not be merged into master because DCP regression tests have not been run yet. Deepak is working on running the DCP regression tests.

lzhan55:
@michael-adler Thanks for running the DCP regression tests! It is great to see that the DCP regression tests passed with ASE plugins changes.

@nakulkorde Was Amir added as ASE code owner? If yes, Amir could approve the pull requests as ASE owner.

deepakcu84:
Thank you Michael.

PR #1268, opened 2019-05-08
opae: cherry-pick code scanner fix from vc

anandaravuri:


libopae/plugins/xfpga/metrics/afu_metrics.c
```diff
@@ -204,8 +204,8 @@ fpga_result add_afu_metrics_vector(fpga_metric_vector *vector,
 
 	*metric_id = *metric_id + 1;
 
-	add_metric_vector(vector, *metric_id, qualifier_name, group_name, "",
-		metric_name, "", metric_units, FPGA_METRIC_DATATYPE_INT, FPGA_METRIC_TYPE_AFU, FPGA_HW_MCP, metric_offset);
+	result = add_metric_vector(vector, *metric_id, qualifier_name, group_name, "",
```
r-rojo: You don't use result at all other than here. Why not remove it and just
```C
return add_metric_vector(...);
```

PR #1269, opened 2019-05-09
Make libopae-c-ase.so the duplicate of libopae-c.so in order to make …

lzhan55:
…the ASE plugin changes backward compatible for ASE applications


libopae/init_ase.c
```diff
@@ -60,9 +62,45 @@ STATIC const char *_opae_sys_cfg_files[SYS_CFG_PATHS] = {
...
+	vfprintf(fp, fmt, argp);
+	err = pthread_mutex_unlock(&log_lock);
+	if (err)
+		fprintf(stderr, "pthread_mutex_unlock() failed: %s",
+			strerror(err));
+	va_end(argp);
+}
+
+/* Find the canonicalized configuration file opae_ase.cfg. If null, the file
+	was not found. Otherwise, it's the first configuration file found from a
```
tswhison: Indent is off (x3 lines)

libopae/init_ase.c
```diff
@@ -60,9 +62,45 @@ STATIC const char *_opae_sys_cfg_files[SYS_CFG_PATHS] = {
...
+	list of possiblepaths. Note: The char * returned is allocated here, caller
```
tswhison: s/possiblepaths/possible paths/

libopae/init_ase.c
```diff
@@ -133,7 +171,37 @@ STATIC char *find_ase_cfg()
...
+	/* try to read loglevel from environment */
+	char *s = getenv("LIBOPAE_LOG");
+	if (s) {
+		g_loglevel = atoi(s);
+#ifndef LIBOPAE_DEBUG
+		if (g_loglevel >= OPAE_LOG_DEBUG)
+			fprintf(stderr,
+				"WARNING: Environment variable LIBOPAE_LOG is "
+				"set to output debug\nmessages, "
+				"but libopae-c was not built with debug "
```
tswhison: s/libopae-c/libopae-c-ase/

michael-adler:
This fixes the original problem: links now work.

However, "with_ase" no longer works. I wonder whether the ordering problem someone suggested a while ago is happening now.

We don't really have to do a preload, now that ASE is actually loaded as a plugin. I looked through the plugin manager and didn't see a way to do it, but what if an environment variable could trigger loading a specific configuration file? If it would, "with_ase" could just set the right environment variable and then invoke the normal binary.

libopae/CMakeLists.txt
```diff
@@ -41,12 +41,19 @@ target_include_directories(opae-c PUBLIC
   PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}
   PUBLIC ${libjson-c_INCLUDE_DIRS})
 
-add_library(opae-c-ase SHARED init_ase.c)
+set(SRC_ASE
+    pluginmgr.c
+    api-shell.c
```
r-rojo: Why are we adding pluginmgr.c, api-shell.c, and props.c in this library? Aren't those compiled into libopae-c.so?

lzhan55: @r-rojo This is to resolve the concerns Michael had at https://github.com/OPAE/opae-sdk/pull/1265.

libopae/init_ase.c
lzhan55: fixed

libopae/init_ase.c
lzhan55: fixed

libopae/init_ase.c
lzhan55: fixed

libopae/init_ase.c
tswhison: align these 3 lines with the F in Find from 99.

libopae/init.c
```diff
@@ -110,9 +199,24 @@ __attribute__((constructor)) STATIC void opae_init(void)
 	if (g_logfile == NULL)
 		g_logfile = stdout;
 
+	with_ase = getenv("WITH_ASE");
+	if (with_ase) {
+		cfg_path = find_ase_cfg();
+
+		if (cfg_path == NULL) {
```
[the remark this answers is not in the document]
tswhison: On second thought, maybe this is OK.. What do others think?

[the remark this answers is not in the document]
lzhan55: Right.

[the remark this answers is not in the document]
tswhison: I think the OPAE_ERR( ) on 207 should mention that WITH_ASE was set.  Otherwise, the user might not fully understand what happened.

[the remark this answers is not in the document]
lzhan55: Let it failed when not finding ase config? There were some race conditions when we updated our comments.

michael-adler:
@lzhan55 - Let me know when you think it is ready to try again.

lzhan55:
@michael-adler you can try again. thanks.

michael-adler:
It works!

Can't init_ase.c be made much smaller now instead of duplicating the search code? It could either set the WITH_ASE environment variable in a constructor (not sure if that would work) or you could add some global boolean variable with the same side effect as WITH_ASE that init_ase.c could set and the code in init.c could read.

lzhan55:
@michael-adler Does DCP regression working? Thanks. For now I would like to keep the search code duplicated. When WITH_ASE is set in the script, the application will search for opae_ase.cfg even if it is linked with libopae-c.so library. If libopae-c-ase.so is linked directly, it should search only for opae_ase.cfg. Either way ASE will be executed correctly.

michael-adler:
@deepakcu84 said he would test when you are ready

lzhan55:
@tswhison @nakulkorde Can you review the code change so that I can merge the changes into feature/ase_plugin3 so that Deepak can run regression on the feature branch some day? Thanks.

tswhison:
I like Michael's idea to remove the duplicated code.  If this goes into the feature branch as is, can we revisit that before it goes into master?

lzhan55:
@tswhison Do you mean in init_ase.c, we only set the environment variables WITH_ASE, and call the fpgaInitialize with opae_ase.cfg file from libopae-c.so? I can try now.

tswhison:
You might experiment with adding both init.c and init_ase.c to libopae-c-ase.  The constructor in init_ase.c will fire first, because it has been prioritized.  I think it could then just set WITH_ASE as you suggest.

lzhan55:
@tswhison There are two cases in ASE: 
1. ASE application is linked with libopae-c-ase.so directly. So we duplicated the libopae-c.so to make it work
2. ASE application is invoked with with_ase.sh as "with_ase ./hello_fpga". In this case, the application will be linked directly to libopae-c.so. 

In order to satisfy both use cases, I think we should keep the code as it is with some duplicating code of search for opae_ase.cfg. I don't see a reason to try something like making libopae-c-ase.so a proxy library. Please correct me if my understanding is wrong.

michael-adler:
You already have the new init.c that responds to WITH_ASE linked into libopae-c-ase. The build in its current configuration guarantees that. The only thing now required in init_ase.c is setenv("WITH_ASE") in a constructor so that it runs before the init.c code.

lzhan55:
@michael-adler your assumption is that the application is linked with both libopae-c-ase.so and libopae-c.so, right? But WITH_ASE case, ASE is linked to libopae-c.so, not libopae-c-ase.so.

michael-adler:
All of libopae-c's plugin manager code is now inside libopae-c-ase. You did that yesterday.

lzhan55:
@michael-adler I will sync up with you in the afternoon.

michael-adler:
@tswhison is on the same track. You could just talk to him.

lzhan55:
@michael-adler It is OK as long as it passed the checks.

PR #1273, opened 2019-05-10
Simple WITH_ASE

michael-adler:


michael-adler:
@lzhan55 - This is what we meant

lzhan55:
OK. It might be confusing for people, but it works.

lzhan55:
I will merge it into lzhan55/ase_plugin9. Thanks.

PR #1280, opened 2019-05-14
Added some unit tests for ASE configuration

lzhan55:


testing/opae-c/test_init_c.cpp
```diff
@@ -166,3 +172,245 @@ TEST(init, log_file) {
...
+    char *cfg_file = nullptr;
+    char *opae_path;
+    std::string  inst_cfg_file_, opae_cfg_file_;
+    std::string  rel_cfg_file_, rel_cfg_file2_;
+
+    // copy it to a temporary buffer that we can use dirname with
+    std::string inst_cfg_path = (OPAE_ASE_CFG_INST_PATH? OPAE_ASE_CFG_INST_PATH : "");
+    std::copy(inst_cfg_path.begin(), inst_cfg_path.end(), &buffer_[0]);
+    char *inst_cfg_dir = dirname(buffer_);
+    char *inst_cfg_file = basename(buffer_);
```
tswhison: Should probably check that these are valid before referencing them.

testing/opae-c/test_init_c.cpp
```diff
@@ -166,3 +172,245 @@ TEST(init, log_file) {
...
+
+    std::string cfg_dir = (inst_cfg_dir? inst_cfg_dir : "");
+    inst_cfg_file_ = cfg_dir + std::string(inst_cfg_file) + std::string(".backup");
```
tswhison: What happens if cfg.backup already existed?

testing/opae-c/test_init_c.cpp
```diff
@@ -166,3 +172,245 @@ TEST(init, log_file) {
...
+    opae_path = getenv("OPAE_PLATFORM_ROOT");
+    std::string rel_cfg_path = (opae_path? opae_path: "");
+    rel_cfg_file_ = rel_cfg_path + std::string("/share/opae/ase/opae_ase.cfg");
+    rel_cfg_file2_ = rel_cfg_path + std::string("/share/opae/ase/opae_ase.cfg.backup");
+    rename(rel_cfg_file_.c_str(), rel_cfg_file2_.c_str());
+
+    // find_ase_cfg at HOME directory
+    cfg_file = find_ase_cfg();
+    EXPECT_NE(cfg_file, nullptr);
+
```
tswhison: Need to free cfg_file here.

testing/opae-c/test_init_c.cpp
```diff
@@ -166,3 +172,245 @@ TEST(init, log_file) {
...
+    char *cfg_file = nullptr;
+    char *opae_path;
+    std::string  inst_cfg_file_, opae_cfg_file_;
+    std::string  rel_cfg_file_, rel_cfg_file2_;
+
+    // copy it to a temporary buffer that we can use dirname with
+    std::string inst_cfg_path = (OPAE_ASE_CFG_INST_PATH? OPAE_ASE_CFG_INST_PATH : "");
+    std::copy(inst_cfg_path.begin(), inst_cfg_path.end(), &buffer_[0]);
+    char *inst_cfg_dir = dirname(buffer_);
+    char *inst_cfg_file = basename(buffer_);
```
tswhison: Should verify these before referencing them.

testing/opae-c/test_init_c.cpp
```diff
@@ -166,3 +172,245 @@ TEST(init, log_file) {
...
+    opae_path = getenv("OPAE_PLATFORM_ROOT");
+    std::string rel_cfg_path = (opae_path? opae_path: "");
+    rel_cfg_file_ = rel_cfg_path + std::string("/share/opae/ase/opae_ase.cfg");
+    rel_cfg_file2_ = rel_cfg_path + std::string("/share/opae/ase/opae_ase.cfg.backup");
+    rename(rel_cfg_file_.c_str(), rel_cfg_file2_.c_str());
+
+    // find_ase_cfg at HOME directory
+    cfg_file = find_ase_cfg();
+    EXPECT_NE(cfg_file, nullptr);
+
```
tswhison: Need to free cfg_file here.

testing/opae-c/test_init_c.cpp
```diff
@@ -166,3 +172,206 @@ TEST(init, log_file) {
...
+    // let's rename the opae_ase.cfg in OPAE_ASE_CFG_SRC_PATH and OPAE_ASE_CFG_INST_PATH
+
+    // copy it to a temporary buffer that we can use dirname with
+    std::string src_cfg_path = (OPAE_ASE_CFG_SRC_PATH? OPAE_ASE_CFG_SRC_PATH : "");
+    std::copy(src_cfg_path.begin(), src_cfg_path.end(), &buffer_[0]);
+    char *src_cfg_dir = dirname(buffer_);
+    char *src_cfg_file = basename(buffer_);
+
+    std::string cfg_dir = (src_cfg_dir? src_cfg_dir : "");
+    src_cfg_file_ = cfg_dir + std::string(src_cfg_file) + std::string(".backup");
```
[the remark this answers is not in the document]
lzhan55: ASE config file opae_ase.cfg.backup will be overwritten if it exists.

testing/opae-c/test_init_c.cpp
```diff
@@ -166,3 +172,206 @@ TEST(init, log_file) {
...
+    if (stat(cfg_dir_, &st)) {
+      std::string dir = cfg_dir_;
+      // find the first '/' after $HOME
+      int pos = dir.find('/', home.size());
+      while (pos != std::string::npos) {
+        std::string sub = dir.substr(0, pos);
+        // sub is $HOME/<dir1>, then $HOME/<dir1>/<dir2>, ...
+        // if this directory doesn't exist, create it
+        if (stat(sub.c_str(), &st) && sub != "") {
+          ASSERT_EQ(mkdir(sub.c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH),
```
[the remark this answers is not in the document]
lzhan55: It already does the checking. Thanks for reminding.

testing/opae-c/test_init_c.cpp
```diff
@@ -166,3 +172,245 @@ TEST(init, log_file) {
...
+  init_ase_cfg_p() : buffer_ {0} {}
+
+  virtual void SetUp() override {
+    // let's rename the opae_ase.cfg in OPAE_ASE_CFG_SRC_PATH and OPAE_ASE_CFG_INST_PATH
+
+    // copy it to a temporary buffer that we can use dirname with
+    std::string src_cfg_path = (OPAE_ASE_CFG_SRC_PATH? OPAE_ASE_CFG_SRC_PATH : "");
+    std::copy(src_cfg_path.begin(), src_cfg_path.end(), &buffer_[0]);
+    char *src_cfg_dir = dirname(buffer_);
+    char *src_cfg_file = basename(buffer_);
```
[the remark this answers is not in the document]
lzhan55: Will do null pointer checking.

testing/opae-c/test_init_c.cpp
```diff
@@ -166,3 +172,245 @@ TEST(init, log_file) {
...
+ *             Then the call is successful<br>
+ */
+TEST_P(init_ase_cfg_p, find_ase_cfg_2) {
+    char *cfg_file = nullptr;
+    std::string inst_cfg_file_;
+
+    // find_ase_cfg at OPAE_ASE_CFG_INST_PATH
+    cfg_file = find_ase_cfg();
+    EXPECT_NE(cfg_file, nullptr);
+
```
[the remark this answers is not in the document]
lzhan55: Right!

testing/opae-c/test_init_c.cpp
```diff
@@ -166,3 +172,245 @@ TEST(init, log_file) {
...
+   // rename opae_ase.cfg under releae directory
+    std::string cfg_dir = (inst_cfg_dir? inst_cfg_dir : "");
+    inst_cfg_file_ = cfg_dir + std::string(inst_cfg_file) + std::string(".backup");
+    rename(OPAE_ASE_CFG_INST_PATH, inst_cfg_file_.c_str());
+
+    // find_ase_cfg at release directory
+    cfg_file = find_ase_cfg();
+    EXPECT_NE(cfg_file, nullptr);
+    rename(inst_cfg_file_.c_str(), OPAE_ASE_CFG_INST_PATH);
```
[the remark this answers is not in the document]
lzhan55: OK.

testing/opae-c/test_init_c.cpp
lzhan55: Will do null pointer checking.

testing/opae-c/test_init_c.cpp
lzhan55: Will be overwritten.

testing/opae-c/test_init_c.cpp
lzhan55: Right

testing/opae-c/test_init_c.cpp
lzhan55: Will do the checking.

testing/opae-c/test_init_c.cpp
lzhan55: Right.

testing/opae-c/test_init_c.cpp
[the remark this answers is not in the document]
lzhan55: Right.

testing/opae-c/test_init_c.cpp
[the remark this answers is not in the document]
lzhan55: done

testing/opae-c/test_init_c.cpp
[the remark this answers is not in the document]
lzhan55: done

testing/opae-c/test_init_c.cpp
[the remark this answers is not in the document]
tswhison: I'd feel better if we used mkstemp instead.  It "generates a unique temporary filename".
You would do something like:
```c
char tmpfile[32];
strcpy(tmpfile, "opae_ase.cfg.XXXXXX");
close(mkstemp(tmpfile));
```
Then, move the existing opae_ase.cfg to the tmpfile name.

testing/opae-c/test_init_c.cpp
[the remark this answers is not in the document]
tswhison: Null checks won't help.  man dirname says that these functions always return a pointer to a NULL-terminated string.  You will need to see what the invalid values are and use strcmp.

testing/opae-c/test_init_c.cpp
tswhison: NULL pointer checks won't work. See man dirname.

testing/opae-c/test_init_c.cpp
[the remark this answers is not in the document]
lzhan55: Will change it. Good suggestion!

testing/opae-c/test_init_c.cpp
tswhison: Use mkstemp instead.

testing/opae-c/test_init_c.cpp
tswhison: Will need to use strcmp..

testing/opae-c/test_init_c.cpp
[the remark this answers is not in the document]
lzhan55: Below are the definitions in config.h.in: 
#define OPAE_ASE_CFG_SRC_PATH "@CMAKE_CURRENT_SOURCE_DIR@/libopae/plugins/ase/opae_ase.cfg"
#define OPAE_ASE_CFG_INST_PATH "@CMAKE_INSTALL_PREFIX@/share/opae/ase/opae_ase.cfg"

It will always return some directory name by definition. Do you think we still need to check?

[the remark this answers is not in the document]
tswhison: It's probably not critical; however please remove the nullptr checks for clarity.

testing/opae-c/test_init_c.cpp
lzhan55: done

testing/opae-c/test_init_c.cpp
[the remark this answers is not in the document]
lzhan55: done

testing/opae-c/test_init_c.cpp
```diff
@@ -166,3 +172,274 @@ TEST(init, log_file) {
...
+        free(cfg_file);
+
+    // copy it to a temporary buffer that we can use dirname with
+    std::string inst_cfg_path = (OPAE_ASE_CFG_INST_PATH? OPAE_ASE_CFG_INST_PATH : "");
+    std::copy(inst_cfg_path.begin(), inst_cfg_path.end(), &buffer_[0]);
+    char *inst_cfg_dir = dirname(buffer_);
+    std::string cfg_dir = (inst_cfg_dir? inst_cfg_dir : "");
+
+    // rename opae_ase.cfg under installation directory
+    strcpy(tmpfile2, "opae_ase.cfg.XXXXXX");
```
ibarraza: opae_ase.cfg.XXXXXX doesn't seem to get cleaned up after the test.

testing/opae-c/test_init_c.cpp
```diff
@@ -166,3 +172,274 @@ TEST(init, log_file) {
...
+    char tmpfile2[32];
+
+    // copy it to a temporary buffer that we can use dirname with
+    std::string inst_cfg_path = (OPAE_ASE_CFG_INST_PATH? OPAE_ASE_CFG_INST_PATH : "");
+    std::copy(inst_cfg_path.begin(), inst_cfg_path.end(), &buffer_[0]);
+    char *inst_cfg_dir = dirname(buffer_);
+    std::string cfg_dir = (inst_cfg_dir? inst_cfg_dir : "");
+
+    // rename opae_ase.cfg under installation directory
+    strcpy(tmpfile2, "opae_ase.cfg.XXXXXX");
```
ibarraza: opae_ase.cfg.XXXXXX doesn't seem to get cleaned up after the test.

testing/opae-c/test_init_c.cpp
```diff
@@ -166,3 +172,274 @@ TEST(init, log_file) {
...
```
ibarraza: opae_ase.cfg.XXXXXX doesn't seem to get cleaned up after the test.

testing/opae-c/test_init_c.cpp
lzhan55: Did you see test failures? I will double check.

ibarraza: No, I just mean that the files get left behind after the tests. You unlink the file in the first test but not the others.

testing/opae-c/test_init_c.cpp
```diff
@@ -290,7 +290,9 @@ class init_ase_cfg_p : public ::testing::TestWithParam<const char*> {
       dirs_.pop();
     }
     // restore the opae_ase.cfg file at OPAE_ASE_CFG_SRC_PATH
-    rename(tmpfile, (char *)OPAE_ASE_CFG_SRC_PATH);
+    if (rename(src_cfg_file_.c_str(), (char *)OPAE_ASE_CFG_SRC_PATH)) {
+        OPAE_ERR("Rename failed. Need to manully remove the opae_ase.cfg.XXXXXX files");
```
tswhison: Probably should fail the test in this case too, or else the user may not notice..

testing/opae-c/test_init_c.cpp
lzhan55: It is root caused now.

testing/opae-c/test_init_c.cpp
lzhan55: It is root caused now.

testing/opae-c/test_init_c.cpp
```diff
@@ -301,6 +314,7 @@ class init_ase_cfg_p : public ::testing::TestWithParam<const char*> {
   std::stack<std::string> dirs_;
   std::string src_cfg_file_;
   char tmpfile[32];
```
tswhison: tmpfile_

testing/opae-c/test_init_c.cpp
lzhan55: Fixed now.

testing/opae-c/test_init_c.cpp
```diff
@@ -227,7 +227,16 @@ class init_ase_cfg_p : public ::testing::TestWithParam<const char*> {
     strcpy(tmpfile, "opae_ase.cfg.XXXXXX");
     close(mkstemp(tmpfile));
     src_cfg_file_ = cfg_dir + std::string("/") + std::string(tmpfile);
-    rename(OPAE_ASE_CFG_SRC_PATH, src_cfg_file_.c_str());
+    struct stat st;
+    // check if the file exists or not
+    if (!stat(OPAE_ASE_CFG_SRC_PATH, &st)) {
```
tswhison: If either the stat or the rename fails, is it still valid to run the tests?

testing/opae-c/test_init_c.cpp
lzhan55: I added some additional logic to check the file existence before renaming. Now it should work. Sometimes it failed because of the permission issue. Should be OK.

testing/opae-c/test_init_c.cpp
lzhan55: This is the source location. If user installs the OPAE, they can search for other locations for config file.

testing/opae-c/test_init_c.cpp
lzhan55: Is fixed.

testing/opae-c/test_init_c.cpp
tswhison: If this rename fails, then the user's config file is overwritten?  If so, we need to stop the test flow here.

testing/opae-c/test_init_c.cpp
tswhison: If there is any possibility of overwriting the user's config file, ie if stat passes but rename fails, then we should stop immediately.

testing/opae-c/test_init_c.cpp
lzhan55: This is outdated code. I will add the comment to the latest code.

testing/opae-c/test_init_c.cpp
lzhan55: Make it fail now.

PR #1285, opened 2019-05-21
Fix a few compilation problems on old compilers.

michael-adler:


michael-adler:
Needed to compile in DCP test environment (gcc 4.8.5)

PR #1299, opened 2019-06-04
Fix a bug in ASE that incorrectly handled munmap of a buffer.

michael-adler:
Wrong level was passed to ASE's internal page table manager, leading
to a SEGV. Tested with MPF/VTP's random tester, which is the only
code that exercies repeated alloc/free of pinned buffers.


michael-adler:
@OPAE/ase - The moved ASE tree is still tagged as belonging to @OPAE/api for approval.

PR #1300, opened 2019-06-04
Work around a glitch in the ASE DDR model

michael-adler:
The DDR simulation model's Avalon ready/enable protocol deliberately uses clock offsets, which wreak havoc even in the Qsys-supplied bridges. Adding a small delay to the simulated DDR clock passed to the AFU causes the AFU to sample signals when they are stable.


michael-adler:
This appears to have been the cause of random failures in hello_mem_afu simulation.

Tests: http://sw-pert.altera.com/pert/pert.php?test_run_id=3410442

PR #1305, opened 2019-06-06
Add fpgaWriteMMIO512()

michael-adler:


michael-adler:
@OPAE/ci and @OPAE/api -- I'm done. Sorry for all the system-generated intermediate e-mail while getting the tests working.

Tested using hardware and in ASE.

The tests in the test system that check for parameter errors are active. I had to disable the tests that actually try to do the 512-bit MMIO writes because the Travis machines don't appear to support AVX-512, so trigger illegal instructions errors. We can re-enable them once the test machines are updated.

testing/ase/test_mmio_c_ase.cpp
```diff
@@ -145,6 +156,8 @@ TEST(mmio_c_ase, offset_overflow) {
             FPGA_INVALID_PARAM);
```
jarrodblackburn: Not that this is related to your change but shouldn't this 64-bit test be passing in offsets that are 8 byte aligned (i.e. +8 instead of +4)?

michael-adler: Some of the tests are to confirm that illegal address alignment triggers an error.

libopae/plugins/xfpga/mmio.c
```diff
@@ -319,6 +319,53 @@ fpga_result __FPGA_API__ xfpga_fpgaReadMMIO64(fpga_handle handle,
 	return result;
 }
 
+static inline void copy512(const void *src, void *dst)
```
tswhison: Seems like we should wrap this in a preprocessor check for the necessary AVX support and error out / provide an alternate implementation when vmovdqu64 it is not found.  eg, our ARM customers will not be happy with this as is.
Also, do we need to tell gcc that zmm0 has been clobbered?

michael-adler: You're right we should have the function fail on non-x86 architectures, for now, determined at compile time. How is that checked? In the future, if we get write combining working or other methods for making it work on other platforms, even better.

On x86, it does trigger an illegal instruction at run time when not supported. To me, getting the feature in for the most common platform seems like the right choice. It is performance critical for some algorithms. Application writes can decide what machines they support and whether they want to back off to 64 bit writes based on the CPU's support for AVX-512.

I'm pretty sure the asm syntax is correct and gcc will detect the use of zmm0. If anyone knows better syntax let me know.

tswhison: re: preprocessor checks for AVX instructions: https://stackoverflow.com/questions/28939652/how-to-detect-sse-sse2-avx-avx2-avx-512-avx-128-fma-kcvi-availability-at-compile

michael-adler: It has to be at run time (at least for x86 targets).

michael-adler:
@HananiaAmir I'm glad you are pushing this through! I was going to ask since I need ASE support for 512 bit MMIO in order to test new PIM features.

libopae/plugins/xfpga/mmio.c
HananiaAmir: I added flags field in fpga_handle.
OPAE_FLAG_HAS_MMX512 flag is set in the open function if GCC and the system support AVX512.
This flag is checked in xfpga_fpgaWriteMMIO512.

PR #1311, opened 2019-06-14
opae.admin: initial check-in of opae.admin pkg

r-rojo:
opae.admin is meant to be a Python package for administrative tasks and
uses the interfaces exported by the OPAE kernel driver.
This first check-in supports basic discovery of sysfs nodes and creates
objects that are instances of sysfs_class class. Client applications can
extend this by deriving from the sysfs_class.


python/opae.admin/opae/admin/sysfs.py
```diff
@@ -0,0 +1,282 @@
...
+        return self.parent.root
+
+    @property
+    def pci_address(self):
+        """pci_address get the pci address of the node"""
+        return self._pci_address['pci_address']
+
+    @property
+    def bdf(self):
+        """pci_address get the pci address of the node"""
```
tswhison: Comment seems to be a copy/paste of line 139.  Should make it specific to bdf() I think.

python/opae.admin/opae/admin/sysfs.py
```diff
@@ -0,0 +1,282 @@
...
+                fd.write(val)
+
+
+    @property
+    def sysfs_path(self):
+        return self._sysfs_path
+
+
+class pci_node(sysfs_node):
+    PCI_BUS_SYSFS = '/sys/bus/pci/devices'
```
tswhison: Thinking of access to our fpga's device nodes (/dev/intel-fpga-port.X, /dev/intel-fpga-fme.0), what would be the means of determining this?  Would we need to define a subclass of pci_node called intel_fpga_accelerator_node, and then provide a means to retrieve the device file name?  Since this is specific to the intel fpga device, it doesn't seem right to put them into pci_node.

python/opae.admin/opae/sysfs/sysfs.py
```diff
@@ -0,0 +1,262 @@
...
+from opae.utils.log import loggable, LOG
+
+
+PCI_ADDRESS_PATTERN = (r'(?P<pci_address>'
+                       r'(?P<segment>[\da-f]{4}):(?P<bdf>(?P<bus>[\da-f]{2}):'
+                       r'(?P<device>[\da-f]{2})\.(?P<function>\d)))')
+PCI_ADDRESS_RE = re.compile(PCI_ADDRESS_PATTERN, re.IGNORECASE)
+
+
+class sysfs_node(loggable):
```
tswhison: I'm confused by this.  There is also a class definition for sysfs_node in admin/sysfs.py.  Was this intentional?

python/opae.admin/opae/sysfs/sysfs.py
```diff
@@ -0,0 +1,262 @@
...
+        with self._open('w') as fd:
+            fd.write(val)
+
+
+    @property
+    def sysfs_path(self):
+        return self._sysfs_path
+
+
+class pci_node(sysfs_node):
```
tswhison: Duplicated as well?

python/opae.admin/opae/admin/sysfs.py
r-rojo: The idea is that we can have a pac class that derives from sysfs_class. Using this can create instances of card-specific classes. This pac class would know about the /dev/... interface.

python/opae.admin/opae/sysfs/sysfs.py
r-rojo: No, this one should go away.

PR #1318, opened 2019-06-23
Update PIM to support release-specific platform interfaces

michael-adler:
- Current default behavior is unchanged
- Detect release-specific RTL tree and use it, when present
- Update ASE top-level so releases may define a platform's wire interface


michael-adler:
Did a full nightly build on 19.1 branch with no errors.
Tests behave as they did before: http://sw-pert.altera.com/pert/pert.php?test_run_id=3460905

PR #1323, opened 2019-06-26
 Update ModelSim ASE configuration to track unpacked arrays

michael-adler:
ModelSim, by default, does not include "Memory" when matching signals by
wildcard. Even simple unpacked arrays are considered memory. The change
here, suggested by Francis O'Brien, causes ModelSim to include unpacked
arrays in data recorded during simulation. The impact on simulation speed
is minimal. As Francis noted, very large arrays are still not included
due to other filters.


michael-adler:
Tests: http://sw-pert.altera.com/pert/pert.php?test_run_id=3470300

PR #1329, opened 2019-07-19
Change ASE emulated DDR parameter to eliminate a warning.

michael-adler:
Emulated DDR reads were emitting warnings about buffer overflow because
AV_MAX_PENDING_READS was set to 9 in the model. We believe that the parameter
only affects the warning and can be set to 0 to disable messages. Timing
appears to be unchanged.


michael-adler:
Tests look ok: http://sw-pert.altera.com/pert/pert.php?test_run_id=3537716

All the DMA tests and hello_mem tests passed.

PR #1330, opened 2019-07-24
Add byte write data encoding to CCI-P

michael-adler:
Note: this is merely a change in write request encoding. Old platforms ignore the encoding, which uses fields that were previously reserved. Actual support requires an updated platform.

### DIFF
--- a/platforms/platform_if/rtl/device_if/ccip_if.vh
+++ b/platforms/platform_if/rtl/device_if/ccip_if.vh
@@ -43,6 +43,13 @@
 // Speculative loads: eREQ_RDLSPEC and the error flag in t_ccip_c0_RspMemHdr
 `define CCIP_RDLSPEC_AVAIL 1
 
+// Byte enable, controlling which bytes in a line are updated.
+// Be careful! This flag only indicates whether the CCI-P data structures
+// can encode byte enable. The flag does not indicate whether the platform
+// actually honors the encoding. For that, AFUs must check the value of
+// ccip_cfg_pkg::BYTE_EN_SUPPORTED.
+`define CCIP_BYTE_EN_AVAIL 1
+
 
 import ccip_if_pkg::*;
 
--- a/platforms/platform_db/platform_defaults/defaults_ccip.json
+++ b/platforms/platform_db/platform_defaults/defaults_ccip.json
@@ -28,6 +28,13 @@
 
    "comment":
       [
+         "Boolean (0/1) indicating whether the platform supports byte-enable",
+         "to update only a portion of a cache line."
+      ],
+   "byte-en-supported": 0,
+
+   "comment":
+      [
          "Minimum number of outstanding lines that must be in flight to",
          "saturate bandwidth on each virtual channel.  (Index of the parameter",
          "is the virtual channel number.)  Maximum bandwidth is typically a",
--- a/platforms/platform_if/rtl/device_cfg/ccip_cfg_pkg.sv
+++ b/platforms/platform_if/rtl/device_cfg/ccip_cfg_pkg.sv
@@ -42,6 +42,11 @@
 // the package long.
 //
 
+// Provide defaults for new parameters
+`ifndef PLATFORM_PARAM_CCI_P_BYTE_EN_SUPPORTED
+  `define PLATFORM_PARAM_CCI_P_BYTE_EN_SUPPORTED 0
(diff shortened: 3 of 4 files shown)


michael-adler:
Tests ok: http://sw-pert.altera.com/pert/pert.php?test_run_id=3548045

platforms/platform_if/rtl/device_if/ccip_if.vh
```diff
@@ -43,6 +43,13 @@
 // Speculative loads: eREQ_RDLSPEC and the error flag in t_ccip_c0_RspMemHdr
 `define CCIP_RDLSPEC_AVAIL 1
 
+// Byte enable, controlling which bytes in a line are updated.
+// Be careful! This flag only indicates whether the CCI-P data structures
+// can encode byte enable. The flag does not indicate whether the platform
+// actually honors the encoding. For that, AFUs must check the value of
+// ccip_cfg_pkg::BYTE_EN_SUPPORTED.
+`define CCIP_BYTE_EN_AVAIL 1
```
deepakcu84: Perhaps choose a better name to match description in the comment? CCIP_BYTE_EN_AVAIL sounds similar to CCIP_BYTE_EN_SUPPORTED. How about CCIP_BYTE_ENCODABLE?

michael-adler: I see your point, but you're saying this now? This is just a cherry-pick of the earlier change from today. The "AVAIL" also matches the CCIP_RDLSPEC_AVAIL, which is used downstream, so would require further changes.

Even if I change it to something like "CCIP_BYTE_EN_ENCODING_AVAIL", isn't it still confusing? I could make it "CCIP_BYTE_EN_ENCODING_AVAIL_BUT_ALSO_CHECK_THAT_THE_PLATFORM_SUPPORTS_IT". 
:-)

Any better ideas or should we do the easy thing and go with this?

deepakcu84: That's creative! Go with the old name :) Sounds much better and I don't mind reading the comment now :)

michael-adler: Glad you like it. :-)

We can still change it in a separate commit if the name turns out to be unpopular. That will make it easier anyway, since we'll have a commit with just the name change to cherry pick back to release branches.

deepakcu84: Yeah, I recommend changing it if end user needs to be aware of the variable. Okay otherwise.

michael-adler: Yes, an AFU might use it.

Are CCIP_ENCODING_HAS_BYTE_WR and CCIP_ENCODING_HAS_RDLSPEC better?

deepakcu84: What is RDLSPEC? and why split RD and WR?

michael-adler: Splitting out WR for CCIP_ENCODING_HAS_BYTE_WR because only WR is currently encoded. If we were to add read in the future we have a clean place to indicate it. (It's unlikely we will do this - probably valuable only for an FPGA reading a host's device memory where reads have device side effects.)

RDLSPEC is used only in MPF right now, but could be put in a FIM so merited a real encoding. It's useful for emitting a read prefetch using virtual addresses when it is possible that the virtual address is invalid (has no translation). Speculative reads return a flag indicating failure and don't trigger a hard failure, which would happen with a normal reference to an untranslated address. Kernels reading buffers of unknown size are likely to benefit the most from this.

deepakcu84: Ok. Makes sense. I'd add a comment for RDLSPEC // speculative read

michael-adler: Do you like CCIP_ENCODING_HAS_BYTE_WR and CCIP_ENCODING_HAS_RDLSPEC? Anyone else have an opinion?

deepakcu84: I'm okay with CCIP_ENCODING_HAS_BYTE_WR and CCIP_ENCODING_HAS_RDLSPEC with a comment added for RDLSPEC

PR #1333, opened 2019-07-26
Add CCI-P byte range write to ASE simulation.

michael-adler:


michael-adler:
Tests ok: http://sw-pert.altera.com/pert/pert.php?test_run_id=3555683

The MPF test_random program can either use partial write emulation (using MPF's previously existing module) or it can send requests to a FIM. I used it to validate the ASE implementation. Still on a branch in the BBB repository (mcadler/byte_en).

PR #1335, opened 2019-08-08
Match the fpgaPrepareBuffer() huge page model better in ASE.

michael-adler:
The hardware version of fpgaPrepareBuffer() rounds up requests to huge pages.
The ASE equivalent tried to do this, but ultimately failed when inserting
non-page-size buffers in the simulated page table.


michael-adler:
Tests look normal: http://sw-pert.altera.com/pert/pert.php?test_run_id=3593604

Travis failure is some internal problem. How do we deal with that?

tswhison:
I restarted the job.  Let's see if it passes this time.

PR #1336, opened 2019-08-14
The PIM v2 interface for Avalon memory is now generic.

michael-adler:
The bus definition doesn't have to be specific to "local_mem".
This change affects only PIM v2.


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=3636020

PR #1337, opened 2019-08-16
Make hello_fpga behave better in ASE.

michael-adler:
- Don't time out quickly.
- Wait for activity to complete before releasing memory.

### DIFF
--- a/samples/hello_fpga.c
+++ b/samples/hello_fpga.c
@@ -83,6 +83,7 @@ int usleep(unsigned);
 #define CSR_SRC_ADDR                 0x0120
 #define CSR_DST_ADDR                 0x0128
 #define CSR_CTL                      0x0138
+#define CSR_STATUS1                  0x0168
 #define CSR_CFG                      0x0140
 #define CSR_NUM_LINES                0x0130
 #define DSM_STATUS_TEST_COMPLETE     0x40
@@ -230,6 +231,33 @@ out:
 	return res1 != FPGA_OK ? res1 : res2;
 }
 
+/* Is the FPGA simulated with ASE? */
+bool probe_for_ase()
+{
+	fpga_result r = FPGA_OK;
+	uint16_t device_id = 0;
+	fpga_properties filter = NULL;
+	uint32_t num_matches = 1;
+	fpga_token fme_token;
+
+	/* Connect to the FPGA management engine */
+	fpgaGetProperties(NULL, &filter);
+	fpgaPropertiesSetObjectType(filter, FPGA_DEVICE);
+
+	/* Connecting to one is sufficient to find ASE */
+	fpgaEnumerate(&filter, 1, &fme_token, 1, &num_matches);
+	if (0 != num_matches) {
+		/* Retrieve the device ID of the FME */
+		fpgaGetProperties(fme_token, &filter);
+		r = fpgaPropertiesGetDeviceID(filter, &device_id);
+		fpgaDestroyToken(&fme_token);
+	}
+	fpgaDestroyProperties(&filter);
+
+	/* ASE's device ID is 0xa5e */
+	return ((FPGA_OK == r) && (0xa5e == device_id));
+}
+
 int main(int argc, char *argv[])
 {
 	char               library_version[FPGA_VERSION_STR_MAX];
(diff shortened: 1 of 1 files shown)


samples/hello_fpga.c
```diff
@@ -230,6 +231,33 @@ fpga_result get_bus(fpga_token tok, uint8_t *bus)
...
+
+	/* Connect to the FPGA management engine */
+	fpgaGetProperties(NULL, &filter);
+	fpgaPropertiesSetObjectType(filter, FPGA_DEVICE);
+
+	/* Connecting to one is sufficient to find ASE */
+	fpgaEnumerate(&filter, 1, &fme_token, 1, &num_matches);
+	if (0 != num_matches) {
+		/* Retrieve the device ID of the FME */
+		fpgaGetProperties(fme_token, &filter);
```
tswhison: This call will allocate a new properties object, leaking the one allocated on 244.

samples/hello_fpga.c
```diff
@@ -230,6 +231,33 @@ fpga_result get_bus(fpga_token tok, uint8_t *bus)
 	return res1 != FPGA_OK ? res1 : res2;
 }
 
+/* Is the FPGA simulated with ASE? */
+bool probe_for_ase()
```
tswhison: Could you also add a unit test for this fn to testing/hello_fpga/test_hello_fpga_c.cpp?

michael-adler: No. I'm tired of being the one fixing the DCP regression suite. Please work with them and get someone to maintain these.

samples/hello_fpga.c
michael-adler: What's the right thing to do here? Call fpgaDestroyProperties() before 251 or some other call that adds the fme_token to the filter? If the latter, please give me a code snippet.

samples/hello_fpga.c
tswhison: This isn't the DCP regression suite.  The tests I'm referring to are the unit tests for OPAE.  Your submission causes the coverage to drop (Coverage decreased (-0.05%) to 78.722%).  We try to at least maintain the current coverage with new submissions.  Sorry, but your code will not be accepted without abiding by the submission guidelines.

samples/hello_fpga.c
tswhison: To fix, destroy the original properties object before the 2nd call to fpgaGetProperties.

michael-adler: Fixed now, I think. Is the updated code what you meant?

tswhison: Yes, that will do it.

samples/hello_fpga.c
michael-adler: I read the unit test code and have no idea how a unit test for this would be structured. The code always runs, so it is executed by mock_hello_fpga_c_p.

I don't know how the automated process could have made any conclusions about coverage decreasing. There is a conditional block in probe_for_ase(), but there's no way to write a unit test that forces it down a path in the mock environment. The same is true for the new loop at the end waiting for traffic to stop.

The actual behavior, which is more important than trying to convince the mock environment to do something, is quite well tested within the HW simulation environment and within the DCP ASE testing against NLB0 RTL.

tswhison: I think all the needs to be done is to add the function signature for probe_for_ase() to the top of the test file inside the extern "C" { block; then add a test case which calls the function.  That should be sufficient to keep the coverage unchanged.

michael-adler: Huh? probe_for_ase() is called unconditionally by main() and main() is already probed by the tests.

tswhison: I see what you mean.  I suppose it would require too much mocking to get things to flow into the branch on 249.  OK with this one going in.

michael-adler: Right. The unit tests are fine for testing argument parsing, storage leaks, etc. Not so good when it finally comes time to interact with the HW.

I bet that some semantically unreachable parts of the OPAE SDK could be better tested within the HW infrastructure tests. Something to consider for the future.

That first bug (the failure to destroy a token before overwriting it) is, theoretically, discoverable automatically. Have you thought about that direction instead of by inspection?

tswhison: Actually, yes.  We do have a valgrind component of our testing, though it's not automated and is a bit of a pain to deal with.  Automation would certainly be better.

PR #1338, opened 2019-08-16
Fix a race in ASE.

michael-adler:
Shared buffer allocation and address translation share a lock. Buffer
allocation held the lock too long, making a deadlock possible if the
pipe between the two processes is full.


michael-adler:
Exposed in DCP regression suite streaming_dma test.

PR #1339, opened 2019-08-19
Reset ASE buffer index counter when client disconnects.

michael-adler:
When running in server mode, ASE may accept a new connection after the previous
one exits. The buffer index count was not being reset for new connections.
The RTL side depends on buffer index 0 being the MMIO space, so failed to
detect MMIO on all but the first connection.


michael-adler:
It could make sense in session init, too. I figured at deallocation they were like counters of active state, so can also make sense where they are now while the active buffers are deallocated.

PR #1341, opened 2019-08-28
OPAE in containers: don't enumerate missing device

kad:
In case of container usages with FPGA, sysfs entries might have
more information than actually available to the container.
Ignore devices during enumeration that don't have valid /dev
nodes.

### DIFF
--- a/libopae/plugins/xfpga/enum.c
+++ b/libopae/plugins/xfpga/enum.c
@@ -343,6 +343,18 @@ STATIC fpga_result enum_fme(const char *sysfspath, const char *name,
 
 	snprintf_s_s(dpath, sizeof(dpath), FPGA_DEV_PATH "/%s", name);
 
+	// Make device node exists
+	if (stat(dpath, &stats) != 0) {
+		FPGA_MSG("stat failed: %s", strerror(errno));
+		return FPGA_OK;
+	}
+
+	// In ideal case we need to check that device that we open is really an
+	// FPGA device, but that will prevent easy unittesting
+	//
+	// if (!S_ISCHR(stats.st_mode))
+	// 	return FPGA_OK;
+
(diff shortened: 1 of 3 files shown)


r-rojo:
The tests are failing, can you try to see why?

kad:
@r-rojo it was failing because I was too strict on validation of device files (must be character device). Now I did it a bit more relaxed. However, if you know how to do it better (e.g. mocking stat on those files?) it might be great to put strictness back.

Anyway, original intention for the patch was to get feedback from OPAE developers.
This patch solved failure that I observe, but frankly speaking, I'm not sure it is 100% correct.

Same NLB3 code works correctly within container on DFL drivers and same binary was crashing on machine with Intel-FPGA drivers. In both cases `/dev/*-fme.*` was not present in the container, but in case of Intel-FPGA drivers libopae was trying to open it and with DFL doesn't.

Patch at least asserts that if FME device node is not accessible (e.g. if VF port will be passed to VM or container, and PF don't), libopae will not try to show presence of FME device for Port during enumeration. Does this make sense or it should be solved differently?

bart0sh:
This is still valid fix for the latest OPAE release 1.4.0-1. Without this patch nlb3 fails this way:

```
terminate called after throwing an instance of 'opae::fpga::types::no_driver'
  what():  failed with return code FPGA_NO_DRIVER at: handle.cpp:open():57
/usr/local/bin/test_fpga.sh: line 22:     8 Aborted                 (core dumped) nlb3
```

Any estimations when this PR is merged?

kad:
PR updated on top of master. ping @r-rojo

tswhison:
Hello, we believe we have this addressed as of acee733 on master.  Could you please try with the latest code and let us know if this is resolved in your environment? Thanks

PR #1342, opened 2019-08-29
Use POSIX-friendly commands in the ASE Makefile

michael-adler:
ASE failed to build on Ubuntu, where dash is the default shell. Replace non-POSIX
bash string substitution with variations of the "find" command.


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=3689590

michael-adler:
@OPAE/api - Can someone in the api group please review. Only 3 lines changed in the ASE Makefile.

PR #1343, opened 2019-09-03
Fix fpgabist for PAC regressions

michael-adler:
The Vista Creek (VC) merge left fpgabist broken when used with Darby Creek
regressions. The new VC bist code hard-wired the device ID of Rush Creek and
promoted device ID handling so that it had to match the board, even when a
full BDF was supplied.

The changes here:

- Demote --device-id so it is just one of the components in the search for a board.
  The default value is None. The new algorithm looks for all available FPGAs on the
  host and then filters them using the supplied BDF and device ID. The test can run
  as long as the filter matches a single FPGA.

- Cherry pick a few changes from the DC branch.

- Repair the non-trivial merge of the DC and the VC changes.

- Fix nlb0 clock frequency discovery. It was reporting in correct performance.

### DIFF
--- a/tools/extra/pac/fpgabist/bist_common.py
+++ b/tools/extra/pac/fpgabist/bist_common.py
@@ -87,29 +87,29 @@ def get_all_fpga_bdfs(args):
         if data:
             with open(os.path.join(devpath, "device"), 'r') as fd:
                 device_id = fd.read().strip()
-            if int(device_id, 16) == int(vars(args)['device_id'], 16):
-                bdf_list.append(dict([(k, hex(int(v, 16)).lstrip("0x"))
-                                for (k, v) in data.iteritems()]))
+
+            # Add device ID to the data dictionary
+            data['device_id'] = device_id
+
+            # Add this BDF/device to the list
+            bdf_list.append(dict([(k, int(v, 16))
+                                  for (k, v) in data.iteritems()]))
+
     return bdf_list
 
 
-def get_bdf_from_args(args):
-    pattern = (r'(?P<bus>[a-fA-F0-9]{2}):'
-               r'(?P<device>[a-fA-F0-9]{2})\.(?P<function>[a-fA-F0-9]).*?.')
-    dev_id = r'{:04x}'.format(int(vars(args)['device_id'], 16))
-    bdf_pattern = re.compile(pattern+dev_id)
+# Given a list of all available FPGAs, return a list filtered by
+# the command line arguments.
+def get_bdf_from_args(all_bdfs, args):
     bdf_list = []
-    param = ':{}:{}.{}'.format(
-            hex(int(vars(args)['bus'], 16))
-            if vars(args)['bus'] else '',
-            hex(int(vars(args)['device'], 16))
-            if vars(args)['device'] else '',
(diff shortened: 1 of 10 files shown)


michael-adler:
Test state: http://sw-pert.altera.com/pert/pert.php?test_run_id=3698500

DMA doesn't compile yet due to Qsys, which accounts for a number of the failures.

tools/extra/pac/fpgabist/bist_common.py
```diff
@@ -54,21 +54,26 @@ def check_required_cmds():
...
         output = subprocess.Popen(cmd, stdout=subprocess.PIPE)
         json_data = json.loads(output.communicate()[0])
         accel_data = json_data["afu-image"]["accelerator-clusters"]
         uuid = accel_data[0]["accelerator-type-uuid"].encode("ascii")
         return uuid.lower().replace("-", "")
-    else:
-        bdf_pattern = re.compile(BDF_PATTERN)
-        for fpga in glob.glob('/sys/class/fpga/*'):
+    elif bdf:
+        pattern = '{:02x}:{:02x}.{:x}'.format(bdf['bus'], bdf['device'],
```
buie1: Would it be easier to rename the macro BDF_PATTERN with this pattern you are looking for here?
Did you find an issue using the other pattern or is this just a cleaner regex?

tools/extra/pac/fpgabist/bist_common.py
```diff
@@ -120,11 +125,13 @@ def get_mode_from_path(gbs_path):
     return None
 
 
-def load_gbs(gbs_file, bus_num):
+def load_gbs(gbs_file, bdf):
     print "Attempting Partial Reconfiguration:"
-    cmd = "{} -B 0x{} -v {}".format('fpgaconf', bus_num, gbs_file)
+    cmd = ['fpgaconf', '-B', hex(bdf['bus']), '-D',
```
buie1: This is a good addition here.

tools/extra/pac/fpgabist/bist_dma.py
```diff
@@ -55,26 +55,33 @@ def run_cmd(self, cmd):
...
-            bc.load_gbs(gbs_path, bus_num)
+            bc.load_gbs(gbs_path, bdf)
         ret = 0
         for func, param in self.executables.items():
             if bd_id in dma_list:
                 for i, c in dma_list[bd_id].items():
                     name, size = c
                     if self.dev_id != int(param, 16):
                         continue
-                    cmd = "{} {} -B 0x{} -D {} -S {}".format(func, param,
```
buie1: Is this working? Removing the leading `0x` before this was required so are you passing the leading `0x` with each element of the bdf?

michael-adler: No, DMA is broken still. I need the changes that were cherry picked into the internal master to test it, so was waiting for this pull to go in.

tools/extra/pac/fpgabist/bist_common.py
michael-adler: Ok. I'll do that in the next round with the DMA update.

PR #1347, opened 2019-09-06
fpgabist: dma: allow configuration without TBB

tswhison:
Allow the cmake and build steps to continue in the case that the TBB
package is not found. When TBB is found, fpgabist/dma is included in
the build.

### DIFF
--- a/cmake/modules/FindTbb.cmake
+++ b/cmake/modules/FindTbb.cmake
@@ -45,5 +45,9 @@ find_path(TBB_INCLUDE_DIRS
           ${PC_TBB_INCLUDE_DIRS}
           ${TBB_INCLUDE_DIRS})
 
+if(TBB_LIBRARIES AND TBB_INCLUDE_DIRS)
+  set(TBB_FOUND true)
+endif(TBB_LIBRARIES AND TBB_INCLUDE_DIRS)
+
 include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(TBB REQUIRED_VARS TBB_INCLUDE_DIRS TBB_LIBRARIES)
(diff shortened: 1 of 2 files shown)


tools/extra/pac/fpgabist/CMakeLists.txt
```diff
@@ -34,9 +34,12 @@ set(PYTHON_SRC
      bist_nlb3.py
      fpgabist)
 
- add_subdirectory(dma)
- add_subdirectory(dma_vc)
- add_subdirectory(bist)
+if(TBB_FOUND)
+  add_subdirectory(dma)
```
michael-adler: An else clause here to print a message that DMA won't be built would save some debugging when it happens.

Thank you for resolving the TBB problem.

tswhison: For some reason, FindTbb.cmake isn't finding tbb/tbb.h.  I experimented with it a bit, but have instead made the condition if(TBB_LIBRARIES), which is being populated when tbb is present.  It's not elegant, but the alternative would be to always exclude the dma app.

michael-adler: Is include missing because you have to check the tbb-devel package instead of tbb?

tswhison: I tried editing the pkg_check_modules call like so, but it still isn't found:
```cmake
pkg_check_modules(PC_TBB QUIET tbb-devel)
```

cmake/modules/FindTbb.cmake
```diff
@@ -40,10 +40,14 @@ if(NOT TBB_INCLUDE_DIRS)
   set(TBB_INCLUDE_DIRS "")
 endif()
 find_path(TBB_INCLUDE_DIRS
```
michael-adler: I had a minute to experiment. Your bug is here. When TBB_INCLUDE_DIRS is already defined, the find_path seems to fail (or not run). Change the find_path target variable to TBB_INCLUDE, test for TBB_INCLUDE in the downstream places and it appears to work fine.

tswhison: Thanks for that!

PR #1348, opened 2019-09-09
opae: build: implement INTEL_FPGA_TREE_DIRTY

tswhison:
Implement a preprocessor macro to capture whether the current build
is from a git tree which has modified files.


r-rojo:
Why is this necessary?

tswhison:
It's going to feed into the implementation of a --version option for the apps.

PR #1349, opened 2019-09-09
ase_userguide.md - update supported simulators

HananiaAmir:


michael-adler:
Wouldn't it be better to show the minimum version of the simulator tools instead of some arbitrary maximum? Usually, you specify the oldest that is known to work and assume that future versions continue to support all the features required.

PR #1354, opened 2019-10-14
ASE RTL bug fixes

michael-adler:
- Previous code forced c0Rx header and data to zero when there was no message.
   I've now seen a few broken AFUs that fail on HW but work in simulation because
   of the default zero state.

- Eliminate an internal error in the AVMM response scheduling by managing
  waitrequest in the ASE model driver. The AVMM model was delaying read responses
  when waitrequest was high, which led to the model trying to schedule multiple
  responses in the same cycle.

- Instantiate one AVMM bank in the lowest level module instead of two. Doing two
  at once was needlessly complicated and made it hard to model systems with an odd
  number of banks.

### DIFF
--- a/libopae/plugins/ase/rtl/ase_sim_local_mem.sv
+++ b/libopae/plugins/ase/rtl/ase_sim_local_mem.sv
@@ -80,7 +80,7 @@ module ase_sim_local_mem_avmm
    // emif model
    genvar b;
    generate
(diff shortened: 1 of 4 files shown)


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=3819709

The DDR model bug was observable in the dma_afu when setting --data_size to something larger than the usual test value (e.g. 65536).

PR #1362, opened 2019-11-01
Remove "-v" argument to fpgaconf in fpgabist

michael-adler:
fpgaconf recently changed it's argument parsing, causing "-v" to print the
version. It fails when specified with other arguments, leading to fpgabist
failures.


michael-adler:
DMA and NLB3, among others, were failing before this change:
http://sw-pert.altera.com/pert/pert.php?test_run_id=3874616

michael-adler:
@OPAE/tools - waiting on you here

buie1:
This would change the output of the fpgabist tool since we are removing the verbose output.  Wouldn't it be better to change from `-v` to `-V`

michael-adler:
My frustration is HW regression tests remain broken until this change goes in because nobody tested the consequences of the argument change. You are all free to figure out a reasonable compromise, but let’s at least get regressions working again.

PR #1363, opened 2019-11-04
hello_fpga: don't use version macros from config.h

tswhison:
Relying on config.h broke the ASE regression, which peeks inside
the OPAE tree to build hello_fpga. Use the version API's instead
to query the version.

### DIFF
--- a/samples/hello_fpga.c
+++ b/samples/hello_fpga.c
@@ -149,6 +149,8 @@ fpga_result parse_args(int argc, char *argv[])
 	int getopt_ret;
 	int option_index;
 	char *endptr = NULL;
+	char version[32];
+	char build[32];
 
 	while (-1 != (getopt_ret = getopt_long(argc, argv, GETOPT_STRING,
 						longopts, &option_index))) {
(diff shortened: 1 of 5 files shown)


samples/hello_fpga.c
```diff
@@ -175,10 +177,10 @@ fpga_result parse_args(int argc, char *argv[])
 			break;
 
 		case 'v':
-			printf("hello_fpga %s %s%s\n",
-			       INTEL_FPGA_API_VERSION,
-			       INTEL_FPGA_API_HASH,
-			       INTEL_FPGA_TREE_DIRTY ? "*":"");
+			fpgaGetOPAECVersionString(version, sizeof(version));
```
chipb: What's a bit surprising here is that the `-v` option is reporting not the version of the application (as what I *think* would be the default expectation) but the version of the shared library that it gets linked against. :thinking:

Since they're captured in the same repo, I suppose they're effectively the same thing right now. But there's also not really a reason that a hypothetical `opae-libs-2.0.3` should be incompatible with a hypothetical `opae-samples-2.0.[!3]`, in which case it'll be reporting 2.0.3 erroneously...

How does providing default `#define` values for these three macros sound? If a consuming process isn't going following the contract that they should use CMake to configure their build, they can't reasonably expect better than dummy values for a version string.

tswhison: Thanks for the feedback.  The version mechanisms (both here and in Security-Tools) match the versioning model that we use when releasing the packages.  ie, if a package is versioned 1.2.3, then each of the binaries will report 1.2.3.  I don't see that a separate versioning scheme for each application is beneficial.  If my release/library version is 1.2.3, but an app reports 4.5.6, I'm not sure what conclusion can be drawn about the relationship between the two.  Let's leave it as is for now.  We can always re-visit later if/when needed.

r-rojo: I think @chipb has a point though. I apologize for suggesting getting the version from the library.
I think the application shouldn't rely on whatever it's linked with to determine its version but should instead have the version baked in at compile time.
I agree with @tswhison too that we aren't (at the moment) versioning separately from the version captured by the repo.

PR #1366, opened 2019-11-07
ASE: Ignore CCI-P c1Tx sop when not valid

michael-adler:
A nasty bug in ASE emulation of CCI-P c1Tx: when sop was set in a c1Tx
header the mdata and address were latched, even when the valid bit is
not set. Even worse, the emulator's debug logging printed correct
behavior despite the internal emulation using the wrong values.


michael-adler:
Tests pass: http://sw-pert.altera.com/pert/pert.php?test_run_id=3891346

NLB is failing due to unrelated SW version code changes.

michael-adler:
@OPAE/api - Waiting for you...

PR #1371, opened 2019-11-11
tests: add xfpga plugin test cases

anandaravuri:


testing/xfpga/test_plugin_c.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include "types_int.h"
+#include "adapter.h"
+
+	bool  xfpga_plugin_supports_device(const char *device_type);
+	int xfpga_plugin_initialize(void);
+	int xfpga_plugin_finalize(void);
+	bool  xfpga_plugin_supports_host(const char *hostname);
+	int opae_plugin_configure(opae_api_adapter_table *adapter,
+		const char *jsonConfig);
+}
```
jelonanX: It would look cleaner if we eliminate the tabs and spaces between bool and function definition.

testing/xfpga/test_plugin_c.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include <string>
+#include <vector>
+#include "xfpga.h"
+#include <fcntl.h>
+#include "gtest/gtest.h"
+#include "test_system.h"
+#include "adapter.h"
+#include <dlfcn.h>
+#include <sys/types.h>
+#include <dirent.h>
```
jelonanX: can we clean up the includes to only include what is needed?

testing/xfpga/test_plugin_c.cpp
```diff
@@ -0,0 +1,172 @@
...
+		return adapter;
+	}
+};
+
+/*
+* @test       plugin
+* @brief      Tests: xfpga_plugin_supports_device
+*                    xfpga_plugin_supports_host
+ @details    When passed with valid argument return 0 <br>
+*            When passed with invalid argument return -1 <br>
```
jelonanX: the description here doesn't match the test?

testing/xfpga/test_plugin_c.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+	EXPECT_EQ(opae_plugin_configure(NULL, NULL), -1);
+
+	opae_api_adapter_table *adapter_table = opae_plugin_mgr_alloc_adapter_test("libxfpga.so");
+
+	EXPECT_EQ(opae_plugin_configure(adapter_table, NULL), 0);
+
+	opae_api_adapter_table *adapter_table_invlaid = opae_plugin_mgr_alloc_adapter_test("libxfpga_test.so");
+
+	EXPECT_EQ(opae_plugin_configure(adapter_table_invlaid, NULL), -1);
```
jelonanX: s/invlaid/invalid

testing/xfpga/test_plugin_c.cpp
```diff
@@ -0,0 +1,172 @@
...
+*/
+TEST_P(xfpga_plugin_c_p, test_plugin_2) {
+
+	EXPECT_EQ(opae_plugin_configure(NULL, NULL), -1);
+
+	opae_api_adapter_table *adapter_table = opae_plugin_mgr_alloc_adapter_test("libxfpga.so");
+
+	EXPECT_EQ(opae_plugin_configure(adapter_table, NULL), 0);
+
+	opae_api_adapter_table *adapter_table_invlaid = opae_plugin_mgr_alloc_adapter_test("libxfpga_test.so");
```
jelonanX: s/invlaid/invalid

testing/xfpga/test_plugin_c.cpp
```diff
@@ -0,0 +1,172 @@
...
+	EXPECT_EQ(opae_plugin_configure(adapter_table_invlaid, NULL), -1);
+}
+INSTANTIATE_TEST_CASE_P(xfpga_plugin_c, xfpga_plugin_c_p,
+	::testing::ValuesIn(test_platform::mock_platforms({ "skx-p","dcp-rc" })));
+
+/*
+* @test       plugin
+* @brief      Tests:xfpga_plugin_initialize
+*                   xfpga_plugin_finalize
+* @details    When passed valid argument return 0 <br>
```
jelonanX: description here doesn't match test?

testing/CMakeLists.txt
```diff
@@ -697,6 +697,10 @@ add_unit_test(test_xfpga_threshold_c xfpga-static
     xfpga/test_thresold_c.cpp
 )
 
+add_unit_test(test_plugin_c xfpga-static
+    xfpga/test_plugin_c.cpp
```
ibarraza: Can we rename these test_xfpga_plugin_c for consistency?

testing/xfpga/test_plugin_c.cpp
anandaravuri: fixed

testing/xfpga/test_plugin_c.cpp
anandaravuri: fixed

testing/CMakeLists.txt
anandaravuri: fixed

testing/xfpga/test_plugin_c.cpp
```diff
@@ -0,0 +1,166 @@
...
+	EXPECT_EQ(opae_plugin_configure(NULL, NULL), -1);
+
+	opae_api_adapter_table *adapter_table = opae_plugin_mgr_alloc_adapter_test("libxfpga.so");
+
+	EXPECT_EQ(opae_plugin_configure(adapter_table, NULL), 0);
+
+	opae_api_adapter_table *adapter_table_invalid = opae_plugin_mgr_alloc_adapter_test("libxfpga_test.so");
+
+	EXPECT_EQ(opae_plugin_configure(adapter_table_invalid, NULL), -1);
+}
```
tswhison: adapter_table isn't being freed.

libopae/plugins/xfpga/plugin.c
```diff
@@ -73,6 +73,9 @@ int __FPGA_API__ opae_plugin_configure(opae_api_adapter_table *adapter,
 {
 	UNUSED_PARAM(jsonConfig);
 
+	if (adapter == NULL || adapter->plugin.dl_handle == NULL)
```
tswhison: These checks aren't needed.  See pluginmgr.c : opae_plugin_mgr_load_dflt_plugins( ).

anandaravuri: test cases crash/ core dump  if user pass NULL adapter table pointer.

tswhison: Sure, however, I was pointing out that the code which calls the adapter entry point will have already confirmed that adapter is not NULL and that adapter->plugin.dl_handle is not NULL.

anandaravuri: fixed

testing/xfpga/test_plugin_c.cpp
anandaravuri: fixed

libopae/plugins/xfpga/plugin.c
```diff
@@ -73,6 +73,9 @@ int __FPGA_API__ opae_plugin_configure(opae_api_adapter_table *adapter,
 {
 	UNUSED_PARAM(jsonConfig);
 
+	if (adapter == NULL)
```
tswhison: This check isn't needed.  See pluginmgr.c:679
```c
if (!adapter) {
   OPAE_ERR("malloc failed");
   return ++errors;
}
```

anandaravuri: EXPECT_EQ(opae_plugin_configure(adapter_table, NULL), 0); -> PASS

test case EXPECT_EQ(opae_plugin_configure(NULL, NULL), -1); -> Core dump.
No Null check

tswhison: The problem is that you've inserted a case that does not exist in the code, and then you've added test cases around it.  I'm just pointing out that neither the check nor the test of the check is needed, based on the calling code.  Sure, theoretically you could call this function with a NULL pointer, but it isn't happening today.  Let's not create it.

anandaravuri: Removed test cases and plugin check .

PR #1372, opened 2019-11-11
Added fpgainfo errors.c test cases.

ibarraza:


tools/base/fpgainfo/errors.c
```diff
@@ -180,7 +180,7 @@ static const char *const PORT_ERROR[PORT_ERROR_COUNT] = {
 /*
  * errors command configuration, set during parse_args()
  */
-static struct errors_config {
+STATIC struct errors_config {
```
tswhison: Is this needed after all?  Looks like the struct was re-defined inside the test file.

ibarraza:
Had to reorder a few tests that would set a static variable prior to running the other tests. It would cause the other tests to not take their intended path and reduce the coverage significantly.

testing/fpgainfo/test_fpgainfo_c.cpp
```diff
@@ -493,6 +400,75 @@ TEST_P(fpgainfo_c_p, errors_help) {
...
+
+    strcpy(zero, "fpgainfo");
+    strcpy(one, "errors");
+    EXPECT_EQ(errors_filter(&filter, 2, argv2), FPGA_OK);
+
+    strcpy(two, "-c");
+    EXPECT_EQ(errors_filter(&filter, 3, argv3), FPGA_OK);
+
+    strcpy(two, "--force");
+    EXPECT_EQ(errors_filter(&filter, 3, argv3), FPGA_OK);
```
r-rojo: I don't understand what these tests are doing. Are they testing the errors command or just the filter function for the errors command?

testing/fpgainfo/test_fpgainfo_c.cpp
```diff
@@ -483,6 +333,63 @@ TEST_P(fpgainfo_c_p, errors_command0) {
...
+    num_tokens = matches;
+    ASSERT_EQ(fpgaEnumerate(&filter, 1, tokens, num_tokens, &matches), FPGA_OK);
+
+    strcpy(zero, "fpgainfo");
+    strcpy(one, "errors");
+    strcpy(two, "all");
+    strcpy(three, "-c");
+    EXPECT_EQ(parse_error_args(4, argv), 0);
+
+    EXPECT_EQ(errors_command(tokens, num_tokens, 4, argv), FPGA_OK);
```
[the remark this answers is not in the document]
ibarraza: The parse_errors_args() must be called prior to errors_command() as it sets errors_config.clear static variable. Otherwise the call to errors_command() will not take the clear errors path when called. In the regular flow of the tool, this is how it would go - although it is kinda weird.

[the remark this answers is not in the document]
r-rojo: 👍

testing/fpgainfo/test_fpgainfo_c.cpp
ibarraza: Not entirely sure about this one. Didn't write it, only moved it. The target seems to actually be parse_error_args(), which gets called inside of errors_filter().

PR #1380, opened 2019-11-21
Fix memory leaks in fpgainfo

jelonanX:


libopae/plugins/xfpga/metrics/metrics_max10.c
```diff
@@ -215,6 +215,10 @@ fpga_result  enum_max10_metrics_info(struct _fpga_handle *_handle,
 		}
 		metric_name[strlen(metric_name)-1] = '\0';
 
+		if (tmp) {
+			free(tmp);
+		}
```
tswhison: I think tmp is also leaked at the end of this loop if no error cases are encountered.

tswhison: Oh, nevermind.  It's freed on 275.

PR #1396, opened 2019-12-03
Enable building of opae-tests packages

dasm:
Added opae-tests.(deb|rpm) packages which are required for local testing
of OPAE SDK


nakulkorde:
These changes look good, but since we are planning to branch out for opae 1.4 release this week, is it ok to put this PR on hold for a couple of days?

dasm:
@nakulkorde probably would be nice to have this option included in next release. We're still week away from release.

nakulkorde:
opae 1.4 is the public release of VC. It doesn't add any new features but only enables support for the upstream driver patch-set 2.

dasm:
@nakulkorde if this PR looks good why would we need to wait another week before delivering? I'm required to do another update because my branch is out-of-date.
If this state would persist, then I would be required to update the very same branch approx. 10x

nakulkorde:
@dasm I understand it is a pain to update the branch every so often. But the idea for this release is to avoid any changes unless required for the specific release. We expect to finish work for this release by tomorrow, once we branch out for the release then this can be merged into master.

Unless @rlee5-intel thinks it is ok to merge this in before we branch out for the release?

rlee5-intel:
The next release 1.4.0 is a very targeted release.  We are branching out tomorrow morning and releasing next Monday.  I would recommend that we do not merge this change in 1.4.0 release.  Right after 1.4.0 we will start working on 1.4.1 immediately and we can pick up this change.  Thanks!

nakulkorde:
We have branched out for the release and this can now go in.

PR #1426, opened 2020-01-25
Trix/fedora upstream

trixirt:


scripts/create-rpms.sh
```diff
@@ -7,6 +7,13 @@ if [ $? -eq 1 ]; then
 	exit 1
 fi
 
+#check if rpmdevtools is installed
+yum list installed rpmdevtools
+if [ $? -eq 1 ]; then
```
tswhison: The success checks would be a bit safer if coded like..
```bash
if [ $? -ne 0 ]; then
```

nakulkorde: good point.

PR #1440, opened 2020-02-06
Add reference counting to ASE's pinned buffer model.

michael-adler:
When pinning pages on real FPGAs, the kernel keeps a reference count of
the number of times a page is pinned. Pages must be released the same number
of times before the kernel unpins them. ASE was not keeping the equivalent
count, causing translations to be lost the first time a buffer was released.
New code now uses 8 bit counters to track references.


michael-adler:
Tests behave as expected: http://sw-pert.altera.com/pert/pert.php?test_run_id=4110999

PR #1448, opened 2020-02-11
change libjson0 dependency to libjson-c-dev

ddcc:
At least on Ubuntu, libjson0 is a deprecated transitional package specific for 16.04. libjson-c-dev is the generic libjson development library that depends on a specific libjson-c2/c3/c4 binary build, so use that instead. I'm not sure what the situation is on RHEL, they seem to have a libjson-c-devel package as well?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,24 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${OPAE_SDK_SOURCE}/cmake"
 include(OPAE)
 
 ############################################################################
+## Get distribution information ############################################
+############################################################################
+find_program(LSB_RELEASE_EXEC lsb_release)
+if (LSB_RELEASE_EXEC)
+  execute_process(COMMAND ${LSB_RELEASE_EXEC} -is
+      OUTPUT_VARIABLE LSB_DISTRIBUTOR_ID
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+  execute_process(COMMAND ${LSB_RELEASE_EXEC} -rs
+      OUTPUT_VARIABLE LSB_RELEASE_NUMBER
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+  message(STATUS "Detected current distribution: '${LSB_DISTRIBUTOR_ID}', version: '${LSB_RELEASE_NUMBER}'")
+else()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -389,7 +389,7 @@ define_pkg(devel
   GROUP "libs"
   DISPLAY_NAME "opae-libs"
   DESCRIPTION "OPAE runtime"
-  DEB_DEPENDS "uuid-dev , libjson0"
+  DEB_DEPENDS "uuid-dev , libjson-c-dev"
```
r-rojo: For a runtime library (opae-libs), it should depend on runtime packages.
We should change uuid-dev to uuid but for json-c, we should be more explicit/deliberate  about the package.
While libjson-c-dev pulls in the correct runtime package, that is also pulling in more dependencies than are really needed.
It would be nice if we could use a CMake variable here instead.
Something like:
```
DEB_DEPENDS uuid ${json-c-pkg}
```

And hopefully, we could detect which distro we're on to set that variable (json-c-pkg).

ddcc:
I've added some logic to detect the current distribution/version, but I've only tested it on Ubuntu 18.04 and 19.10.

ddcc:
ping

ddcc:
Yeah, on any system without `lsb_release`, it should fall back to `libjson-c-dev`. I did test on the IL Academic environment, which is CentOS and does have `lsb_release`.

ddcc:
On newer Ubuntu releases, the `libjson0` transitional package isn't even in the repositories anymore, so I'd like to resolve this without needing a workaround.

IMHO, separating runtime and development dependencies based on distribution detection is orthogonal to this goal.

tswhison:
Hi Dominic, I'm bringing in your changes over here: https://github.com/OPAE/opae-libs/pull/18  Once they're in, we can close this one.

ddcc:
Okay, I saw that there's been some reorganization of the repositories. I've submitted a few other pull requests to opae-libs and opae-sim, hope they're in the right place.

tswhison:
The 2nd part is coming in via https://github.com/OPAE/opae-sdk/pull/1486.  Your other pull requests seem to be in order.  Thank you for contributing!

tswhison:
Closing as implemented by  OPAE/opae-libs#18 and  #1486.

ddcc:
In #1486, shouldn't the development library depend on libjson-c-dev instead of ${JSON_C_DEBIAN_PACKAGE}?

PR #1454, opened 2020-02-13
tools: fix fpgabist tools clang issues

anandaravuri:


tools/extra/c++utils/option.h
```diff
@@ -73,6 +73,7 @@ class option
     {
 
     }
+    virtual ~option() {};
```
tswhison: End semicolon is not needed.

PR #1455, opened 2020-02-14
Set the source0 value in the opae.spec to the upstream source.

trixirt:
rpmbuild will fetch the source tarball from its upstream location
if it is not found locally.  So improve the Source0 value to point
to that location.  Change the create-rpm script to use the new
tarball name.

Fixes the error with fedora-review

WARNING: Cannot download url: https://github.com/OPAE/opae/opae.tar.gz
INFO: No upstream for (Source0): opae.tar.gz

### DIFF
--- a/opae.spec
+++ b/opae.spec
@@ -7,7 +7,7 @@ Group:          Development/Libraries
 Vendor:         Intel Corporation
 Requires:       uuid, json-c, python
 URL:            https://github.com/OPAE/%{name}-sdk
-Source0:        https://github.com/OPAE/%{name}/%{name}.tar.gz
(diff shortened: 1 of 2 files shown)


trixirt:
I do not think this is the best place for the change.
There is a serious problem with a usr/ dir in the released tarball vs that created.
The release tarball is what should be used to create the rpm.

So have started a bare repo with just the spec file here
https://github.com/trixirt/opae-rpm
The opae-1.4.0-1 branch

tswhison:
Hey @anandaravuri, have we comprehended this change in the latest RPM spec?  If so, OK to close this one?

PR #1459, opened 2020-02-16
Check if .git directory exists before finding the git exe

trixirt:
When building from a tarball, there will be not a .git dir.
So any git operations will not be valid.  So check for the
.git directory before doing any git operations.


trixirt:
Patch also used in opae-rpm here
https://github.com/trixirt/opae-rpm/commit/3fbb57c0e8c4586ee285a7e6afcfbceaec9c1f7e

PR #1463, opened 2020-02-19
Add INTEL_FPGA_API_VERSION version to shared objects.

trixirt:
Most of the shared objects are of the form <name>.so.<INTEL_API_VERSION>
In this case <name>.so.1.4.0 with standard symlinks <name>.so.1 and
<name>.so.  For those that do not, provide the versioning to make them
consistent.

### DIFF
--- a/libopae/CMakeLists.txt
+++ b/libopae/CMakeLists.txt
@@ -68,6 +68,9 @@ set_property(TARGET opae-c PROPERTY C_STANDARD 99)
 set_target_properties(opae-c PROPERTIES
   VERSION ${INTEL_FPGA_API_VERSION}
   SOVERSION ${INTEL_FPGA_API_VER_MAJOR})
+set_target_properties(opae-c-ase PROPERTIES
+  VERSION ${INTEL_FPGA_API_VERSION}
+  SOVERSION ${INTEL_FPGA_API_VER_MAJOR})
 
 # Add coverage flags
 if(CMAKE_BUILD_TYPE STREQUAL "Coverage")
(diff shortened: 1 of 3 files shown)


libopae/plugins/xfpga/metrics/bmc/CMakeLists.txt
```diff
@@ -52,6 +52,11 @@ set_property(TARGET staticbmc PROPERTY POSITION_INDEPENDENT_CODE ON)
 
 add_library(modbmc MODULE ${SRC})
 
+# Target properties
+set_target_properties(modbmc PROPERTIES
```
tswhison: Module libraries should not set VERSION/SOVERSION

tools/base/fpgad/plugins/fpgad-vc/CMakeLists.txt
```diff
@@ -49,6 +49,9 @@ endif(CMAKE_BUILD_TYPE STREQUAL "Coverage")
 
 # Target properties
 set_property(TARGET fpgad-vc PROPERTY C_STANDARD 99)
+set_target_properties(fpgad-vc PROPERTIES
```
tswhison: Module libraries should not set VERSION/SOVERSION

tools/base/fpgad/plugins/fpgad-xfpga/CMakeLists.txt
```diff
@@ -49,6 +49,9 @@ endif(CMAKE_BUILD_TYPE STREQUAL "Coverage")
 
 # Target properties
 set_property(TARGET fpgad-xfpga PROPERTY C_STANDARD 99)
+set_target_properties(fpgad-xfpga PROPERTIES
```
tswhison: Module libraries should not set VERSION/SOVERSION

PR #1467, opened 2020-02-21
add fedora spec file and cmake changes

anandaravuri:


opae.spec
```diff
@@ -0,0 +1,221 @@
...
+         board_vc \
+         fpgaconf \
+         fpgainfo \
+         fpgametrics \
+         userclk \
+         hello_fpga \
+         hello_events \
+         object_api \
+         mmlink  \
+         opae-c-ase \
```
tswhison: Please exclude opae-c-ase.

opae.spec
```diff
@@ -0,0 +1,221 @@
...
+
+
+
+mkdir -p %{buildroot}%{_sysconfdir}/systemd/system/
+
+%clean
+
+%post
+mkdir -p %{_sysconfdir}/ld.so.conf.d
+echo "" > %{_sysconfdir}/ld.so.conf.d/opae-c.conf
```
tswhison: Why do we create this file?  It's empty.

opae.spec

opae.spec
anandaravuri: Remove create file code
Fixed

PR #1468, opened 2020-02-21
Format LICENSE to 80 columns.

trixirt:
It is expected that any file should be read from an 80 column terminal.
Add only newlines and spaces.


trixirt:
The start copyright likely needs to be changed to whenever the project started.

PR #1476, opened 2020-02-27
Change libsafestr to a shared library

trixirt:
In general it is better to have shared libraries over static
unless there is a specific need for a static library.


southerngs:
I think we are planning to deprecate the safestring library in OPAE.  This isn't really a core part of OPAE and we never really wanted to redistribute it.  The static library was simply a convenience feature.  I think we should try to remove this instead of changing to dynamic library.  I'll let @tswhison comment though because he probably knows more.

tswhison:
Gabriel is right.  I'm in the process of removing it now.

PR #1485, opened 2020-03-05
safestr: replace safe_string APIs with standard C strings.

tswhison:


tswhison:
The tests are going to fail until we update the opae-libs subtree to incorporate those changes to remove safe_string, too.

PR #1487, opened 2020-03-09
fedora rpm spec file update

anandaravuri:


opae.spec
```diff
@@ -138,68 +129,68 @@ DESTDIR=%{buildroot}  cmake -DCOMPONENT=toolmmlink -P cmake_install.cmake
...
 
 %post
-ldconfig
-
-%postun
-ldconfig
-
-%pre
+mkdir -p %{_sysconfdir}/ld.so.conf.d
+echo "" > %{_sysconfdir}/ld.so.conf.d/opae-c.conf
```
tswhison: Why is this needed?

opae.spec
```diff
@@ -138,68 +129,68 @@ DESTDIR=%{buildroot}  cmake -DCOMPONENT=toolmmlink -P cmake_install.cmake
...
-
-%files samples
-%defattr(-,root,root,-)
 %{_bindir}/hello_fpga
-%{_bindir}/hello_events*
 %{_bindir}/object_api
+%{_bindir}/hello_events
 %{_bindir}/hello_cxxcore
+%{_bindir}/afu_json_mgr
+%{_bindir}/packager
```
tswhison: we are upstreaming packager?

anandaravuri: including in 1.4.1 release, converting rbf file to gbs.

opae.spec

PR #1489, opened 2020-03-10
Resolve issues with documentation build

nakulkorde:


nakulkorde:
@tswhison the documentation build is going to fail here since if no tag is detected, it resets the documentation branch to master, which doesn't have the right stuff for building documentation. Should I merge this in?

tswhison:
If you've verified it to work in your environment, then I'd say it's OK for it to go in.

nakulkorde:
Yea I verified it my environment.

PR #1497, opened 2020-03-18
rpmbuild: update version in spec file

r-rojo:


r-rojo:
won't update version until we are making 1.4.1 release.

PR #1500, opened 2020-03-18
Add man page building via sphinx

nakulkorde:


doc/sphinx/CMakeLists.txt
```diff
@@ -82,6 +86,16 @@ add_custom_target(docs ALL
   WORKING_DIRECTORY "${BINARY_BUILD_DIR}"
   COMMENT "Building HTML documentation with Sphinx")
 
+add_custom_target(manpages ALL
```
r-rojo: at some point it may be a good idea to look at how we can avoid duplication and wrap sphinx tasks in a cmake function/macro 😄

nakulkorde: Yea I agree.

PR #1501, opened 2020-03-20
add fedora_review check on RPMs created

r-rojo:
Add a script, docker_rpm.sh that does the following
  * create a .gitattributes file that lists what files/patterns to ignore when archiving
  * calls git archive to create a source tarball for rpmbuld
  * generates a buildrpm.sh script that:
    * builds the RPMs from the spec file
    * runs fedora-review
    * copies the artifacts to a temp directory
  * pulls a prebuilt docker image (opae/fedora_builder) that
    * uses fedora as a base
    * installs necessary packages to build opae and its RPMs
    * adds root to the mock group
    * sets up the rpmbuild tree
    * configures mock options to not use nspawn
    * copies spec file and source tarball from host
    * sets the entrypoint to buildrpm.sh
  * calls docker run with the new image and mounts a host dir to the temp dir

Add github workflow to call this script and get artifacts


scripts/docker_rpm.sh
```diff
@@ -0,0 +1,69 @@
...
+#
+
+cat > .gitattributes << EOF
+.git* export-ignore
+.travis.yml export-ignore
+opae.spec.in export-ignore
+opae.spec export-ignore
+opae-libs/external export-ignore
+opae-libs/tests export-ignore
+opae-libs/plugins/ase export-ignore
```
tswhison: ase is no longer here, right?

r-rojo:
@nakulkorde, Yes, I would like to expand on this and move away from travis.

PR #1502, opened 2020-03-23
Add Python 3 support for platforms scripts.

michael-adler:


michael-adler:
@OPAE/ase and @OPAE/tools please review.

I'm not sure what happened Travis. The failure was only after merging master. Is it a transient error?

michael-adler:
Yes, I tried it on 2.7. Might have missed something, but when I did the basic functionality still worked.

PR #1514, opened 2020-04-09
tests: add opae sub directory function to cmakelist

anandaravuri:


anandaravuri:
> Does the RPM build process fail without this? I'm wondering why we didn't see an issue with this before now?

Create RPM script  remove ras/hello_events/object_api  test cases, opae_xpz_fun() ignores folder during build

PR #1519, opened 2020-04-13
Fix another Python 3 bug

michael-adler:


michael-adler:
Tested with both Python 2 and 3.

PR #1526, opened 2020-04-15
update opae rpm spec file 

anandaravuri:


opae.spec
```diff
@@ -191,11 +191,11 @@ DESTDIR=%{buildroot}  cmake -DCOMPONENT=jsonschema -P cmake_install.cmake
 
 
 %changelog
-* Tue Dec 17 2019 Korde Nakul <nakul.korde@intel.com> 1.4.0-1
```
nakulkorde, line 194: I think in changelog, we should not delete the old data. We should keep them as it is and the append new changelog data after that.

anandaravuri: fixed

PR #1542, opened 2020-04-28
opae: update OPAE spec file

anandaravuri:


tools/extra/pac_hssi_config/pac_hssi_config.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#! /usr/bin/env python3
```
tswhison, line 1: These scripts should go in a separate pull request.

anandaravuri: fixed , move to another pull request

PR #1555, opened 2020-05-13
1.4.1/fpgaport

trixirt:
I am working through the VM usecase. 
This tool is in 1.4.0 but missing in 1.4.1
Instead of writing a new tool, I generalized the old tool to work with both drivers.


trixirt:
A change is master does not help 1.4.1.
Only intel driver supports darby creek and vista creek.  So while opae-sdk works with both, this tool does not.

PR #1568, opened 2020-05-21
tools: add DFL driver mac address

anandaravuri:


scripts/cover.sh
```diff
@@ -43,6 +43,7 @@ lcov --remove coverage.total \
     '*opae-libs/plugins/xfpga/usrclk/**' \
     '*tests/**' \
     '*tools/extra/c++utils/**' \
+    '*tools/libboard/board_a10gx/**' \
```
tswhison, line 46: are you not planning to test this one?

anandaravuri: RC card board module , we are not going to support RC card in next Release.

tswhison: Can we do this in a separate pull request?

anandaravuri: fixed, removed changes in merge request.

PR #1581, opened 2020-06-10
packaging: add OPAE Debian package creation.

tswhison:


packaging/opae/version
```diff
@@ -0,0 +1,28 @@
...
+# IMPLIED WARRANTIES OF  MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT  SHALL THE COPYRIGHT OWNER  OR CONTRIBUTORS BE
+# LIABLE  FOR  ANY  DIRECT,  INDIRECT,  INCIDENTAL,  SPECIAL,  EXEMPLARY,  OR
+# CONSEQUENTIAL  DAMAGES  (INCLUDING,  BUT  NOT LIMITED  TO,  PROCUREMENT  OF
+# SUBSTITUTE GOODS OR SERVICES;  LOSS OF USE,  DATA, OR PROFITS;  OR BUSINESS
+# INTERRUPTION)  HOWEVER CAUSED  AND ON ANY THEORY  OF LIABILITY,  WHETHER IN
+# CONTRACT,  STRICT LIABILITY,  OR TORT  (INCLUDING NEGLIGENCE  OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+PROJECT_VERSION='1.4.0'
```
HananiaAmir, line 27: It looks like that it was 1.4.1-1 - Since the new version 1.4.0 is lower, I just want to verify that 1.4.0 is the correct version.

tswhison: Right, this file needs to match what opae-libs/CMakeLists.txt says, and it currently says 1.4.0.  Otherwise, the .so versioning is off and debuild errors out.

PR #1588, opened 2020-06-19
tools: Static code analysis -  Buffer overflow

anandaravuri:
-Buffer Overflow - Array Index Out of Bounds

### DIFF
--- a/tools/libboard/board_d5005/board_d5005.c
+++ b/tools/libboard/board_d5005/board_d5005.c
@@ -46,7 +46,7 @@
 #include "board_d5005.h"
 #include "../board_common/board_common.h"
 
-#define MACADDR_LEN 18
+#define MACADDR_LEN 19
 
 // DFL SYSFS
 #define DFL_SYSFS_BMCFW_VER                     "dfl-fme*/spi-altera*/spi_master/spi*/spi*/bmcfw_version"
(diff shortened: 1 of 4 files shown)


tools/libboard/board_d5005/board_d5005.c
```diff
@@ -127,7 +127,7 @@ fpga_result read_max10fw_version(fpga_token token, char *max10fw_ver, size_t len
 		return FPGA_INVALID_PARAM;
 	}
 
-	res = read_sysfs(token, DFL_SYSFS_MAX10_VER, buf, FPGA_VAR_BUF_LEN);
+	res = read_sysfs(token, DFL_SYSFS_MAX10_VER, buf, FPGA_VAR_BUF_LEN -1);
```
tswhison, line 130: Could we make the spacing around the -1 consistent?  eg, line 160 uses
```c
MACADDR_LEN-1
```
Let's converge on the form that uses a space on either side of the -
```c
X - 1
```

tools/libboard/board_n3000/board_n3000.c
```diff
@@ -541,6 +541,12 @@ fpga_result print_mac_info(fpga_token token)
 	mac.byte[1] = buf[4];
 	mac.byte[2] = buf[3];
 	mac.byte[3] = 0;
+
+	if (n < 0 || n > 0xFFFF) {
```
tswhison, line 545: Should we also use endptr to verify the string conversion from line 538?

PR #1595, opened 2020-07-06
opae.admin: update spi path for dfl drivers

tswhison:
The glob to find the location of the spi path in sysfs has changed
to dfl-fme.*.*/subdev_spi_altera.*.auto/... Update opae.admin to use
the new glob pattern for dfl drivers.

### DIFF
--- a/python/opae.admin/opae/admin/fpga.py
+++ b/python/opae.admin/opae/admin/fpga.py
@@ -208,7 +208,7 @@ class fme(region):
     def spi_bus(self):
         if os.path.basename(self.sysfs_path).startswith('dfl'):
             return self.find_one('dfl-fme.*.*/'
-                                 'subdev_spi_altera.*.auto/'
+                                 '*spi*/'
                                  'spi_master/spi*/spi*')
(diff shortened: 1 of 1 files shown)


python/opae.admin/opae/admin/fpga.py
```diff
@@ -207,7 +207,7 @@ def i2c_bus(self):
     @property
     def spi_bus(self):
         if os.path.basename(self.sysfs_path).startswith('dfl'):
-            return self.find_one('dfl-fme.*.*/spi*/spi_master/spi*/spi*')
+            return self.find_one('dfl-fme.*.*/subdev_spi_altera.*.auto/spi_master/spi*/spi*')
```
fpgamatt, line 210: This change did work.  You could also use *spi* instead of subdev_spi_altera.*.auto

r-rojo: As @fpgamatt says, you can probably use:
```Python
            return self.find_one('dfl-fme.*.*/*spi*/spi_master/spi*/spi*')
```

fpgamatt:
Looks good to me, but I don't see an approval button.

PR #1597, opened 2020-07-10
fpgasupdate: update per new driver flow

tswhison:
The driver flow was changed such that reading the "error" sysfs
node in other than the "idle" state results in EBUSY. To fix, rather
than polling "error" periodically, only check it on state transition
and only after having verified the "idle" status.


tswhison:
btw, this change was verified with the latest driver.

rweight:
Looks good to me. Thanks!

PR #1605, opened 2020-07-28
hssi: add sample app

tswhison:
This pull request is to merge branch tswhison/hssi to feature/hssi.


samples/hssi/hssi.cpp
```diff
@@ -0,0 +1,654 @@
...
+#define MBOX_TIMEOUT 1000
+
+bool parse_filter_address(properties::ptr_t filter, const char *sbdf)
+{
+  regex_t re;
+  regmatch_t matches[5];
+
+  memset(&matches, 0, sizeof(matches));
+
+  std::string sbdf_expr("([0-9a-fA-F]{4}):([0-9a-fA-F]{2}):([0-9a-fA-F]{2})\\.([0-7])");
```
r-rojo, line 96: You can simplify this by using this pattern:
```c++
 std::string sbdf_expr("(([0-9a-fA-F]{4}):)?([0-9a-fA-F]{2}):([0-9a-fA-F]{2})\\.([0-7])");
```

That gives you two more groups but they'll start at index 2.
```c++
if (matches[2].rm_so > -1) {
  std::string segment(sbdf + matches[2].rm_so, matches[2].rm_eo - matches[2].rm_so);
  filter->segment = static_cast<uint16_t>(std::stoul(segment, nullptr, 16));
} else {
  filter->segment = 0;
}

```

samples/hssi/hssi.cpp
```diff
@@ -0,0 +1,654 @@
...
+
+  if (tokens.size() > 1) {
+    std::cerr << "warning: More than one accelerator found." << std::endl;
+  }
+
+  return handle::open(tokens[0], 0);
+}
+
+#define INVALID_MAC 0xffffffffffffffffULL
+uint64_t mac_string_to_bits(const std::string &str)
```
r-rojo, line 171: Are there any Linux calls to do this?

samples/hssi/hssi.cpp
tswhison: As is it reads a little more cleanly, I think.

samples/hssi/hssi.cpp
tswhison: I looked but didn't find any.  Maybe @fpgamatt can point one out?  There are routines for the IPv4 dotted quad conversion, but didn't see any for ethernet addresses.

fpgamatt: ether_ntoa() and ether_aton()

tswhison: Thanks

PR #1628, opened 2020-08-10
dummy_afu: add count and timeout args

r-rojo:
* Add count argument that runs the test command given number of times
* Add timeout argument that waits up to given timeout (msec) for test
  command

### DIFF
--- a/tests/dummy_afu/test_dummy_afu.cpp
+++ b/tests/dummy_afu/test_dummy_afu.cpp
@@ -198,10 +198,10 @@ TEST_P(dummy_afu_p, main_sleep_timeout) {
 TEST_P(dummy_afu_p, main_sleep_notimeout) {
   args_.push_back(strdup("dummy_afu"));
   args_.push_back(strdup("-t"));
-  args_.push_back(strdup("100"));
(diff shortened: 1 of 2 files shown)


tests/dummy_afu/test_dummy_afu.cpp
```diff
@@ -124,6 +159,51 @@ TEST_P(dummy_afu_p, main_mmio) {
   EXPECT_EQ(0, app_->main(args_.size(), const_cast<char**>(args_.data())));
 }
 
+/*
+ * @test       main_mmio_count
+ * @brief      Test: test main with mmio subcommand and count > 1
+ * @details    Test the main entry point for dummy_afu
+ */
+TEST_P(dummy_afu_p, main_mmio_count) {
```
tswhison, line 167: Is something calling free() on all these strdup'ed strings?

r-rojo: Yes, line 131.
https://github.com/OPAE/opae-sdk/blob/ded0b4602b6c014542de6297a6783a2ad6d14db8/tests/dummy_afu/test_dummy_afu.cpp#L121-L133

PR #1632, opened 2020-08-12
hssi: enable 64-bit register access

tswhison:
Enable 64-bit writes to TRAFFIC_CTRL_CMD and TRAFFIC_CTRL_DATA.


samples/hssi/hssi_mbox.h
```diff
@@ -33,26 +33,19 @@
 #include <stdint.h>
 #include <time.h>
 
-#if 0
 #define TRAFFIC_CTRL_CMD  0x0030
+#define TRAFFIC_CTRL_DATA 0x0038
+
+#if 1
```
r-rojo, line 39: ?

tswhison: This is because I'm preserving both the 32-bit and the 64-bit mailbox accessors.

PR #1634, opened 2020-08-13
Use cmake variables to control building of extra tools.

trixirt:
OPAE_BUILD_EXTRA_TOOLS is a top level control for building
all of the extra tools.  Then each tool has its own variable
controlling its building.  Because fpgadiag depends on c++utils
override the user choice to disable c++utils if fpgadiag is
being built.


trixirt:
Depends on https://github.com/OPAE/opae-libs/pull/74

r-rojo (approved):

PR #1653, opened 2020-09-01
opae.admin: refactor secure device sysfs paths

tswhison:
The sysfs paths to the secure device mananger sysfs nodes
changed from m10bmc-secure.*.auto for both D5005 and N3000
to d5005bmc-secure.*.auto and n3000bmc-secure.*.auto,
respectively. The affected tools are fpgasupdate and rsu.

Update the RSU mechanism to observe the new sysfs layout.
The secure manager driver implements two sysfs entries for
RSU: available_images and image_load. available_images contains
a null-terminated string describing the available modes.

$ cat available_images
bmc_factory bmc_user retimer_fw

Writing the desired string to image_load causes the driver
to perform the associated RSU operation.

### DIFF
--- a/python/opae.admin/opae/admin/fpga.py
+++ b/python/opae.admin/opae/admin/fpga.py
@@ -216,6 +216,17 @@ class fme(region):
         return self.find_one('spi*/spi_master/spi*/spi*')
 
     @property
+    def ifpga_sec(self):
+        spi = self.spi_bus
+        if spi:
+            ifpga_sec = spi.find_one(
+                'd5005bmc-secure.*.auto/ifpga_sec_mgr/ifpga_sec*')
+            if not ifpga_sec:
+                ifpga_sec = spi.find_one(
+                    'n3000bmc-secure.*.auto/ifpga_sec_mgr/ifpga_sec*')
+            return ifpga_sec
+
+    @property
     def altr_asmip(self):
         return self.find_one('altr-asmip*.*.auto')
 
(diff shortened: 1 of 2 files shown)


python/opae.admin/opae/admin/fpga.py
```diff
@@ -380,17 +380,16 @@ class fpga_base(class_node):
...
-        (0x8086, 0x0b30): {'fpga': {'user': 1,
-                                    'factory': 0},
-                           'bmcimg': {'user': 0,
-                                      'factory': 1}},
-        (0x8086, 0x0b2b): {'fpga': {'user': 0,
-                                    'factory': 1},
-                           'bmcimg': {'user': 1,
-                                      'factory': 0}},
-        (0x8086, 0x09c4): {'fpga': {'user': 0}}
+        (0x8086, 0x0b30): {'bmcimg': {'user': 0,
```
anandaravuri, line 385: BOOT_PAGES  can be removed, valid for Production driver

tswhison: There is still logic that accesses these structures.  I have instrumented the table so that the code flows through unscathed.

python/opae.admin/opae/admin/fpga.py
```diff
@@ -442,34 +441,55 @@ def supports_rsu(self):
...
+            self.log.exception(msg)
+            raise TypeError(msg)
+
+        available_images = ifpga_sec.find_one('update/available_images').value
+        image_load = ifpga_sec.find_one('update/image_load')
 
-        node = self.fme.spi_bus.find_one(node_path)
-        node.value = page
+        if boot_type == 'bmcimg':
+            if factory:
```
rweight, line 476: I'm not a python expert...
If factory an integer value, containing the value passed to the tool (0 or 1). Unfortunately, the values for factory/user are swapped between VC and DC. It is not clear to me if this logic is taking that into acount.

tswhison: Good call.  Actually, factory is being passed in from the top-level command option (-f) to the rsu script.
https://github.com/OPAE/opae-sdk/pull/1653/files#diff-8a2882609782f926ff14ff3440bf3167R153

PR #1657, opened 2020-09-02
docs: update the fpgad doc

figozhang:
docs: update the fpgad doc

### DIFF
--- a/doc/src/fpga_tools/fpgad/fpgad.md
+++ b/doc/src/fpga_tools/fpgad/fpgad.md
@@ -5,15 +5,13 @@
 `fpgad [--socket=<sock>] [--null-bitstream=<file>]`
 
 ## DESCRIPTION ##
-```fpgad``` periodically monitors and reports the error status reflected in the device driver's error status sysfs files.
-```fpgad``` establishes the channel to communicate events to the Open Programmable Accelerator Engine (OPAE) application. 
-```fpgad``` programs a NULL bitstream in response to an AP6 (power) event. ```fpgad``` is only available on the Integrated FPGA
-Platform. You cannot run ```fpgad``` on the PCIe&reg; Accelerator Card (PAC).
+fpgad monitors the device sensors, checking for sensor values that are out of the prescribed range. 
 
-If your system does not support interrupts, you must run ```fpgad``` before the API calls `fpgaRegisterEvent` and
-`fpgaUnregisterEvent` can succeed.
+When any of the sensors is detected to be out of bounds, fpgad will focus on keeping the server from rebooting by masking PCIE AER, and send a message to system administrator. System administrator can take further actions like stop the application and stop the FPGA, but fpgad just focus on monitor the sensors and will not take any cooling actions. 
 
-Use SIGINT to stop ```fpgad```.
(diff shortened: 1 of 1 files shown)


doc/src/fpga_tools/fpgad/fpgad.md
```diff
@@ -5,15 +5,16 @@
...
 ## DESCRIPTION ##
-```fpgad``` periodically monitors and reports the error status reflected in the device driver's error status sysfs files.
-```fpgad``` establishes the channel to communicate events to the Open Programmable Accelerator Engine (OPAE) application. 
-```fpgad``` programs a NULL bitstream in response to an AP6 (power) event. ```fpgad``` is only available on the Integrated FPGA
-Platform. You cannot run ```fpgad``` on the PCIe&reg; Accelerator Card (PAC).
+fpgad monitors the device sensors, checking for sensor values that are out of the prescribed range. 
 
-If your system does not support interrupts, you must run ```fpgad``` before the API calls `fpgaRegisterEvent` and
-`fpgaUnregisterEvent` can succeed.
+When any of the sensors is detected to be out of bounds, fpgad responds by removing the OPAE kernel driver for a configurable cool-down period.
```
tswhison, line 10: This statement is not true.

doc/src/fpga_tools/fpgad/fpgad.md
```diff
@@ -5,15 +5,16 @@
...
 
-Use SIGINT to stop ```fpgad```.
+When it is being removed, the OPAE kernel driver’s shutdown path will signal any application processes that are actively using an AFU to exit, and it will reset the hardware devices, allowing them to reach a quiescent state.
```
tswhison, line 12: This statement describes the behavior of a patch that was added to the production driver some time ago.  This statement is not true of the dfl driver.

doc/src/fpga_tools/fpgad/fpgad.md
```diff
@@ -5,15 +5,16 @@
...
+Once the cool-down period has expired, fpgad will force a PCIe bus rescan, reloading the OPAE kernel driver to resume normal operation.  Any application processes that were forced to exit by the OPAE kernel driver will need to be manually restarted.
```
tswhison, line 13: We shouldn't mention about restarting applications.  None will have been stopped by the driver.

doc/src/fpga_tools/fpgad/fpgad.md
figozhang: When any of the sensors is detected to be out of bounds, fpgad will focus on keeping the server from rebooting (by masking AER) if/when the VC card is kicked off the PCIe bus by the BMC if a threshold is exceeded.

doc/src/fpga_tools/fpgad/fpgad.md
figozhang: yes, i will remove it

doc/src/fpga_tools/fpgad/fpgad.md
figozhang: ok, i will fix it

tswhison (approved):
👍

PR #1661, opened 2020-09-03
hssi: add 100G AFU test

tswhison:


fpgamatt:
Hi Tim,

We should not enable this test in the test run until we have a mechanism for loading the 100 design.  We may have to make this test a remote test because we'll have to reload a sof with the 100g mac in the FIM.

It is great to have this code in our back pocket and backed up in github, but we'll need to make modifications before we turn on the test.

tswhison:
Thanks @fpgamatt.  These are the changes for the hssi application headed into the external repo.  Will we need to also modify the hssi app when bringing up the 100G tests?  If so, we can let this one sit until those edits are made.

fpgamatt (approved):

r-rojo (approved):

PR #1666, opened 2020-09-04
hssi: add pause after loopback enable

tswhison:
Empirical evidence suggests that a brief pause is necessary after
$ ethtool --features <ifc> loopback on

Without the pause, it was observed that the number of rx packets
(serial test) can be 0 when num_packets (100) was expected.

### DIFF
--- a/samples/hssi/hssi_cmd.h
+++ b/samples/hssi/hssi_cmd.h
@@ -81,6 +81,7 @@ public:
     else
       cmd += std::string(" loopback off");
     run_process(cmd);
+    usleep(1000);
   }
 
   template <typename X>


fpgamatt (approved):
I thought we already had a delay in there, but as we saw previously, this fixes the problem.

tswhison:
> I thought we already had a delay in there, but as we saw previously, this fixes the problem.

The delay was actually submitted as part of #1661 and hasn't merged yet.  Spoke with @dmlomartire about last nights failures, and agreed to add it for tonight.

r-rojo (approved):

PR #1667, opened 2020-09-04
build: consolidate inclusion of externals

tswhison:
During development, both dummy_afu and hssi were loading CLI11
and spdlog. Move those macro invocations to a single location
(external CMakeLists.txt) and process external/ prior to the
samples so that the variables are available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,7 @@ endif()
 ## Sub-projects ############################################################
 ############################################################################
 opae_add_subdirectory(opae-libs)
+opae_add_subdirectory(external)
 opae_add_subdirectory(platforms)
 opae_add_subdirectory(tools)
 opae_add_subdirectory(samples)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ endif()
 ## Sub-projects ############################################################
 ############################################################################
 opae_add_subdirectory(opae-libs)
+opae_add_subdirectory(external)
```
r-rojo, line 62: 🤓

samples/dummy_afu/CMakeLists.txt
```diff
@@ -24,32 +24,6 @@
 ## ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
 ## POSSIBILITY OF SUCH DAMAGE.
 
-set(CLI11_URL
```
r-rojo, line 27: Can we keep URL and TAG variables configurable as cmake variables?

r-rojo (approved):

samples/dummy_afu/CMakeLists.txt
tswhison: I saw that, but they were only being used in one place.  Where else can they be used?

anandaravuri (approved):

r-rojo (approved):

PR #1675, opened 2020-09-10
hssi: increase the mbox timeout to 1 second..

tswhison:
and add printouts for mbox exceptions before throwing them.


fpgamatt (approved):
This makes sense.  If the mailbox fails, everything behind it is broken.

anandaravuri (approved):

PR #1700, opened 2020-10-05
afu-test: make afu-test a shared library

r-rojo:
reverting header only library configuration
because cmake 2.8 doesn't support INTERFACE targets


tswhison (approved):

asgardkm (approved):

r-rojo:
There were rpm build errors. I am abandoning this PR in favor with another that can successfully build RPMs using this new library.

PR #1705, opened 2020-10-16
release: update opae release branch targets

asgardkm:
Update the following opae dependencies for opae-sdk `release/2.0.1-1` so that said dependencies point to a snapshot release branch/tag, and not to continue tracking `master`.
- `opae-libs`
  - Subtree has been updated to handle `opae-test` and `opae-sim`
- `opae-legacy`

This commit should not be pulled back into `master`, as the `master` branch should naturally track the latest HEAD of each opae dependency. Rather, this commit will be the tag target for release `2.0.1-1`.


asgardkm:
Closing - the git history diff between the two branches seems off (the version change should already be in `release/2.0.1-1`).

Will open new PR shortly.

PR #1715, opened 2020-10-22
mmlink: fix mmlink build

r-rojo:
* set c standard to c99 for mml-stream
* set library versions (VERSION/SOVERSION) for mml-srv and mml-stream
* set compiler attribute for allowing aliasing in streaming/common.c


tswhison (approved):

anandaravuri (approved):

asgardkm (approved):
build on el7 - thanks!

PR #1716, opened 2020-10-22
docker: install kernel headers from elrepo

r-rojo:
Change docker file to:
* install elrepo and import its gpg key
* install kernel-ml-headers (mainline)

### DIFF
--- a/docker/centos7.6/Dockerfile
+++ b/docker/centos7.6/Dockerfile
@@ -1,4 +1,7 @@
 FROM centos:7.6.1810
+RUN yum install -y https://www.elrepo.org/elrepo-release-7.0-4.el7.elrepo.noarch.rpm
+RUN rpm --import https://www.elrepo.org/RPM-GPG-KEY-elrepo.org
+RUN yum --enablerepo=elrepo-kernel install -y kernel-ml-headers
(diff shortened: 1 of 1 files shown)


r-rojo:
I ran the rpm build jobs using this branch and they passed:
https://github.com/OPAE/opae-sdk/actions/runs/321317938

asgardkm:
@r-rojo is the addition of the el headers for pulling in the `vfio.h` headers for compilation?

I don't see a log difference between the CI build logs for this PR and prior PRs (such as https://github.com/OPAE/opae-sdk/runs/1288414341?check_suite_focus=true#step:6:205)

However, I think that the output string in CMake compilation (which would allow making this comparision) is being suppressed here:
https://github.com/OPAE/opae-sdk/blob/master/tools/extra/fpgadiag/CMakeLists.txt#L130

Where `else( SUPPORTS_VFIO)` should be `else()`

r-rojo:
@asgardkm , yes. This is for building the vfio related libraries/applications (with the caveat being that this code will not work on older kernels because the data structures/constants are different).

This is for the CI jobs that build RPMs and run after a PR is merged into master. The CI jobs you mentioned shouldn't change.

The else syntax works either way in CMake.

tswhison (approved):
nice

asgardkm (approved):

PR #1717, opened 2020-10-22
fpgainfo: add support for revision 2 of port errors

fpgamatt:
With Intel OFS, the definition of the port error bits have
changed to reflect the transition from CCI-P to AXI-S TLP.
The fpgainfo command now checks the revision sysfs entry
to use the correct port error bit definitions.

### DIFF
--- a/tools/fpgainfo/errors.c
+++ b/tools/fpgainfo/errors.c
@@ -177,6 +177,26 @@ static const char *const PORT_ERROR[PORT_ERROR_COUNT] = {
 	"Tx Channel 1 byte_len and byte_start should be zero when mode is not eMOD_BYTE"
 };
 
+#define REV2_PORT_ERROR_COUNT 16
+static const char *const REV2_PORT_ERROR[REV2_PORT_ERROR_COUNT] = {
+	"Tx valid violation",
+	"Tx mwr insufficient data",
+	"Tx mwr data payload overrun",
+	"mmio insufficient data",
+	"mmio data payload overrun",
+	"mmio read while reset",
+	"mmio write while reset",
+	"mmio timeout",
+	"unexpected mmio response",
+	"tag occupiied",
+	"unaligned address",
+	"max tag",
+	"max read request size",
+	"max payload",
+	"malformed TLP",
+	"Tx request couter overflow",
+};
+
 /*
  * errors command configuration, set during parse_args()
  */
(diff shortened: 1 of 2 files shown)


tools/fpgainfo/errors.c
```diff
@@ -316,6 +337,29 @@ fpga_result errors_filter(fpga_properties *filter, int argc, char *argv[])
...
+		OPAE_MSG("Failed to get token Object");
+		return res;
+	}
+
+	res = fpgaObjectRead64(fpga_object, value, 0);
+	if (res != FPGA_OK) {
+		OPAE_MSG("Failed to Read object ");
+	}
+
+	res = fpgaDestroyObject(&fpga_object);
```
r-rojo, line 356: If line 351 fails, we lose the error code here.

tools/fpgainfo/errors.c
```diff
@@ -400,8 +444,15 @@ static void print_errors_info(fpga_token token, fpga_properties props,
 			       error_value);
 
 			if (!strcmp(errinfos[i].name, "Errors")) {
-				size = PORT_ERROR_COUNT;
-				error_string = PORT_ERROR;
+				uint64_t revision = 0;
+				get_port_error_revision(token, &revision);
```
r-rojo, line 448: Do we or should we bail here if we can't successfully determine the revision?

tools/fpgainfo/errors.c
fpgamatt: I will fix that cut n paste error.

tools/fpgainfo/errors.c
fpgamatt: My theory is that if the revision cannot be determined, then assume the original version.

I suppose another approach is that we don't decode at all if there is no revision.

r-rojo: Either way (if we decode or not), I think we should print out an error/warning to stderr indicating that revision couldn't be determined.

tools/fpgainfo/errors.c
```diff
@@ -316,6 +337,29 @@ fpga_result errors_filter(fpga_properties *filter, int argc, char *argv[])
 	return res;
 }
 
+static fpga_result get_port_error_revision(fpga_token token, uint64_t *value)
+{
+	fpga_result res = FPGA_OK;
+	fpga_object fpga_object;
+
+	res = fpgaTokenGetObject(token,"error*/revision" , &fpga_object, FPGA_OBJECT_GLOB);
```
tswhison, line 345: s/token,"error*/revision" ,/token, "error*/revision",/

tools/fpgainfo/errors.c
```diff
@@ -400,8 +445,27 @@ static void print_errors_info(fpga_token token, fpga_properties props,
 			       error_value);
 
 			if (!strcmp(errinfos[i].name, "Errors")) {
-				size = PORT_ERROR_COUNT;
-				error_string = PORT_ERROR;
+				uint64_t revision = 0;
+				res = get_port_error_revision(token, &revision);
+				if (res != FPGA_OK) {
+					printf("could not find port error revision - skipping decode\n");
```
r-rojo, line 451: Should probably use OPAE_ERR macro here.

r-rojo (approved):

PR #1729, opened 2020-10-27
mmlink: allow graceful termination

r-rojo:
* add terminate function in remote_dbg interface
* add server_terminate in streaming server implementation
  * this function sets a terminate flag to 1 which can be checked if the
  server should terminate if interrupted.
* use `sigaction` instead of `signal` when setting up signal handler
  * `signal` is a wrapper around `sigaction` with some default flags
  set, in particular, SA_RESTART. When SA_RESTART is not set, blocking
  calls like accept will be interrupted. This allows the server
  listening routine to break out of the loop and check if it should
  terminate.


tools/extra/mmlink/remote_dbg/streaming/server.c.orig
```diff
@@ -0,0 +1,899 @@
+// Copyright(c) 2020, Intel Corporation
```
tswhison, line 1: Did you intend to submit this file?

r-rojo: Oops. No. I'll remove. Thanks.

tswhison (dismissed):

tswhison (approved):

PR #1731, opened 2020-10-27
opae.admin: added regmap_debugfs.py

fpgamatt:
Add a tool to help access the debugfs interface to regmaps
created in the linux kernel.

### DIFF
--- a/python/opae.admin/opae/admin/tools/regmap_debugfs.py
+++ b/python/opae.admin/opae/admin/tools/regmap_debugfs.py
@@ -1,0 +1,163 @@
+"""
+Copyright(c) 2020, Intel Corporation
+
+Redistribution  and  use  in source  and  binary  forms,  with  or  without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of  source code  must retain the  above copyright notice,
+  this list of conditions and the following disclaimer.
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+* Neither the name  of Intel Corporation  nor the names of its contributors
+  may be used to  endorse or promote  products derived  from this  software
+  without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,  BUT NOT LIMITED TO,  THE
+IMPLIED WARRANTIES OF  MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED.  IN NO EVENT  SHALL THE COPYRIGHT OWNER  OR CONTRIBUTORS BE
+LIABLE  FOR  ANY  DIRECT,  INDIRECT,  INCIDENTAL,  SPECIAL,  EXEMPLARY,  OR
+CONSEQUENTIAL  DAMAGES  (INCLUDING,  BUT  NOT LIMITED  TO,  PROCUREMENT  OF
+SUBSTITUTE GOODS OR SERVICES;  LOSS OF USE,  DATA, OR PROFITS;  OR BUSINESS
+INTERRUPTION)  HOWEVER CAUSED  AND ON ANY THEORY  OF LIABILITY,  WHETHER IN
+CONTRACT,  STRICT LIABILITY,  OR TORT  (INCLUDING NEGLIGENCE  OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+"""
+
+import argparse
+import os.path
+import sys
+
+
+class RegmapDebugfs():
+    """RegmapDebugfs is a class for accessing the debugfs
+       inteface of regmaps in the linux kernel"""
+
+    def __init__(self, regmapdir):
+
+        self.registers_path = os.path.join(regmapdir, 'registers')
+        self.range_path = os.path.join(regmapdir, 'range')
+        self.ranges = []
+        self.record_len = 0
+        self.register_stride = 0
+
+    def get_register_info(self):
+        """get information about regmap"""
+        with open(self.registers_path, 'r') as _file:
+            line0 = _file.readline()
+            line1 = _file.readline()
+            self.record_len = len(line0)
+            list0 = line0.strip().split(': ')
+            list1 = line1.strip().split(': ')
+            self.register_stride = int(list1[0], 16) - int(list0[0], 16)
+
+    def get_range_info(self):
+        """get the list of address ranges for the regmap"""
+        with open(self.range_path, 'r') as _file:
+            for line in _file.readlines():
+                list0 = line.strip().split('-')
+                range_dict = {
+                    'min': int(list0[0], 16),
+                    'max': int(list0[1], 16),
+                    'num_records': ((int(list0[1], 16) - int(list0[0], 16)) /
+                                    self.register_stride) + 1,
+                }
+                self.ranges.append(range_dict)
+
+    def reg_read(self, addr):
+        """reg_read returns the value read from the given
+           address in the regmap"""
+
+        num_records = 0
+        for range_dict in self.ranges:
+            if range_dict['min'] <= addr <= range_dict['max']:
+                offset = (addr - range_dict['min'])
+                num_records += offset / self.register_stride
+                break
+            else:
+                num_records += range_dict['num_records']
+        else:
+            sys.exit("bad read address")
+
+        with open(self.registers_path, 'r') as _file:
+            _file.seek(num_records * self.record_len)
+            line0 = _file.readline()
+
+        list0 = line0.strip().split(': ')
+        raddr = int(list0[0], 16)
+
+        if raddr != addr:
+            msg = "bad addr read back 0x{:x} != 0x{:x}".format(raddr, addr)
+            sys.exit(msg)
+
+        return int(list0[1], 16)
+
+    def reg_write(self, addr, val):
+        """reg_write writes val to the addr in the regmap"""
+
+        for range_dict in self.ranges:
+            if range_dict['min'] <= addr <= range_dict['max']:
+                break
+
+        else:
+            sys.exit("bad write address")
+
+        with open(self.registers_path, 'rb+') as _file:
+            _file.write("{:x} {:x}".format(addr, val).encode())
+
+
+def is_regmap_dir(regmapdir):
+    """check for a valid regmap directory in debugfs"""
+    if not os.path.isdir(regmapdir):
+        msg = f"{regmapdir} is not a valid directory"
+        raise argparse.ArgumentTypeError(msg)
+
+    range_path = os.path.join(regmapdir, 'range')
+    if not os.path.isfile(range_path):
+        msg = f"{range_path} is not a valid file"
+        raise argparse.ArgumentTypeError(msg)
+
+    registers_path = os.path.join(regmapdir, 'registers')
+    if not os.path.isfile(registers_path):
+        msg = f"{registers_path} is not a valid file"
+        raise argparse.ArgumentTypeError(msg)
+
+    return regmapdir
+
+
+def main():
+    """main entry point"""
+
+    descr = "access to debugfs interface of regmaps"
+    parser = argparse.ArgumentParser(description=descr)
+
+    parser.add_argument('regmap', type=is_regmap_dir,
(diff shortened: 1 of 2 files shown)


tswhison (commented):
👍

python/opae.admin/opae/admin/tools/regmap_debugfs.py
```diff
@@ -0,0 +1,149 @@
...
+import os.path
+
+
+class RegmapDebugfs():
+    """RegmapDebugfs is a class for accessing the debugfs
+       inteface of regmaps"""
+
+    def __init__(self, regmapdir):
+        if not os.path.isdir(regmapdir):
+            raise Exception("bad regmap dir")
```
dasm, line 39: It is discouraged to `raise Exception` if function is directed towards end user.
Probably better would be to use `sys.exit("Bad regmap dir")`

Additionally, you could check for `isdir` already in argument by doing something similar to:
```
def check_dir(path):
    if os.path.isdir(path):
        return path
    else:
        raise argparse.ArgumentTypeError(f"{path} is not a valid directory")
```

It would work similar to this:
```
import argparse
import os


def check_path(path):
    if not os.path.isdir(path):
        raise argparse.ArgumentTypeError(f"{path} is not a valid directory path")
    return path


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument('regmap', type=check_path)

    args = parser.parse_args()

if __name__ == '__main__':
    main()
```

python/opae.admin/opae/admin/tools/regmap_debugfs.py
```diff
@@ -0,0 +1,149 @@
...
+
+        _range = os.path.join(regmapdir, 'range')
```
dasm, line 41: Underscore notation `_range` in Python is considered as not used variable [PEP8](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles). I see it being used on line 60.
Maybe you could reconsider changing its name?

python/opae.admin/opae/admin/tools/regmap_debugfs.py
```diff
@@ -0,0 +1,149 @@
...
+        if not os.path.isfile(_range):
+            raise Exception("could not find range entry")
+
+        _registers = os.path.join(regmapdir, 'registers')
+        if not os.path.isfile(_registers):
+            raise Exception("could not find registers entry")
+
+        self._registers = _registers
+
+        with open(self._registers, 'r') as _file:
```
dasm, line 51: All kind of operations should be handled outside of `def __init__` function.
You should avoid doing anything what could take long time, in a body of initialization.

Maybe you could do something like:
```
regmap = RegmapDebugfs()
regmap.retrieve_information()
regmap.read(addr)
```

This way, you could separate init from file operations.

python/opae.admin/opae/admin/tools/regmap_debugfs.py
```diff
@@ -0,0 +1,149 @@
...
+CONTRACT,  STRICT LIABILITY,  OR TORT  (INCLUDING NEGLIGENCE  OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+"""
+
+import argparse
+import os.path
+
+
+class RegmapDebugfs():
```
dasm, line 33: What does "regmapdebugfs" mean?
Is there better name for this class?

python/opae.admin/opae/admin/tools/regmap_debugfs.py
```diff
@@ -0,0 +1,149 @@
...
+            list0 = line0.strip().split(': ')
+            list1 = line1.strip().split(': ')
+            self._register_stride = int(list1[0], 16) - int(list0[0], 16)
+
+        self.ranges = []
+        with open(_range, 'r') as _file:
+            for line in _file.readlines():
+                list0 = line.strip().split('-')
+                range_dict = {}
+                _min = int(list0[0], 16)
```
dasm, line 64: Matthew, you're abusing underscore notations everywhere.

If you don't want to pass this variable, maybe you could do something similar to this:
```
range_dict = {
  'min': int(list0[0], 16),
  'max': int(list0[1], 16),
  'num_records': ((int(list0[1], 16]) - int(list0[0], 16) / self._register_stride) + 1,
}
```

python/opae.admin/opae/admin/tools/regmap_debugfs.py
```diff
@@ -0,0 +1,149 @@
...
+            raise Exception("could not find range entry")
+
+        _registers = os.path.join(regmapdir, 'registers')
+        if not os.path.isfile(_registers):
+            raise Exception("could not find registers entry")
+
+        self._registers = _registers
+
+        with open(self._registers, 'r') as _file:
+            line0 = _file.readline()
```
dasm, line 52: Is there better description of what this line is doing? `line0` and `line1` are not very straightforward, and do not explain the purpose of this.

python/opae.admin/opae/admin/tools/regmap_debugfs.py
fpgamatt: I like the check_path approach you suggested.  I will update.

python/opae.admin/opae/admin/tools/regmap_debugfs.py
fpgamatt: Ok, I can separate out retrieving information from the __init__ method.

python/opae.admin/opae/admin/tools/regmap_debugfs.py
fpgamatt: The idea here is to read two lines from the debugfs and compare the addresses to get the stride between registers.  In some cases the addresses are word address, and in other cases the addresses are byte addresses.  I'll add comments and make better names.

python/opae.admin/opae/admin/tools/regmap_debugfs.py
```diff
@@ -0,0 +1,166 @@
...
+            if range_dict['min'] <= addr <= range_dict['max']:
+                found = True
+                offset = (addr - range_dict['min'])
+                num_records = num_records + offset / self.register_stride
+                break
+            else:
+                num_records = num_records + range_dict['num_records']
+
+        if not found:
+            raise Exception("bad address")
```
dasm, line 84: @fpgamatt as I mentioned in a previous review. Usage of `raise Exception` for end user is discouraged. User can be scared of seeing traceback, similar to this:

With raise
```
import argparse
import os
import sys


def raise_func():
    raise Exception("some exception")

def main():
    parser = argparse.ArgumentParser()

    args = parser.parse_args()
    raise_func()

if __name__ == '__main__':
    main()
```
Returns
```
uzyszkodnik@dell:/mnt/c/Users/Home/git/tools $ python3 check_dir.py
  File "check_dir.py", line 16, in <module>
    main()
  File "check_dir.py", line 13, in main
    raise_func()
  File "check_dir.py", line 7, in raise_func
    raise Exception("some exception")
Exception: some exception
```

With `sys.exit`
```
uzyszkodnik@dell:/mnt/c/Users/Home/git/tools $ cat check_dir.py
import argparse
import os
import sys


def raise_func():
    sys.exit("some exception")

def main():
    parser = argparse.ArgumentParser()

    args = parser.parse_args()
    raise_func()

if __name__ == '__main__':
    main()
```
Returns:
```
uzyszkodnik@dell:/mnt/c/Users/Home/git/tools $ python3 check_dir.py
some exception
uzyszkodnik@dell:/mnt/c/Users/Home/git/tools $ echo $?
1
```

python/opae.admin/opae/admin/tools/regmap_debugfs.py
```diff
@@ -0,0 +1,166 @@
...
+        if raddr != addr:
+            msg = "bad addr read back 0x{:x} != 0x{:x}".format(raddr, addr)
+            raise Exception(msg)
+
+        return int(list0[1], 16)
+
+    def reg_write(self, addr, val):
+        """reg_write writes val to the addr in the regmap"""
+
+        found = False
```
dasm, line 102: This one can be shortened to something similar to:
```
for range_dict in self.ranges:
    if range_dict['min'] <= addr <= range_dict['max']:
        break
else:
    raise Exception("bad address")
```
This is looping over all addresses. If none of addresses causes break of loop, it will execute `else` clause.

python/opae.admin/opae/admin/tools/regmap_debugfs.py
```diff
@@ -0,0 +1,166 @@
...
+        num_records = 0
+        for range_dict in self.ranges:
+            if range_dict['min'] <= addr <= range_dict['max']:
+                found = True
+                offset = (addr - range_dict['min'])
+                num_records = num_records + offset / self.register_stride
+                break
+            else:
+                num_records = num_records + range_dict['num_records']
```
dasm, line 81: Python uses `+=` notation, which translates:
```
num_records = num_records + range_dict['num_records']
```
to
```
num_records += range_dict['num_records']
```

python/opae.admin/opae/admin/tools/regmap_debugfs.py
```diff
@@ -0,0 +1,166 @@
...
+                break
+
+        if not found:
+            raise Exception("bad address")
+
+        with open(self.registers_path, 'rb+') as _file:
+            _file.write("{:x} {:x}".format(addr, val).encode())
+
+
+def is_regmap_dir(regmapdir):
```
dasm, line 115: 👍 looks nice!

python/opae.admin/setup.py
```diff
@@ -39,7 +39,8 @@
             'fpgaotsu = opae.admin.tools.fpgaotsu:main',
             'fpgaport = opae.admin.tools.fpgaport:main',
             'bitstreaminfo = opae.admin.tools.bitstream_info:main',
-            'opaevfio = opae.admin.tools.opaevfio:main'
+            'opaevfio = opae.admin.tools.opaevfio:main',
+            'regmap-debugfs = opae.admin.tools.regmap_debugfs:main'
```
dasm, line 43: You could add comma at the end of this line. Next time, if any addition will need to be done to this file, it will be only marked as "added" instead of "removed and added".

dasm (commented):
Just couple more minor things.
I would suggest to at least reconsider `raise` vs `sys.exit`. However, if you're fine with that, we can merge it.

python/opae.admin/opae/admin/tools/regmap_debugfs.py
fpgamatt: got it.

python/opae.admin/opae/admin/tools/regmap_debugfs.py
fpgamatt: This looks cool, but I'll use sys.exit.

python/opae.admin/setup.py
fpgamatt: If it is okay to end with a comma, I'll change it, because it is better.

python/opae.admin/opae/admin/tools/regmap_debugfs.py
```diff
@@ -0,0 +1,163 @@
...
+            sys.exit("bad write address")
+
+        with open(self.registers_path, 'rb+') as _file:
+            _file.write("{:x} {:x}".format(addr, val).encode())
+
+
+def is_regmap_dir(regmapdir):
+    """check for a valid regmap directory in debugfs"""
+    if not os.path.isdir(regmapdir):
+        sys.exit(f"{regmapdir} is not a valid directory")
```
dasm, line 117: Oh no. You should leave here `raise ArgumentTypeError` because it is the feature of `argparse`. That's the only place where you should have it. Sorry for not being clear.

python/opae.admin/opae/admin/tools/regmap_debugfs.py
```diff
@@ -0,0 +1,163 @@
...
+                }
+                self.ranges.append(range_dict)
+
+    def reg_read(self, addr):
+        """reg_read returns the value read from the given
+           address in the regmap"""
+
+        found = False
+        num_records = 0
+        for range_dict in self.ranges:
```
dasm, line 75: You could use here too, the same concept like in the other place.
```
for range_dict in self.ranges:
   (code)
else:
    sys.exit("bad read address")
```

fpgamatt: I was thinking since I already had an else: with the if, that would be a problem.  I will change to be consistent with the write method.

python/opae.admin/opae/admin/tools/regmap_debugfs.py
fpgamatt: No problem.  I will raise the ArugumentTypeError.

dasm (approved):
Thanks @fpgamatt for your work.

PR #1734, opened 2020-10-29
Fix rtl_src_config Quartus tag for SignalTap files

michael-adler:


michael-adler:
The tag for STP has always been wrong and wasn't noticed.

anandaravuri (approved):

r-rojo (approved):

PR #1749, opened 2020-11-05
build: defer spdlog and CLI11

r-rojo:
Use DEFER option in opae_external_project_add for both spdlog and CLI11
external projects. Add these two externals as dependencies to afu-test
target.


asgardkm (approved):
https://github.com/OPAE/opae-sdk/pull/1749/checks?check_run_id=1360959342#step:6:207 - nice!

I see that there are no explicit build steps for spdlog and CLI11 in the log above. However, comparing to old logs, there's no diff in build behavior here (looking at 2 day old cmake-based external "build" in the configuration step here - https://github.com/OPAE/opae-sdk/runs/1355803971?check_suite_focus=true#step:4:155

PR #1753, opened 2020-11-07
Split tarball generation for rpm generation

trixirt:
The offical fedora rpm depends on the url source tarball.
Split tarball generation out so it is easier to generate.

Generalize the naming of the tarball to follow the convention
of <package>-<version>.tar.gz.

Fix opae.spec to use the new source loction opae-<version>/

Generalize the tarball excludes all hidden files


asgardkm:
Thanks for the PR! Our current active development is occurring on the `master` branch. Our release branches are snapshots of `master` which are currently being used to facilitate version updates. They are currently not used as development branch targets.

May you please open a similar PR against `master`? Looping in @r-rojo , @tswhison to confirm.

Also, may you please update files in the `opae-libs` directory directly in it's repo here? https://github.com/OPAE/opae-libs
Once changes have been merged into master, a CI action will automatically open up a PR to pull in the changes as a subtree into `opae-libs` (example: https://github.com/OPAE/opae-sdk/pull/1751)

trixirt:
These changes have made it into the release/2.0.0-2 branch, so this pr is no longer needed

PR #1759, opened 2020-11-11
Add -S/--SHA256 option and make SHA384 default

asgibbons:


python/pacsign/pacsign/pacsign.py
```diff
@@ -200,8 +206,8 @@ def main():
             False, "Invalid key manager module %s" % args.HSM_manager
         )
 
-    if args.cert_type == "RK_384":
-        common_util.assert_in_error(False, "384-bit keys not supported")
+    #if args.cert_type == "RK_384":
+    #    common_util.assert_in_error(False, "384-bit keys not supported")
```
fpgamatt, line 210: I would just go ahead and delete these two lines.

asgibbons: Done.

dmlomartire:
I see the second user image (fim2) was added as a new image type.  What about factory image?  That was also a new image type that they want to remotely update.  Will that be handled via a separate pull request?

fpgamatt:
I don't think Scott was aware of a second factory image at the time of this PR.

fpgamatt (approved):

dmlomartire:
> I don't think Scott was aware of a second factory image at the time of this PR.

Actually, the request is the initial factory image (not a second).  Right now, there is no way to remote update the factory image (BBS).  That was a new request for Arrow Creek.

PR #1770, opened 2020-12-04
update fedora spec file and tarball generation script

anandaravuri:
Tom PR changes "Split tarball generation for rpm generation #1753" in incorporated
 https://github.com/OPAE/opae-sdk/pull/1753

The offical fedora rpm depends on the url source tarball.
Split tarball generation out so it is easier to generate.

Generalize the naming of the tarball to follow the convention
of -.tar.gz.

Fix opae.spec to use the new source loction opae-/

Generalize the tarball excludes all hidden files


michelleyho (approved):

packaging/opae/rpm/create
```diff
@@ -31,19 +31,36 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
+           'pybind11-devel' \
+           'python3-pip' \
+           'rpm-build' \
+           'python3-virtualenv' \
+           'python3-pip' \
+           'systemd' \
+           'systemd-rpm-macros' \
            'rpmdevtools'
 do
     yum list installed "${pkg}" >/dev/null 2>&1
```
r-rojo, line 66: use dnf instead of yum

anandaravuri: fixed

r-rojo (dismissed):

packaging/opae/rpm/create
```diff
@@ -66,56 +83,29 @@ do
...
-  --exclude=opae-libs/include/opae/cxx/.clang-format \
-  --exclude=opae-libs/libopaecxx/.clang-format \
-  --exclude=opae-libs/pyopae/.clang-format \
-  --exclude=opae-libs/.clang-format \
-  --exclude=opae-libs/.clang-format \
-  -z -c -f opae.tar.gz opae-sdk
-
-mv opae.tar.gz "${RPMBUILD_DIR}/SOURCES"
+
+${SDK_DIR}/scripts/create-tarball.sh opae-${version}
```
tswhison, line 97: Why call out to a separate script for this?

anandaravuri: create-tarball.sh can be used in different scripts in feature.

tswhison: The idea behind the build structure in the /packaging directory is that it would contain everything needed to make the build happen.  ie, to build the opae RPM, everything needed should be located in packaging/opae/rpm.  You're assuming that both the RPM .spec file and this external script will be in other locations.  We should move them into packaging/opae/rpm.

r-rojo (approved):

opae.spec
```diff
@@ -1,15 +1,17 @@
...
+Release:        2%{?dist}
 License:        BSD
 ExclusiveArch:  x86_64
 
 Group:          Development/Libraries
 Vendor:         Intel Corporation
 Requires:       uuid, json-c, python3
 URL:            https://github.com/OPAE/%{name}-sdk
-Source0:        https://github.com/OPAE/opae-sdk/releases/download/%{version}-1/%{name}.tar.gz
+Source0:        https://github.com/OPAE/opae-sdk/releases/download/%{version}-1/%{name}-%{version}.tar.gz
```
r-rojo, line 12: I think the '1' in %{version}-1 should be replaced by a variable.
Perhaps using %{release} or %{Release} could work since the release is set in line 4.

opae.spec
```diff
@@ -166,15 +142,15 @@ DESTDIR=%{buildroot}  cmake -DCOMPONENT=toolfpgad_api -P cmake_install.cmake
 DESTDIR=%{buildroot}  cmake -DCOMPONENT=toolfpgad_vc -P cmake_install.cmake
 
 prev=$PWD
-pushd %{_topdir}/BUILD/opae/python/opae.admin/
+pushd %{_topdir}/BUILD/%{name}-%{version}/python/opae.admin/
 %{__python3} setup.py install --single-version-externally-managed  --root=%{buildroot} --record=$prev/INSTALLED_OPAE_ADMIN_FILES
 popd
```
tswhison, line 148: It appears that the packaging directive for opae.io has been removed.  Could you please add it back?

anandaravuri: OPAE 2.0.0-2 doesn't support  opae.io ,  OPAE.IO  changes are not in OPAE2 .0.0

tswhison: Ah yes, not there for 2.0.  nvm

tswhison (dismissed):

gwencasey96:
Hi Ananda, Tim and Rodrigo,

If it's alright with you I'm reviewing 1770 in this email.

We were able to use the clean script to build successfully on Fedora 32.
However we did have a couple of problems when we tried to use the script on
Fedora rawhide.

ex/
There is failure in the build log
tools/extra/fpgadiag/cmakeFiles/eth-group.dir/eth-group.cpp.o"

Which looks like a new compiler warning is causing an error.

Is it necessary to resolve these problems for this pr or is this something
for later ?

Please let me know if you have any questions about testing this on rawhide.

Thank you
Gwen

packaging/opae/rpm/create
```diff
@@ -66,56 +83,30 @@ do
...
-  --exclude=opae-libs/pyopae/.clang-format \
-  --exclude=opae-libs/.clang-format \
-  --exclude=opae-libs/.clang-format \
-  -z -c -f opae.tar.gz opae-sdk
-
-mv opae.tar.gz "${RPMBUILD_DIR}/SOURCES"
+rm -rf _build
+mkdir _build
+cd _build
+cmake .. -DOPAE_BUILD_LEGACY=ON  -DOPAE_BUILD_TESTS=ON
```
r-rojo, line 92: This command is erroring out because line 88 changes directory to the parent of the codebase root directory (${SDK_DIR}).
A better way would be to not change directories and use source/build paths when calling cmake.

```bash
CMAKE_BUILD_DIR="${SDK_DIR}/_build"
rm -rf ${CMAKE_BUILD_DIR}
mkdir ${CMAKE_BUILD_DIR}
cmake -S ${SDK_DIR} -B ${CMAKE_BUILD_DIR
```

packaging/opae/rpm/create
```diff
@@ -66,56 +83,30 @@ do
...
 
 cd "${SCRIPT_DIR}"
 cp ./opae.spec "${RPMBUILD_DIR}/SPECS"
 
 ${SDK_DIR}/packaging/changelog.py rpm \
     --git-dir="${SDK_DIR}/.git" \
     --project="opae" \
-    --project-version="${PROJECT_VERSION}" \
+    --project-version="${version}" \
     >> "${RPMBUILD_DIR}/SPECS/opae.spec"
```
r-rojo, line 110: This appends the latest changelog commits to the the end of the opae.spec copy which violates the requirement that the changelog be in reverse descending order.
We either need to insert the changelog commits in opae.spec or use the output of the this script to insert it in the appropriate location. I would recommend we do the former (committing it in opae.spec) instead of calling it in create. Doing it this way also allows for the creation of the RPM(s) depend only on the tarball and the .spec file itself.
It would looks something like the following (note, I had to manually set the release in this text as this changelog script only uses the version):

```
* Mon Dec 14 2020 The OPAE Dev Team <opae@lists.01.org> - 2.0.0-2
- fix code review comments
- fix  codeing style
- add create tarball script
- update fedora spec file and tarball generation script
```

opae.spec
```diff
@@ -58,43 +59,18 @@ OPAE headers, tools, sample source, and documentation
...
-         opae-c++-utils\
-         opae-c++-nlb\
-         nlb0\
-         nlb3\
-         nlb7\
-         mmlink\
-         fpgad\
-         fpgad-api\
-         fpgad-vc\
+%cmake .. -DCMAKE_INSTALL_PREFIX=/usr  -DOPAE_PRESERVE_REPOS=ON -DOPAE_BUILD_LEGACY=ON -DOPAE_BUILD_SAMPLES=ON -B $PWD
```
r-rojo, line 69: We need to specify python version as 3.9 as it defaults to 2.7.
```
-DOPAE_PYTHON_VERSION=3.9
```

r-rojo (changes requested):

packaging/opae/rpm/create
tswhison: The idea was that there should be no changelog stored in opae.spec (the last thing in the file would be the %changelog line).  This way, we don't need to maintain super long changelogs, and we're not shipping changelogs for releases where they don't apply.

r-rojo:
@gwencasey96,
I tried to reproduce the build errors using rawhide container image (fedroa34) and did not run into the build error you mention with eth_group.cpp. I did run into other issues using the create script and a few more with the opae.spec file which I've addressed those in this PR. 
Also, the spec file uses the `%make_build` macro which results in calling make with multiple jobs and fails for me in my container environment (while compiling a different file hssi.cpp).
Can you provide the details on the failure with eth_group.cpp? If the failure happens but the root cause isn't obvious, try changing the opae.spec file to use make with a single job.
```
/usr/bin/make -O -j1 V=1 VERBOSE=1
```

opae.spec
anandaravuri: by default rpm creation  process uses python version 3.9 , no need to set python version in cmake

packaging/opae/rpm/create
anandaravuri: fixed

opae.spec

packaging/opae/rpm/create
anandaravuri: fixed

opae.spec
r-rojo: This is to tell OPAE build system which Python to look for. This default used to be 2.7 but was changed to 3.6 before it was changed to 3

opae.spec
```diff
@@ -279,7 +256,13 @@ popd
 
 
 %changelog
-* Thu Sep 17 2020  Ananda Ravuri <ananda.ravuri@intel.com> 2.0.0-1
+* Mon Dec 14 2020 The OPAE Dev Team <opae@lists.01.org> - 2.0.0-2
+- fix code review comments
+- fix  codeing style
```
r-rojo, line 261: s/codeing/coding/

packaging/opae/rpm/create
r-rojo: it's not fixed because this script is still appending changelog to the opae.spec copy in ${RPMBUILD_DIR}/SPECS.

anandaravuri: ${RPMBUILD_DIR}/SPECS   add change log specific to PR

%changelog-> Release info, highlights of this release

opae.spec
```diff
@@ -166,15 +142,15 @@ DESTDIR=%{buildroot}  cmake -DCOMPONENT=toolfpgad_api -P cmake_install.cmake
 DESTDIR=%{buildroot}  cmake -DCOMPONENT=toolfpgad_vc -P cmake_install.cmake
 
 prev=$PWD
-pushd %{_topdir}/BUILD/opae/python/opae.admin/
+pushd %{_topdir}/BUILD/%{name}-%{version}/python/opae.admin/
 %{__python3} setup.py install --single-version-externally-managed  --root=%{buildroot} --record=$prev/INSTALLED_OPAE_ADMIN_FILES
```
r-rojo, line 146: --record is used to record the files installed but the file (INSTALLED_OPAE_ADMIN_FILES) isn't referenced

opae.spec
```diff
@@ -166,15 +142,15 @@ DESTDIR=%{buildroot}  cmake -DCOMPONENT=toolfpgad_api -P cmake_install.cmake
...
 popd
 
-pushd %{_topdir}/BUILD/opae/python/pacsign
+pushd %{_topdir}/BUILD/%{name}-%{version}/python/pacsign
 %{__python3} setup.py install --single-version-externally-managed --root=%{buildroot} --record=$prev/INSTALLED_PACSIGN_FILES
```
r-rojo, line 150: --record is used to record the files installed but the file (INSTALLED_PACSIGN_FILES) isn't referenced.

opae.spec
```diff
@@ -300,4 +281,4 @@ popd
 - Various compiler warning fixes
 - Various memory leak fixes
 - Various Static code scan bug fixes
-- Added new FPGA MMIO API to write 512 bits
+- Added new FPGA MMIO API to write 512 bits
```
r-rojo, line 284: missing EOL char

r-rojo (commented):
For the Python files, we should either do one of the following choices:
* remove --record in the setup call and be a little more explicit in the glob pattern for the Python files installed.
```
%files devel
#...
/usr/lib/python3*/site-packages/opae
/usr/lib/python3*/site-packages/pacsign
```

* concatenate the two files (INSTALLED_OPAE_ADMIN_FILES, INSTALLED_PACSIGN_FILES) into a file that gets referenced in the %files devel section, like the following:

```
cat $prev/INSTALLED_OPAE_ADMIN_FILES >$prev/DEVEL_PYTHON_INSTALLED_FILES
cat $prev/INSTALLED_PACSIGN_FILES >>$prev/DEVEL_PYTHON_INSTALLED_FILES

%files devel -f DEVEL_PYTHON_INSTALLED_FILES
```

r-rojo (dismissed):

opae.spec
```diff
@@ -166,15 +142,15 @@ DESTDIR=%{buildroot}  cmake -DCOMPONENT=toolfpgad_api -P cmake_install.cmake
 DESTDIR=%{buildroot}  cmake -DCOMPONENT=toolfpgad_vc -P cmake_install.cmake
 
 prev=$PWD
-pushd %{_topdir}/BUILD/opae/python/opae.admin/
+pushd %{_topdir}/BUILD/%{name}-%{version}/python/opae.admin/
 %{__python3} setup.py install --single-version-externally-managed  --root=%{buildroot} --record=$prev/INSTALLED_OPAE_ADMIN_FILES
 popd
 
-pushd %{_topdir}/BUILD/opae/python/pacsign
+pushd %{_topdir}/BUILD/%{name}-%{version}/python/pacsign
```
r-rojo, line 149: still missing this macro:
```
%{?python_disable_dependency_generator}
```

anandaravuri: fixed

packaging/opae/rpm/create
trixirt: a change log is required in the fedora project so when this spec file is moved there at the least the release builds will need to be tracked

opae.spec
trixirt: Could the cmake infra auto detect the python version and bail when it is not supported ?  I am thinking of the use case where 3.8 would also work and the user would need to know to upgrade to 3.9 and to use 3.9 in the configure step

r-rojo: By adding `-DOPAE_PYTHON_VERSION=3.9` we are essentially telling cmake to bail if the Python that is found is not 3.9.
That is why we should change it from 2.7 to 3.9. We could also omit the version from the call to `find_package` but I'd rather not and be explicit about the required Python version.

packaging/opae/rpm/create
anandaravuri: fixed
Remove  ${RPMBUILD_DIR}/SPECS  from  RPM create file.
change log hard coded in opae.spec file

gwencasey96 (commented):
The build was successful on Fedora rawhide as of yesterday and successful on Fedora 33 as of today-but you may have one extra rpmbuild directory you do not need with the tarball that create built

opae-libs/plugins/xfpga/common_int.h
```diff
@@ -73,12 +73,12 @@
 #define MAP_1G_HUGEPAGE (0x1e << MAP_HUGE_SHIFT) /* 2 ^ 0x1e = 1G */
```
tswhison, line 73: This file is a part of opae-libs, so it needs to be submitted to that repo.

trixirt:
issues above look ok.
I ran into a what is likely a first time install issue.
# rpm -i packaging/opae/rpm/opae-2.0.0-2.fc33.x86_64.rpm 
Failed to preset unit: Unit file fpgad.service does not exist.

running rpmlint on opae and opae-devel show known problems.
A new problem
opae-devel.x86_64: E: non-devel-file-in-devel-package /usr/lib64/libfpgad-api.so.2.0.0
likely means this should either move to the opae package or a non system dir like /usr/lib64/opae/

tswhison:
Hi Tom, regarding the placement of libfpgad-api.so.2.0.0, this library is the glue between the fpgad daemon and the plugins that it loads.  The remainder of the fpgad files are also packaged in opae-devel.  Should we place this file elsewhere when all the other files are in opae-devel?  Just trying to understand the best approach.

trixirt:
i think rpmlint is getting hung up with all the *.so going to -devel.  I think libfpgad-api.so need to be handled like the other versioned *.so's .. I am thinking of something like this

diff --git a/opae.spec b/opae.spec
index 10bf783c..6a18fb68 100644
--- a/opae.spec
+++ b/opae.spec
@@ -183,7 +183,8 @@ popd
 %{_libdir}/libopae-c++-utils.so.2
 %{_libdir}/libopae-c++-nlb.so.%{version}
 %{_libdir}/libopae-c++-nlb.so.2
-
+%{_libdir}/libfpgad-api.so.%{version}
+%{_libdir}/libfpgad-api.so.2
 
 %files devel
 %dir %{_includedir}/opae
@@ -203,6 +204,7 @@ popd
 %{_libdir}/libopae-c++-utils.so
 %{_libdir}/libopae-c.so
 %{_libdir}/libbitstream.so
+%{_libdir}/libfpgad-api.so
 %{_libdir}/opae/libxfpga.so*
 %{_libdir}/opae/libmodbmc.so*
 %{_bindir}/bist_app*
@@ -247,7 +249,6 @@ popd
 %{_bindir}/fpgad
 /etc/opae/fpgad.cfg
 /etc/sysconfig/fpgad.conf
-%{_libdir}/libfpgad-api.so*
 %{_libdir}/opae/libfpgad-vc.so*
 /usr/lib/systemd/system/fpgad.service
 %{_bindir}/eth_group_mdev.sh

tswhison (dismissed):

opae-libs/plugins/xfpga/common_int.h
tswhison: @anandaravuri, please be sure this file is removed from the pull request and submitted to opae-libs instead.

PR #1771, opened 2020-12-04
vfio_plugin: support vf_token

r-rojo:
Opening a VF bound to vfio-pci requires a secret token (in uuid) format
shared with the PF. This change adds support for this by detecting if
the device being opened is has a symlink to a PF (phyfn).
If so, it will open the PF with a random UUID and use the same UUID with
the VF being opened.

* create a new struct, vfio_pair_t
  * contains uuid struct and two opae_vfio ptrs
* create helper functions
  * read_pci_link
    * read a symlink in a sysfs pci device path
  * open_vfio_pair
    * allocate memory for vfio_pair_t
    * check for a physfn symlink
    * if found, generate a UUID and open both the physfn and the device
      using that token.
    * if not found, open just the device without a token
  * close_vfio_pair
    * call vfio_close on the opae_vfio ptrs (if non-null)
    * free memory allocated with open_vfio_pair


anandaravuri (approved):

plugins/vfio/opae_vfio.c
```diff
@@ -306,20 +325,77 @@ vfio_handle *handle_check_and_lock(fpga_handle handle)
...
+	free(ptr);
+	*pair = NULL;
+	return 0;
+}
+
+STATIC vfio_pair_t *open_vfio_pair(const char *addr)
+{
+	char phys_device[PCIADDR_MAX];
+	char secret[GUIDSTR_MAX];
+	vfio_pair_t *pair = malloc(sizeof(vfio_pair_t));
```
tswhison, line 350: no NULL check after malloc?

plugins/vfio/opae_vfio.c
```diff
@@ -306,20 +325,77 @@ vfio_handle *handle_check_and_lock(fpga_handle handle)
...
+	memset(pair, 0, sizeof(vfio_pair_t));
+
+	pair->device = malloc(sizeof(struct opae_vfio));
```
tswhison, line 353: no NULL check after malloc?

plugins/vfio/opae_vfio.c
```diff
@@ -306,20 +325,77 @@ vfio_handle *handle_check_and_lock(fpga_handle handle)
...
+	memset(pair->device, 0, sizeof(struct opae_vfio));
+
+	if (!read_pci_link(addr, "physfn", phys_device, PCIADDR_MAX)) {
+		uuid_generate(pair->secret);
+		uuid_unparse(pair->secret, secret);
+		pair->physfn = malloc(sizeof(struct opae_vfio));
```
tswhison, line 359: no NULL check after malloc?

tswhison (dismissed):

tswhison (approved):

PR #1772, opened 2020-12-08
Add github action for  making a release

r-rojo:
This file contains jobs to run when making a tag.
When a tag is made that matches "v.*", it will trigger this workflow to
do:
* checkout the source
* set the 'release' env. variable (stripping out 'refs/heads/v' character in tag name)
* create a release using create-release action
* create the tarball with source code for the release using
  create-tarball.sh


anandaravuri (approved):

r-rojo:
Deferring this for later

PR #1775, opened 2020-12-10
Remove socket command args from fpgaconf

anandaravuri:


tests/fpgaconf/test_fpgaconf_c.cpp
```diff
@@ -230,8 +229,6 @@ TEST_P(fpgaconf_c_p, parse_args1) {
   char nine[20];
   char ten[20];
   char eleven[20];
-  char twelve[20];
-  char thirteen[20];
   char fourteen[20];
   char fifteen[20];
   char sixteen[20];
```
tswhison, line 234: Can we rename these variables to preserve the order without gaps?

tswhison (approved):

tests/fpgaconf/test_fpgaconf_c.cpp
anandaravuri: fixed

michelleyho (approved):

PR #1780, opened 2020-12-16
Update PACSign.md

asgibbons:
Added SR2 (FIM2, BBS2) for specifying secondary user image.  Added -S/--SHA256 to specify 256-bit hashes.  Default is to generate 384-bit hashes.

### DIFF
--- a/python/pacsign/PACSign.md
+++ b/python/pacsign/PACSign.md
@@ -1,11 +1,11 @@
 # PACSign #
 
 ## SYNOPSIS ##
-`python PACSign.py [-h] {FIM,SR,BBS,BMC,BMC_FW,AFU,PR,GBS} ...`
+`python PACSign.py [-h] {FIM,SR,BBS,FIM2,SR2,BBS2,BMC,BMC_FW,AFU,PR,GBS} ...`
 
 `python PACSign.py <CMD>  [-h] -t {UPDATE,CANCEL,RK_256,RK_384} -H HSM_MANAGER
                           [-C HSM_CONFIG] [-r ROOT_KEY] [-k CODE_SIGNING_KEY]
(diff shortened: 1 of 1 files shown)


asgibbons:
Addresses https://hsdes.intel.com/resource/22011830881

python/pacsign/PACSign.md
```diff
@@ -1,11 +1,11 @@
 # PACSign #
 
 ## SYNOPSIS ##
-`python PACSign.py [-h] {FIM,SR,BBS,BMC,BMC_FW,AFU,PR,GBS} ...`
+`python PACSign.py [-h] {FIM,SR,BBS,FIM2,SR2,BBS2,BMC,BMC_FW,AFU,PR,GBS} ...`
 
 `python PACSign.py <CMD>  [-h] -t {UPDATE,CANCEL,RK_256,RK_384} -H HSM_MANAGER
                           [-C HSM_CONFIG] [-r ROOT_KEY] [-k CODE_SIGNING_KEY]
-                          [-d CSK_ID] [-i INPUT_FILE] [-o OUTPUT_FILE] [-y] [-v]`
+                          [-d CSK_ID] [-S] [-i INPUT_FILE] [-o OUTPUT_FILE] [-y] [-v]`
```
fpgamatt, line 8: wouldn't -s be more consistent than -S?

asgibbons: I consider this an infrequently used option, soon to be deprecated as we're not using 256 any more, so I thought the capital would be best.  It's slightly more effort, and we may want to use lower-case for something else in the future.  I can obviously change it if you think it's important...

fpgamatt (approved):

PR #1782, opened 2020-12-18
Update reader.py

asgibbons:
Fix two cases where SHA256 was being used instead of SHA384.

### DIFF
--- a/python/pacsign/pacsign/reader.py
+++ b/python/pacsign/pacsign/reader.py
@@ -225,8 +225,8 @@ class _READER_BASE(object):
         xy_body.append_data(pub_key.data[:self.sigsize])
         xy_body.append_data(pub_key.data[-self.sigsize:])
 
-        log.info("Calculating Root Entry SHA256 for DC PR")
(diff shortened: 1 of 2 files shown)


dmlomartire (approved):

anandaravuri (dismissed):

python/pacsign/pacsign/reader.py
```diff
@@ -264,9 +264,9 @@ def make_root_entry_dc(self, pub_key):
 
         # X size
         # TODO: Should be 0x20
-        root_entry.append_dword(len(pub_key.data[:32]))
+        root_entry.append_dword(self.sigsize)
```
r-rojo, line 267: How is self.sizesize related to length of pubkey data?

r-rojo (approved):

python/pacsign/pacsign/reader.py
asgibbons: self.sigsize is the number of bytes in the hash and signature.  Either 32 or 48.  All hashes and signatures are the same length.

PR #1792, opened 2021-01-08
opae.admin: add pci_driver tool

r-rojo:
* Modify pci_node.tree function to add driver name
* Add pci_device.py tool
  This is a helper script to manage pci devices.
  To use this tool, call run:
  ```console
  pci_device <pci address> <action> [--other-endpoints]
  ```

  This initial implementation supports four actions:
  * unbind
    This action will unbind any bound drivers
  * remove
    This action will remove the pci device
  * rescan
    This action will trigger a bus rescan from the root port
  * topology
    This action will print out the pcie tree topology.
    This will print out the line showing the target device (as identified by the first position command line argument) as green
    and other endpoints as cyan.
image

  NOTE: If `--other-endpoints` argument is present, actions like remove or
  unbind will operate on the ohter endpoints in the pcie topology.

### DIFF
--- a/python/opae.admin/opae/admin/tools/pci_device.py
+++ b/python/opae.admin/opae/admin/tools/pci_device.py
@@ -1,5 +1,5 @@
 #! /usr/bin/env python
-# Copyright(c) 2020, Intel Corporation
+# Copyright(c) 2021, Intel Corporation
 #
 # Redistribution  and  use  in source  and  binary  forms,  with  or  without
 # modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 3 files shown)


python/opae.admin/opae/admin/tools/pci_device.py
```diff
@@ -0,0 +1,106 @@
+#! /usr/bin/env python
+# Copyright(c) 2020, Intel Corporation
```
tswhison, line 2: 2020-2021 ?

tswhison (dismissed):

python/opae.admin/opae/admin/tools/pci_device.py
r-rojo: 2020-2021 or 2021?

anandaravuri (approved):

python/opae.admin/opae/admin/tools/pci_device.py
fpgamatt: I think either 2020-2021 or 2021 is fine because it should be related to when the code was "published"

PR #1797, opened 2021-01-13
fpgasupdate: updates for 5.10 drivers

tswhison:
Updates the file glob for the secure device, and adds support for
the 'reading' status in the sysfs node for updates.


python/opae.admin/opae/admin/fpga.py
```diff
@@ -415,13 +415,7 @@ def secure_dev(self):
         spi = f.spi_bus
         if spi:
             ifpga_sec = spi.find_one(
```
rweight, line 417: optional nit: change the variable ifpga_sec to fpga_sec?

rweight (approved):
Looks good - one optional change commented inline.

anandaravuri (dismissed):

python/opae.admin/opae/admin/fpga.py
tswhison: fixed

python/opae.admin/opae/admin/tools/fpgasupdate.py
```diff
@@ -423,7 +423,9 @@ def update_fw(args, pac):
     retries = 0
     timeout = 1.0
     max_retries = 60 * 5
-    while status.value == 'read_file' or status.value == 'preparing':
+    # read_file is now deprecated. Leaving it in for backwards compat.
+    while status.value == 'read_file' or status.value == 'reading' \
+        or status.value == 'preparing':
```
r-rojo, line 428: I like to avoid using '\' as much as possible.
Will it work if we change the while clause to something like this?
```Python
    while status.value in ['read_file', 'reading' , 'preparing']:
```

tswhison: Thanks, I like this better.  Fixed

r-rojo (approved):

PR #1798, opened 2021-01-13
vfio-plugin: add guid to token for non-legacy cfg

r-rojo:


anandaravuri (approved):

anandaravuri (commented):
Add GUID for BAR 0, VF0,VF1 has only one bar0

r-rojo:
```c
		if (!opae_vfio_region_get(v, i, (uint8_t**)&mmio, &size)) {
			vfio_token *t = get_token(p, i, FPGA_ACCELERATOR);
			get_guid(1+(uint64_t*)mmio, t->guid);
			t->mmio_size = size;
			t->user_mmio_count = 1;
			t->user_mmio[0] = 0;
		}
```
This snippet looks for all bars (regions) in a PCIe device.
We can't assume the device MMIO layout.
The call to opae_vfio_region_get should return non-zero if if fails to get a region.
If this call is not failing when it should then we should look at fixing that instead.

PR #1805, opened 2021-01-19
vfio-plugin: refactor and fix bugs in enumeration

r-rojo:
* Refactor enumeration in vfio plugin to match PCIe attributes (BDF, socket id)
  * Add functions, pci_matches_filters and pci_matches_filter that are called first during enumeration.
    * If the filters are empty or the the pci attributes in the filters match the device, it is considered for the rest of enumeration
    * Otherwise, it is skipped
  * Move token lists from a static token list into the static list of pci_device_t structures.
  * This helps prevent reading of CSR space by filtering on specific PCIe attributes read from sysfs
    IOW, given a specific PCIe address (segment, bus, device, function), the mmio space will be read only when the PCIe address is 
    set in the filter and it matches a device.
* Fix use of open_vfio_pair so that it is called only when a VF has a PF
physfn _AND_ that device is bound to the vfio-pci driver.
* Read afu_id for BAR0 when treating a whole device as an accelerator.

### DIFF
--- a/plugins/vfio/opae_vfio.c
+++ b/plugins/vfio/opae_vfio.c
@@ -175,7 +175,7 @@ out:
 void free_token_list(vfio_token *tokens)
 {
 	while (tokens) {
-		vfio_token *t = tokens;;
+		vfio_token *t = tokens;
 		tokens = tokens->next;
 		free(t);
 	}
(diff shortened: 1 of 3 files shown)


anandaravuri (dismissed):

plugins/vfio/opae_vfio.c
```diff
@@ -173,24 +172,25 @@ STATIC int parse_pcie_info(pci_device_t *device, char *addr)
 	return res;
 }
 
+void free_token_list(vfio_token *tokens)
+{
+	while (tokens) {
+		vfio_token *t = tokens;;
```
tswhison, line 178: s/;;/;/  :)

tswhison (approved):

anandaravuri (approved):

PR #1810, opened 2021-01-21
vfio-plugin: fix token discovery

r-rojo:
* Remove `features_discover` function because it operates on all
pci_device_t structures in the static list. Now that tokens are
discovered/added to pci_device_t structures during first enumeration,
`features_discover` is no longer necessary.
* Rename `walk` to `vfio_walk`
* Replace call to `features_discover` to `vfio_walk` in `fpgaEnumerate`
implmentation.

### DIFF
--- a/plugins/vfio/opae_vfio.c
+++ b/plugins/vfio/opae_vfio.c
@@ -409,7 +409,7 @@ out_destroy:
 }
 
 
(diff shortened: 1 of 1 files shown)


fpgamatt (approved):
Successfully tested.

tswhison (approved):

PR #1813, opened 2021-01-25
hundeboll/setuptools

hundeboll-silicom:
This fixes packaging on hosts with ancient python (v3.6)


r-rojo (approved):
Thank you for this. I did a basic test on centos7.6 and was able to install both packages using python3. Both packages were also still able to be imported.

PR #1815, opened 2021-01-26
opae.io: add docs markdown

tswhison:


doc/src/fpga_tools/opae.io/opae.io.md
```diff
@@ -0,0 +1,308 @@
...
+
+The interpreter waits for a valid Python command, then attempts
+to execute the given command in the usual way. The only
+differences between the traditional Python command intepreter and
+```opae.io``` are that opae.io provides 1) the notion of an active
+PCIe device and MMIO region and 2) several built-in functions and
+objects that allow manipulating the active device and MMIO region.
+
+### BUILT-IN FUNCTIONS ###
+
```
r-rojo, line 109: It might be nice to reiterate the fact that these built in functions run in the context of the current/active device and region. Maybe something like:
```
The built-in functions below operate in the context of the current or active device and MMIO region that have been opened. Calling these functions without opening both the device and the region will result in an exception being thrown.
```

r-rojo (approved):

doc/src/fpga_tools/opae.io/opae.io.md
tswhison: ok

anandaravuri (approved):

anandaravuri (approved):

PR #1818, opened 2021-01-26
rpm: add opae.io to legacy rpm spec file

r-rojo:
* Write names of installed files into specific package name files for
  * tools-extra
  * devel
* Remove use of python/* glob in %files directive for tools-extra
  * Replace with use `%files -f tools-extra-files.txt`
* Add opae.io in test-rpms.sh file


fpgamatt (approved):
Seems straightforward to me.

tswhison (approved):
nice

michelleyho (approved):

PR #1833, opened 2021-01-30
argsfilter: refactor to accept PCIe address

tswhison:
In addition to allowing -S, -B, -D, -F, refactor
set_properties_from_args() so that it also recognizes the PCIe
address forms ssss:bb:dd.f and bb:dd.f. The address forms
are given precedence over the options.

eg fpgainfo fme -B 0xff -D 0xff -F 7 12:34.5
will set the properties as
segment  == 0
bus      == 0x12
device   == 0x34
function == 5

### DIFF
--- a/tools/argsfilter/argsfilter.c
+++ b/tools/argsfilter/argsfilter.c
@@ -119,8 +119,7 @@ STATIC bool get_pci_address(const char *addr,
 		if ((res) != FPGA_OK) {                                        \
 			optind = 1;                                            \
 			opterr = old_opterr;                                   \
-			fprintf(stderr, "Error %s: %s\n", (desc),              \
-				fpgaErrStr(res));                              \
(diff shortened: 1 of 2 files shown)


tools/argsfilter/argsfilter.c
```diff
@@ -212,6 +243,40 @@ int set_properties_from_args(fpga_properties filter, fpga_result *result,
 		}
 	}
 	*argc -= removed;
+
+	regcomp(&re, sbdf, REG_EXTENDED);
+
```
r-rojo, line 248: No need to check return value?

r-rojo (dismissed):

tools/argsfilter/argsfilter.c
tswhison: Will fix.

r-rojo (approved):

PR #1850, opened 2021-02-05
cmake: set minimum version to 3.10

r-rojo:
* set cmake minimum version to 3.10
* change afu-test to an interface (header-only) library where client libraries/applications inherit compiler/linker options when linking with afu-test.


michelleyho (approved):

tswhison (dismissed):

michelleyho (approved):

hundeboll (approved):
Rmember to change `opae-libs/CmakeLists.txt` too.

r-rojo:
@hundeboll , thanks for the reminder. I did change it in all uses in opae-libs and merged that subtree into this repository prior to changing in this repository.

PR #1865, opened 2021-02-11
 update fedora patch number in spec file

anandaravuri:
- Update fedora patch number 2 in spec and create rpm

### DIFF
--- a/opae.spec
+++ b/opae.spec
@@ -1,7 +1,9 @@
+%define         opae_release 2
+
 Summary:        Open Programmable Acceleration Engine (OPAE) SDK
 Name:           opae
 Version:        2.0.0
-Release:        2
+Release:        %{opae_release}%{?dist}
 License:        BSD
 ExclusiveArch:  x86_64
 
@@ -9,7 +11,7 @@ Group:          Development/Libraries
 Vendor:         Intel Corporation
 Requires:       uuid, json-c, python3
 URL:            https://github.com/OPAE/%{name}-sdk
-Source0:        https://github.com/OPAE/opae-sdk/releases/download/%{version}-%{release}/%{name}-%{version}-%{release}.tar.gz
+Source0:        https://github.com/OPAE/opae-sdk/releases/download/%{version}-%{release}/%{name}-%{version}-%{opae_release}.tar.gz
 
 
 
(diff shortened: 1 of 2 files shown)


opae.spec
```diff
@@ -9,7 +9,7 @@ Group:          Development/Libraries
 Vendor:         Intel Corporation
 Requires:       uuid, json-c, python3
 URL:            https://github.com/OPAE/%{name}-sdk
-Source0:        https://github.com/OPAE/opae-sdk/releases/download/%{version}-%{Release}/%{name}-%{version}.tar.gz
```
r-rojo, line 12: Why can't we use %{Release}? It's set to 2 (or dist) in line 4.

anandaravuri: https://bugzilla.redhat.com/show_bug.cgi?id=1806017 

https://github.com/trixirt/opae-rpm/blob/opae-1.4.0-5/opae.spec

https://fedoraproject.org/wiki/Packaging:DistTag


Tom was updated spec file like below,  followed his  pattern

Version:        1.4.0
Release:        5%{?dist}
Source0:        https://github.com/OPAE/opae-sdk/releases/download/%{version}-1/%{name}-%{version}-1.tar.gz

tswhison: This seems like a step backward in terms of maintainability.  I don't see why we would need to hard-code things and require that so many changes be made each time.

r-rojo: Yes, hardcoding for a release would work to but this doesn't mean we can't use a variable.
We are already setting it in line 4. Using the variable should work just as well as using the literal value except that now we only have to change it in once place if/when we have to change it.

anandaravuri: set %{Release}? to 2 in spec file

michelleyho (approved):

opae.spec
```diff
@@ -1,15 +1,15 @@
 Summary:        Open Programmable Acceleration Engine (OPAE) SDK
 Name:           opae
 Version:        2.0.0
-Release:        2%{?dist}
+Release:        2
```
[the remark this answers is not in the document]
anandaravuri: fixed  rojo comments

r-rojo (approved):

PR #1874, opened 2021-02-13
mmlink: integrate argsfilter

tswhison:
Changes --socket-id short option from -S to -s so as not to
collide with argsfilter (which uses -S for PCIe segment).


tools/extra/mmlink/main.cpp
```diff
@@ -142,26 +136,38 @@ int main( int argc, char** argv )
...
-	printf(" Bus                   : %d\n", mmlinkCmdLine.bus);
-	printf(" Device                : %d\n", mmlinkCmdLine.device);
-	printf(" Function              : %d\n", mmlinkCmdLine.function);
-	printf(" Socket-id             : %d\n", mmlinkCmdLine.socket);
-	printf(" Port                  : %d\n", mmlinkCmdLine.port);
-	printf(" IP address            : %s\n", mmlinkCmdLine.ip);
+	printf(" ------- Command line Input START ----\n");
+	printf(" Socket-id        : %d\n", mmlinkCmdLine.socket);
+	printf(" Port             : %d\n", mmlinkCmdLine.port);
+	printf(" IP address       : %s\n", mmlinkCmdLine.ip);
```
r-rojo, line 170: I wonder if we really need to print this to stdout or make these like debug messages?

r-rojo (approved):

anandaravuri (approved):

tools/extra/mmlink/main.cpp
tswhison: Should be handled in a separate request, at any rate.

PR #1877, opened 2021-02-16
cmake: use CLI11_INCLUDE_DIRS variable

r-rojo:


anandaravuri (approved):

tswhison (approved):
Should probably scrub the cmake for samples/hssi for this, too.  Want to include that here or in a separate request?

r-rojo:
> Should probably scrub the cmake for samples/hssi for this, too. Want to include that here or in a separate request?

Good call, I'll do it here.

michelleyho (approved):

PR #1887, opened 2021-02-24
rpmbuild: refactored scripts to build/test RPMs

r-rojo:
* Move docker/buildit.sh to scripts/build-rpms.sh
* Change build-rpms.sh to take in two arguments:
  * The first is mandatory and points to the cmake build tree root
  * The second is optional and points to the source tree root.
    If omitted, it defaults to the CWD
  * Change how cmake/make is invoked by using command line arguments to
    point to source tree, build tree, or both.
* Change test-rpms.sh to:
  * Take in path to directory where opae*.rpm should be found.
    If it fails to install the RPMs, the script will error out.
  * Capture exit codes of calling installed
    exectuables. Logging of unexpected exit codes will be shown at the
    end.
* Change docker files to copy build/test scripts to /scripts directory
* Change GH actions to call the containers with the proper arguments


r-rojo:
Here is a manual run the GitHub workflow "Build RPMs" that is affected by this change:
https://github.com/OPAE/opae-sdk/runs/1971407575?check_suite_focus=true

michelleyho (approved):

tswhison (approved):

PR #1909, opened 2021-03-10
ofs umd: add ofs umd framework

r-rojo:
ofs umd (user mode driver) is a framework to take driver specs in yml
format and turn it into C data structure, and function
prototypes/implementations.
This commit adds:
* ofs_parse.py script that parses an ofs umd spec in .yml format and
  generates a header file for the given driver.
* ofs_add_driver CMake macro that, given the .yml, driver name, and list
  of sources it will:
  * call ofs_parse.py to generate the .h
  * create a cmake library named after the driver name and use given
    source to compile.
* sample unit tests in tests/ofs_driver that exercise:
  * The cmake macro
  * The functions generated by ofs_parse.py

TODO:
ofs_parse.py is mostly prototype-leve code and will need to be
refactored and cleaned up.

### DIFF
--- a/cmake/modules/OFS.cmake
+++ b/cmake/modules/OFS.cmake
@@ -1,3 +1,4 @@
+#!/usr/bin/cmake -P
 ## Copyright(c) 2021, Intel Corporation
 ##
 ## Redistribution  and  use  in source  and  binary  forms,  with  or  without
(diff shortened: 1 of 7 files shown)


cmake/modules/OFS.cmake
```diff
@@ -0,0 +1,52 @@
+## Copyright(c) 2021, Intel Corporation
```
tswhison, line 1: Needs a shebang on line 1:
```python
#!/usr/bin/cmake -P
```

tswhison (approved):

michelleyho (approved):

tswhison (approved):

PR #1912, opened 2021-03-11
tools: libboard: fix interface->bfd association check

hundeboll-silicom:
Currently, the print_eth_interface_info() function is flawed when
multiple cards of the same type is present in a machine, as it will
print interface info for all interfaces for each card.

The flaw is caused by the use of the passed interface_name, as it is a
glob pattern that matches interface names for all cards with the correct
type.

Fix the issue by using the interface name looked up in the current
iteration instead.

While at it: change the message logged on mismatch from ERR to DBG, as
it is an expected case.

Signed-off-by: Martin Hundebøll <mhu@silicom.dk>


tswhison (approved):
Thanks!

PR #1914, opened 2021-03-12
samples: hssi: add loopback control of Hardware Exerciser

fpgamatt:
Add a parameter to enable/disable loopback in the Hardware
Exerciser as opposed to serial loopback in the phy.


tswhison (approved):
lgtm

PR #1919, opened 2021-03-13
work around missing cli11 default_val (int) function

trixirt:
The function
template <typename X> Option *default_val(const X &val)
Was introduced in 1.9. Previous to that, only the
Option *default_val(std::string val)
was available.

So check the version defines and convert the int to a std::string with
std::to_string().

Fixes building on rhel 8.2 where the cli11 version is 1.8


tswhison (approved):

r-rojo (approved):

r-rojo (changes requested):
Would it  be cleaner to have a macro that did this?
```c++
#if CLI11_VERSION_MAJOR <= 1 && CLI11_VERSION_MINOR <= 8 
  #define CLI11_DEFAULT(x) std::to_string(x)
#else
  #define CLI11_DEFAULT(x) x
#endif
```
Then in the code, we could use like
```c++
app_.add_option("-t,--timeout", timeout_msec_, "test timeout (msec)")->default_val(CLI11_DEFAULT(timeout_msec_));
```

trixirt:
It is a matter of taste, if you want a macro a better would be 

#if CLI11_VERSION_MAJOR <= 1 && CLI11_VERSION_MINOR <= 8 
  #define DEFAULT_VAL(x) default_val(std::to_string(x))
#else
  #define DEFAULT_VAL(x) default_val(x)
#endif

trixirt:
Is there a good place to put the macro ?
It would not be great if the macro was re-defined in every cxx file that used it.

r-rojo:
@trixirt , I'm okay with your suggestion of using DEFAULT_VAL as the macro. It could go in afu_test.h.

r-rojo:
This was superseded by #1975

PR #1939, opened 2021-03-24
qpafilter: adding QPA filter

tswhison:


tools/extra/qpafilter/qpafilter.py
```diff
@@ -0,0 +1,146 @@
...
+                      'is lower than the platform\'s recommended '
+                      'minimum of {} \u00b0C'.format(label,
+                          value_units, self.args.min_temp))
+                min_temp_failures += 1
+
+            print('{} {} {}'.format(label, value, units))
+
+        if min_temp_failures == len(items):
+            print('WARNING: All threshold values are below '
+                  'the recommended minimum ({} \u00b0C).\n'
```
michelleyho, line 74: instead of hardcoding the units in the message, maybe you can use the constant already defined `DEGREES_C`?

tools/extra/qpafilter/qpafilter.py
```diff
@@ -0,0 +1,146 @@
...
+
+    if args.file is None:
+        print('Error: file is a required argument\n')
+        parser.print_help(sys.stderr)
+        sys.exit(1)
+
+    data = read_qpa(args.file)
+    for category, key_vals in data.items():
+        try:
+            verifier = get_verifier(category)(args)
```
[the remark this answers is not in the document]
michelleyho: ah nvm... Rodrigo clarified for me.  very cool.

tools/extra/qpafilter/qpafilter.py
```diff
@@ -0,0 +1,146 @@
...
+
+import argparse
+import re
+import sys
+from collections import defaultdict
+
+SCRIPT_VERSION = '1.0.0'
+
+CATEGORY_PATTERN = r'\+-+\+\n;\s*(?P<category>(?:\w+\s?)+)\s*;\n(?P<values>.*?\n)\+-+\+-+\+'
+DATA_ITEM_PATTERN = r';\s*(?P<label>(?:\w+\s?)+)\s*;\s*(?P<value>.*?)\s;'
```
r-rojo, line 36: We could, in theory, comprehend units in this regex.
Could be something like:
```Python
DATA_ITEM_PATTERN = r';\s*(?P<label>(?:\w+\s?)+)\s*;\s*(?P<value>\d+\.?\d*)\s+(?P<units>.*?)\s;'
```

tools/extra/qpafilter/qpafilter.py
r-rojo: You can also take advantage of Python's format strings 😄 
```
print('WARNING: All threshold values are below '
         f'the recommended minimum ({self.args.min_temp} \u00b0C).\n'
         'Check QPA settings.')
```

tools/extra/qpafilter/qpafilter.py
```diff
@@ -0,0 +1,146 @@
...
+        min_temp_failures = 0
+        for label, value_units in items:
+            value, units = value_units.split(' ')
+            value = float(value)
+
+            if not self.verify_units(units):
+                print('Error: units not {}'.format(DEGREES_C))
+                return False
+            if not self.verify_min_temp(value):
+                print('INFO: QPA threshold value for sensor {} : {}\n'
```
r-rojo, line 64: Do we want to use logging module?

tools/extra/qpafilter/qpafilter.py
tswhison: fixed

tools/extra/qpafilter/qpafilter.py
tswhison: fixed

tools/extra/qpafilter/qpafilter.py
tswhison: implemented

michelleyho (dismissed):

tools/extra/qpafilter/qpafilter.py
```diff
@@ -97,10 +97,23 @@ def __init__(self, args):
...
+            if 'override' in i:
+                override = float(i['override'])
+                warning = (override * critical) / 100.0
+            else:
+                override = 0.0
+                warning = (critical * self.args.warn_temp) / self.args.crit_temp
+
             i['warning'] = str(warning)
 
-            print('{} w:{} c:{}'.format(i['label'], i['warning'], i['critical']))
```
michelleyho: any reason why we don't want logger here?

tswhison: Nope.  The use of print is intentional.  It reminds me that we need to figure out what needs to be printed there.

tools/extra/qpafilter/qpafilter.py
```diff
@@ -104,7 +104,8 @@ def filter(self, items):
                 warning = (override * critical) / 100.0
             else:
                 override = 0.0
-                warning = (critical * self.args.warn_temp) / self.args.crit_temp
+                warning = ((critical * self.args.virt_warn_temp) /
+                          self.args.virt_fatal_temp)
```
michelleyho, line 109: are we doing that 0.5 granularity  calculation here or somewhere else?
For example, if the temp is 80, then we pass the BMC, 160... vs. if the temp is 80.5, we pass to BMC 161?

tswhison: Not doing that yet.  I'll add it once we get an updated input file format.

tools/extra/qpafilter/qpafilter.py
```diff
@@ -0,0 +1,270 @@
...
+        sys.exit(1)
+
+    LOG.setLevel(logging.NOTSET)
+    log_fmt = '[%(levelname)-8s] %(message)s'
+    log_hndlr = logging.StreamHandler(sys.stdout)
+    log_hndlr.setFormatter(logging.Formatter(log_fmt))
+    log_hndlr.setLevel(logging.getLevelName(args.log_level.upper()))
+    LOG.addHandler(log_hndlr)
+
+    args.func(args)
```
r-rojo, line 266: If no subparser is called, args won't have a func attribute to call.

tools/extra/qpafilter/qpafilter.py
```diff
@@ -0,0 +1,270 @@
...
+                       'fatal': fatal,
+                       'units': units,
+                       'warning': 0.0 # placeholder
+                      }
+
+            if label in override_d:
+                inner_d['override'] = override_d[label]
+                LOG.info(f'Setting override for \'{label}\' to {inner_d["override"]}%')
+
+            outer_d[mc.group('category').strip()].append(inner_d)
```
r-rojo, line 168: Can we move this to between lines 152,153? Similar to how you did lines 154-156.
```Python
category = mc.group('category').strip()
# ... ~line 167
    outer_d[category].append(inner_d)
```

tools/extra/qpafilter/qpafilter.py
tswhison: It actually won't get that far, because it will fail on the args.file check above.  I do need to fix that check, though.  If the file isn't given, then args.file won't exist.

r-rojo (approved):

michelleyho (approved):

anandaravuri (approved):

PR #1953, opened 2021-03-30
remove the revision sysfs entry

Figo-zhang:
In fpga-upstream-dev branch, the revision sysfs entry has removed.

Here is the error log:

[root@localhost linux-dfl]# fpgainfo errors all
//****** FME ERRORS ******//
Object Id                        : 0xF000000
PCIe s:b:d.f                     : 0000:08:00.0
Device Id                        : 0x0B30
Socket Id                        : 0x00
Ports Num                        : 01
Bitstream Id                     : 0x2300001001030F
Bitstream Version                : 0.2.3
Pr Interface Id                  : 901dd697-ca79-4b05-b843-8138cefa2846
Boot Page                        : user
PCIe0 Errors                     : 0x0
errors.c:298:print_errors_info() **ERROR** : could not find FME error revision - skipping decode
PCIe1 Errors                     : 0x0
errors.c:298:print_errors_info() **ERROR** : could not find FME error revision - skipping decode
Catfatal Errors                  : 0x0

Signed-off-by: Tianfei zhang <tianfei.zhang@intel.com>


fpgamatt (changes requested):
The revision sysfs entry for errors is in the fpga-ofs-dev branch; so we want the user space code to be able to handle older drivers without the entry and newer drivers with the entry.  The current code does handle both cases, but one might improve of the user friendliness.

Figo-zhang:
https://github.com/OPAE/opae-sdk/pull/1980

PR #1964, opened 2021-03-31
update ofs_cpeng umd interface

r-rojo:
* Update register map
* Update api functions to use changed registers


michelleyho (approved):

fpgamatt (approved):
This is an example of what happens when SW is implemented while the HW is being (re)designed.  SW has to continually take the time to make changes to track the (re)design.

tswhison (approved):

michelleyho (approved):

PR #1971, opened 2021-04-02
qpafilter: mock up blob output

tswhison:
Add class blob_writer that implements a "with protocol".
Extend create_blob_from_qpa() to create the blob output.

Changes short option for --override-temp to -O.  Adds
option -o, --output to specify the output blob name.

### DIFF
--- a/tools/extra/qpafilter/qpafilter.py
+++ b/tools/extra/qpafilter/qpafilter.py
@@ -275,7 +275,7 @@ def read_qpa(in_file, temp_overrides):
             outer_d[category].append(inner_d)
 
     for override in override_d:
-        LOG.warning(f'Temperature override for {override} unused.')
+        LOG.warning(f'Temperature override for "{override}" unused.')
         for possible in difflib.get_close_matches(override, sensors_list):
             LOG.warning(f'Did you mean "{possible}"?')
 
--- a/tools/extra/qpafilter/qpafilter.py
+++ b/tools/extra/qpafilter/qpafilter.py
@@ -28,10 +28,11 @@
 """ Convert QPA text file to blob. """
 
 import argparse
+from collections import defaultdict
 import logging
 import re
+import struct
 import sys
-from collections import defaultdict
 import yaml
 
 SCRIPT_VERSION = '1.0.0'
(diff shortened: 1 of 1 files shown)


tools/extra/qpafilter/qpafilter.py
```diff
@@ -178,23 +210,25 @@ def filter(self, items):
...
     def __init__(self, str_to_int):
         self.str_to_int = str_to_int
-        self.int_to_str = {v: k for k,v in str_to_int.items()}
+        self.int_to_str = {v: k for k, v in str_to_int.items()}
 
     def __getitem__(self, i):
         if isinstance(i, str):
             return self.str_to_int[i]
-        elif isinstance(i, int):
+        if isinstance(i, int):
```
michelleyho, line 229: nice.. is it because the previous block would have returned  if it evaluated true already.  Hence, no need for elif here?

tswhison: Exactly.  The code flow is the same, but pylint is finicky about "unnecessary" elif's.

tools/extra/qpafilter/qpafilter.py
```diff
@@ -259,7 +301,7 @@ def create_blob_from_qpa(args):
                 sys.exit(1)
 
             for sensor in key_vals:
-                wr.write_sensor(sensor)
+                writer.write_sensor(sensor)
 
 
 def dump_blob(args):
```
michelleyho, line 307: is this the human readable equivalent of the 'blob'?

michelleyho (dismissed):

michelleyho (dismissed):

tools/extra/qpafilter/qpafilter.py
tswhison: Right, the dump_blob() fn will create human-readable output, given the binary from a prior "create".

tools/extra/qpafilter/qpafilter.py
```diff
@@ -152,57 +240,70 @@ def read_qpa(in_file, temp_overrides):
...
                        'units': units,
-                       'warning': 0.0 # placeholder
-                      }
+                       'warning': 0.0}  # placeholder
 
             if label in override_d:
-                inner_d['override'] = override_d[label]
-                LOG.info(f'Setting override for \'{label}\' to {inner_d["override"]}%')
+                override_d[label][1] = True
+                inner_d['override'] = override_d[label][0]
```
r-rojo, line 269: Another way you can track whether an override has been used is to pop it out of the container object (dict).
```Python
            if label in override_d:
                inner_d['override'] = override_d.pop(label)
```

tools/extra/qpafilter/qpafilter.py
```diff
@@ -152,57 +240,70 @@ def read_qpa(in_file, temp_overrides):
...
+                LOG.info(f'Setting override for \'{label}\' '
+                         f'to {inner_d["override"]}%')
 
             outer_d[category].append(inner_d)
+
+    for override in override_d:
+        if not override_d[override][1]:
+            LOG.warning(f'Temperature override for {override} unused.')
+
```
r-rojo, line 278: Then here, if you pop off used items from override_d
```Python
for override in override_k.keys():
    LOG.warning(f'{Temperature override for "{override}" unused.')
```

tools/extra/qpafilter/qpafilter.py
```diff
@@ -265,16 +265,14 @@ def read_qpa(in_file, temp_overrides):
...
+                inner_d['override'] = override_d.pop(label)
                 LOG.info(f'Setting override for \'{label}\' '
                          f'to {inner_d["override"]}%')
 
             outer_d[category].append(inner_d)
 
     for override in override_d:
-        if not override_d[override][1]:
-            LOG.warning(f'Temperature override for {override} unused.')
+        LOG.warning(f'Temperature override for {override} unused.')
```
r-rojo, line 278: You should probably surround the string label in quotes.

tools/extra/qpafilter/qpafilter.py
```diff
@@ -273,6 +276,8 @@ def read_qpa(in_file, temp_overrides):
 
     for override in override_d:
         LOG.warning(f'Temperature override for {override} unused.')
+        for possible in difflib.get_close_matches(override, sensors_list):
```
r-rojo, line 279: 😄

r-rojo (approved):

r-rojo (approved):

PR #1972, opened 2021-04-02
First set of unittests for qpafilter

michelleyho:
First set of unittests

### DIFF
--- a/tools/extra/qpafilter/test_qpafilter.py
+++ b/tools/extra/qpafilter/test_qpafilter.py
@@ -1,0 +1,74 @@
+import pytest
+import qpafilter
+import unittest
+import io
+import struct
+from unittest import mock
+
+def test_two_way_map():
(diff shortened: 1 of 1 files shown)


tools/extra/qpafilter/test_qpafilter.py
```diff
@@ -0,0 +1,74 @@
+import pytest
```
r-rojo, line 1: Missing header at the top, imports should be done in order.

r-rojo (dismissed):

tswhison (dismissed):
What Rodrigo said. :)

tools/extra/qpafilter/test_qpafilter.py
michelleyho: fixed!

r-rojo (approved):

PR #1973, opened 2021-04-02
qpafilter: implement qpafilter dump qpafilter.blob

tswhison:
Implement class blob_reader. blob_reader implements the iteration
protocol to decode and return each item in the given blob file,
presenting the human-readable data.

### DIFF
--- a/tools/extra/qpafilter/qpafilter.py
+++ b/tools/extra/qpafilter/qpafilter.py
@@ -51,6 +51,10 @@ DEGREES_c = '\u00b0c'
 BLOB_START_MARKER = b'\xff\xa5\x5a\xff\xff\xa5\x5a\xff\xff\xa5\x5a\xff'
 BLOB_END_MARKER = b'\xfe\xa5\x5a\xef\xfe\xa5\x5a\xef\xfe\xa5\x5a\xef'
 
+# TODO: endian-ness?
+BLOB_STRUCT_FORMAT = '<LLL'
+BLOB_STRUCT_SIZE = 12
+
 LOG = logging.getLogger()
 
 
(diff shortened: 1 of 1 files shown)


tools/extra/qpafilter/qpafilter.py
```diff
@@ -102,18 +106,75 @@ def write_sensor(self, sensor):
...
+
+    def __iter__(self):
+        return self.Iterator(self.blob,
+                             self.start_marker_index,
+                             self.end_marker_index,
+                             self.sensor_map,
+                             self.threshold_map)
+
+    @property
+    def valid(self):
```
r-rojo, line 171: Did you think if implementing __bool__?
https://docs.python.org/3/reference/datamodel.html#object.__bool__

tswhison: neat

tools/extra/qpafilter/qpafilter.py
```diff
@@ -102,18 +106,75 @@ def write_sensor(self, sensor):
...
+                     start_marker_index,
+                     end_marker_index,
+                     sensor_map,
+                     threshold_map):
+            self.blob = blob
+            self.start_marker_index = start_marker_index
+            self.first_entry_index = self.start_marker_index + BLOB_STRUCT_SIZE
+            self.end_marker_index = end_marker_index
+            self.num_entries = ((self.end_marker_index -
+                                 self.first_entry_index) / BLOB_STRUCT_SIZE)
```
r-rojo, line 146: do we also want to see if( end_marker_index-first_entry_index)%BLOB_STRUCT_SIZE is 0?

tswhison: I thought about that, but do you think it's necessary?  That would mean that some entity other than the tool created the blob, right?

tools/extra/qpafilter/qpafilter.py
```diff
@@ -102,18 +106,74 @@ def write_sensor(self, sensor):
...
+                    raw_value / 2)
+
+    def __iter__(self):
+        return self.Iterator(self.blob,
+                             self.start_marker_index,
+                             self.end_marker_index,
+                             self.sensor_map,
+                             self.threshold_map)
+
+    def __bool__(self):
```
r-rojo, line 170: 👍

tools/extra/qpafilter/qpafilter.py
r-rojo: Probably not necessary. I was thinking of cases when the blob gets corrupted but we don't have to overcomplicate any more than this. I suppose checking for the start and end markers is sufficient.

tools/extra/qpafilter/qpafilter.py
```diff
@@ -102,18 +106,74 @@ def write_sensor(self, sensor):
...
+            self.struct_iter = struct.iter_unpack(
+                BLOB_STRUCT_FORMAT,
+                self.blob[self.start_marker_index + len(BLOB_START_MARKER):])
+            self.sensor_map = sensor_map
+            self.threshold_map = threshold_map
+
+        def __next__(self):
+            if self.entry == self.num_entries:
+                raise StopIteration
+            self.entry += 1
```
r-rojo, line 157: Do we use this?

r-rojo (approved):

tools/extra/qpafilter/qpafilter.py
tswhison: Yeah, line 155.  It's used to check when to stop iterating.

r-rojo: Ah, we can also see if (sensor_id + threshold_id + raw_value) == BLOB_END_MARKER after line line 158 instead of relying on num_entries being a multiple of 12

tswhison: Hmm, BLOB_END_MARKER is 12 bytes now; but I'm not sure what it might end up being. The BMC team might want to use something of a different size.

r-rojo: I meant that implicitly by the calculation of num_entries.

tools/extra/qpafilter/qpafilter.py
```diff
@@ -102,18 +106,74 @@ def write_sensor(self, sensor):
...
+                             self.end_marker_index,
+                             self.sensor_map,
+                             self.threshold_map)
+
+    def __bool__(self):
+        """Return indication of whether the input blob contains
+           valid start and end markers."""
+        return (self.start_marker_index != -1 and
+                self.end_marker_index != -1)
+
```
michelleyho, line 175: should also verify that end_marker_index should occur after the start_marker_index?

r-rojo (approved):

michelleyho (approved):

PR #1974, opened 2021-04-06
qpafilter: add formatted output for dump

tswhison:
Adds option -o,--output that defaults to stdout.
Adds option -F,--format with values csv, json, and yaml.

Add classes sensor_formatter, which is the base formatter class;
and classes csv_sensor_formatter, json_sensor_formatter, and
yaml_sensor_formatter to specialize the formatting per the requested
format type.

### DIFF
--- a/tools/extra/qpafilter/qpafilter.py
+++ b/tools/extra/qpafilter/qpafilter.py
@@ -120,12 +120,13 @@ class blob_reader:
     """Given an input file pointer, a sensor map, and a threshold map,
        provide an iterator interface to reading the raw sensor data
        items sequentially."""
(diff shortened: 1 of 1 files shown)


tools/extra/qpafilter/qpafilter.py
```diff
@@ -456,6 +553,13 @@ def parse_args():
     dump.add_argument('file', type=argparse.FileType('rb'), nargs='?',
                       help='Input blob file')
 
+    dump.add_argument('-o', '--output', default='stdout',
+                      help='Output text file (default=stdout)')
+
```
r-rojo, line 558: I *think* we can set the type to argparse.FileType('w'), set default to sys.stdout, and set nargs='?'.

PR #1979, opened 2021-04-07
Homichel/qpafilter unittest

michelleyho:
Added a couple more testcases.

### DIFF
--- a/tools/extra/qpafilter/test_qpafilter.py
+++ b/tools/extra/qpafilter/test_qpafilter.py
@@ -108,6 +108,12 @@ def test_blob_writer():
     with open('/tmp/blob_written.bin','rb') as fIn:
         data = fIn.read()
         assert data == qpafilter.BLOB_START_MARKER + qpafilter.BLOB_END_MARKER
+
+def test_get_filter():
+    category = "Temperature and Cooling"
+    actual = qpafilter.get_filter(category)
+    assert actual == qpafilter.temp_filter
+    # TODO negative case: pass in non existing category... check for key error/exception?
        
 def test_get_verifier(): 
     category = "Temperature and Cooling"
--- a/tools/extra/qpafilter/test_qpafilter.py
+++ b/tools/extra/qpafilter/test_qpafilter.py
@@ -108,6 +108,38 @@ def test_blob_writer():
     with open('/tmp/blob_written.bin','rb') as fIn:
         data = fIn.read()
         assert data == qpafilter.BLOB_START_MARKER + qpafilter.BLOB_END_MARKER
-        
-        
+       
+def test_get_verifier(): 
+    category = "Temperature and Cooling"
+    actual = qpafilter.get_verifier(category)
+    assert actual == qpafilter.temp_verifier
+    # TODO negative case: pass in non existing category... check for key error/exception?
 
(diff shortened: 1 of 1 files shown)


tswhison (commented):
Hey @michelleyho, do you plan to address the TODO's in this PR or in another?

michelleyho:
@tswhison I was thinking another PR.  The notes are meant to be things to revisit again later, after discussing with you and Rodrigo.  For example, for the ones in (test_get_filter, test_get_verifier), I can see a possible testcase for something but I think it may fail right now, which is why I didn't want to add it yet.

tools/extra/qpafilter/test_qpafilter.py
```diff
@@ -108,6 +108,44 @@ def test_blob_writer():
...
+       
+def test_get_verifier(): 
+    category = "Temperature and Cooling"
+    actual = qpafilter.get_verifier(category)
+    assert actual == qpafilter.temp_verifier
+    # TODO negative case: pass in non existing category... check for key error/exception?
+
+def test_read_sensors():
+    mock_open = mock.mock_open(read_data='Sample sensor 1: 8')
+    with mock.patch('builtins.open', mock_open):
```
r-rojo, line 126: 😎 I like your use of patching...

tools/extra/qpafilter/test_qpafilter.py
```diff
@@ -108,6 +108,44 @@ def test_blob_writer():
...
+; Sample Sensor 3 Temp                          ; 100  °c   ;
++-----------------------------------------------+-----------+
+''' 
+   fileIn = io.StringIO(data)
+
+   valid_actual = qpafilter.read_qpa(fileIn, valid_temp_overrides)
+   assert valid_actual["Temperature and Cooling"] == [{'label': 'Sample Sensor 1 Temperature', 'fatal': '55.0', 'units': '°C', 'warning': 0.0,  'override': '50'}, {'label': 'Sample Sensor 2 Temperature', 'fatal': '54.5', 'units': '°F', 'warning': 0.0, 'override':'80'}, {'label': 'Sample Sensor 3 Temp', 'fatal': '100', 'units': '°c', 'warning': 0.0}]
+
+   # TODO:  Anyway to check this?
+   _ = qpafilter.read_qpa(fileIn, invalid_temp_overrides)
```
r-rojo, line 150: So I looked into this a bit and found this:
https://docs.pytest.org/en/stable/capture.html
https://docs.pytest.org/en/stable/logging.html

These two links talk about test fixtures (like `capsys` , `caplog`, etc.) that automagically transform your test function into one with a fixture. 😄 
All you have to do is include one of these fixtures names as the first argument to your test function and you can access it for any assertions and/or test control.

```Python
def test_read_qpa(caplog):
    # ...
    # clear caplog log records
    caplog.clear()
    _ = qpafilter.read_qpa(fileIn, invalid_temp_overrides)
    # get the log messages in caplog.records
    messages = [r.message for r in caplog.records]
    # assert we get expected log messages
    assert '50 is not a valid temperature override' in messages[0]
    assert '"Sample Sensor 13 Temp" unused' in messages[1]
```

r-rojo (approved):

PR #1986, opened 2021-04-09
qpafilter: implement header version 0

tswhison:
Update blob_writer and blob_reader -> blob_reader_v0 to
comprehend the version 0 blob header format, including
the length field and 2 CRC's.

Add function get_blob_reader which parses the blob header
to discover the version, returning the appropriate reader
object matching the blob version.

Co-authored-by: Rodrigo Rojo <rodrigo.rojo@intel.com>

### DIFF
--- a/tools/extra/qpafilter/qpafilter.py
+++ b/tools/extra/qpafilter/qpafilter.py
@@ -28,14 +28,17 @@
 """ Convert QPA text file to blob. """
 
 import argparse
-from collections import defaultdict
 import difflib
+import io
 import json
 import logging
+import os
 import re
 import struct
 import sys
 import yaml
+import zlib
+from collections import defaultdict
 
 SCRIPT_VERSION = '1.0.0'
 
@@ -49,10 +52,10 @@ TEMPERATURE_CATEGORY = 'Temperature and Cooling'
 DEGREES_C = '\u00b0C'
 DEGREES_c = '\u00b0c'
 
-BLOB_START_MARKER = b'\xff\xa5\x5a\xff\xff\xa5\x5a\xff\xff\xa5\x5a\xff'
-BLOB_END_MARKER = b'\xfe\xa5\x5a\xef\xfe\xa5\x5a\xef\xfe\xa5\x5a\xef'
+BLOB_HEADER_MAGIC = 0xa7cbef29
+BLOB_HEADER_SIZE = 20
+BLOB_HEADER_FORMAT = '<IIII'
 
-# TODO: endian-ness?
 BLOB_STRUCT_FORMAT = '<LLL'
 BLOB_STRUCT_SIZE = 12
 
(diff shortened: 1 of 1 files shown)


r-rojo (approved):

r-rojo:
👍

tools/extra/qpafilter/qpafilter.py
```diff
@@ -107,80 +112,99 @@ def write_sensor(self, sensor):
...
                                        warning))
-        self.outfile.write(struct.pack(BLOB_STRUCT_FORMAT,
+        self.payload.write(struct.pack(BLOB_STRUCT_FORMAT,
                                        sensor_id,
                                        upper_fatal_id,
                                        fatal))
 
 
-class blob_reader:
+class blob_reader_v0:
```
michelleyho, line 125: Will we  be having other blob reader versions in the same file in the future?

r-rojo: The version is for the format or spec of the binary file. If the format ever needs to change, we can use the  new version when we create the file with that changed spec. Any reader (software or hardware) can parse it according to that versioned spec.

tools/extra/qpafilter/qpafilter.py
```diff
@@ -107,80 +112,99 @@ def write_sensor(self, sensor):
...
        items sequentially."""
+
+    version = 0
+
     def __init__(self, blobfp, sensor_map, threshold_map):
-        self.blob = blobfp.read()
-        self.start_marker_index = self.blob.find(BLOB_START_MARKER)
-        self.end_marker_index = self.blob.find(BLOB_END_MARKER)
+        self.raw_hdr = blobfp.read(BLOB_HEADER_SIZE)
+        self.blob_hdr = struct.unpack('<IIIII', self.raw_hdr)
```
michelleyho, line 134: If i understood this correctly.. after this line, the self.blob_hdr will contain 5 elements in the list?  blob_hdr[0] -->header magic number, blob_hdr[1] --> payload length, blob_hdr[2] -->blob version, blob_hdr[3]-->payload crc,  and blob_hdr[4]->header crc

michelleyho (approved):

tools/extra/qpafilter/qpafilter.py
tswhison: Yes, I'm trying to establish a pattern that would allow for multiple readers to exist in the same script.  I figured that, if the writer needs to be updated (next blob version), then a new version of the writer could just replace the previous writer; but at the same time, there's no reason why the same script couldn't dump various versions of the blob.

tools/extra/qpafilter/qpafilter.py
tswhison: Correct, you've got it.

PR #1988, opened 2021-04-09
Add fmt to libraries to link with

trixirt:
On RHEL 8.2, there is this repesentative build error

/usr/bin/ld: CMakeFiles/hssi.dir/hssi.cpp.o: undefined reference to symbol '_ZTVN3fmt2v612format_errorE'
//usr/lib64/libfmt.so.6: error adding symbols: DSO missing from command line

So add the libfmt to the libraries to link with.

### DIFF
--- a/ofs/apps/cpeng/CMakeLists.txt
+++ b/ofs/apps/cpeng/CMakeLists.txt
@@ -29,5 +29,6 @@ opae_add_executable(TARGET hps
     LIBS
         ofs_cpeng
         afu-test
+	fmt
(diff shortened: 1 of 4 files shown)


r-rojo:
Thanks, Tom, for pointing this out. However, some distros may not have an spdlog that links with an external fmt library.
Because of this, I've added a few more checks before linking with fmt.
Please see #1990 for this change.

trixirt:
https://github.com/OPAE/opae-sdk/pull/1990
Is much better, let's use it!

PR #1990, opened 2021-04-10
cmake: link with fmt if spdlog is found

r-rojo:
Some distros include spdlog that uses format operations from an external
fmt library. When this is the case, finding spdlog will set
-DSPDLOG_FMT_EXTERNAL in its definitions.
This change checks for this when it finds the spdlog package that comes
as part of the distro. If it detects fmt is external, it will set
FMT_LIBRARIES to "fmt::fmt" which is the link interface used when
find_package(fmt) finds fmt on the system. When linking afu_test with
spdlog, it will check if FMT_LIBRARIES is defined and call
find_package(fmt) again from its directory so that it will import the
link target, "fmt::fmt"

### DIFF
--- a/afu-test/CMakeLists.txt
+++ b/afu-test/CMakeLists.txt
@@ -28,6 +28,13 @@ cmake_minimum_required (VERSION  3.10)
 project(afu-test)
 add_library(afu-test INTERFACE)
 
+if (fmt_LIBRARIES)
+    # if we found fmt before (from external/CMakeLists.txt)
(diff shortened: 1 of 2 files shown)


tswhison (approved):

michelleyho (dismissed):

michelleyho (approved):

trixirt (approved):
This change fixes the built issues on 8.2

r-rojo:
@trixirt , Thanks for testing this out!

PR #1993, opened 2021-04-10
Homichel/qpafilter unittest

michelleyho:
A couple more unittests, mostly for function "create_blob_from_qpa"

### DIFF
--- a/tools/extra/qpafilter/test_qpafilter.py
+++ b/tools/extra/qpafilter/test_qpafilter.py
@@ -30,9 +30,31 @@ import pytest
 import qpafilter
 import struct
 import unittest
+import tempfile
 
+from pathlib import Path
 from unittest import mock
 
+
+@pytest.fixture
+def invalid_qpa_data():
+   return {'Temperature and Cooling': [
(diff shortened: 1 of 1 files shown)


michelleyho:
@tswhison.. I'm looking at the function 'create_blob_from_qpa', and these lines specifically:
```
  if not filt.filter(key_vals):
       LOG.error(f'{category}: filtering failed.')
       sys.exit(1)
 ```
 the filter function in the class temp_filter, always return True?  Or I must not be seeing how it can return False ever.

r-rojo (approved):
Very nice! 😄

tswhison:
> @tswhison.. I'm looking at the function 'create_blob_from_qpa', and these lines specifically:
> 
> ```
>   if not filt.filter(key_vals):
>        LOG.error(f'{category}: filtering failed.')
>        sys.exit(1)
> ```
> 
> the filter function in the class temp_filter, always return True? Or I must not be seeing how it can return False ever.

Good observation.  Although the temp_filter.filter( ) function always returns true, it's possible that some other filter that might appear in the future (voltage or current) might need to return false in some case.  I was just leaving that possibility open.

tswhison (approved):

PR #2003, opened 2021-04-20
opae-sdk: fix hssi setup.py installation

anandaravuri:
- added dummy.cpp file to install python  files to /usr/lib64/python3.9/site-packages/hssi/
  Without dummy.cpp python  files install into folder  /usr/lib/python3.9/site-packages/hssi/

  -fix setupy package


tswhison:
There's got to be a way to do this without injecting such a workaround.  Maybe @r-rojo can help.

PR #2004, opened 2021-04-20
Rm extern cli11

trixirt:


michelleyho (approved):

r-rojo:
Hi @trixirt, Can you provide some details on what this change is meant to accomplish and why?
From a quick glance, it seems like you want to remove cli11 as a dynamic external.
I'm not sure we want to do this just yet as it may not be available on other/older distros we are still supporting.

trixirt:
Yes, removing the external project.
I am trying to mimimize the use of all the external projects.
In general external projects introduce overhead, license to scan, code to deliver.  When building they add risk that on some distro they will not build.  When a snapshot of the external project is delivered, there is no path to fix the project.  For example if a cve was opened against ci11, how would this be fixed ?

r-rojo:
@trixirt , thanks for your explanation. Yes, I think adding externals adds overhead but I wouldn't want to completely remove them without having something to replace them (for distros that may not have that dependency). Let me think about a solution that can satisfy both scenarios.

trixirt:
https://github.com/OPAE/opae-sdk/pull/2006 splits out the rpm part
and leaves the external cli11

trixirt:
The rpm part has be moved to https://github.com/OPAE/opae-sdk/pull/2051

PR #2006, opened 2021-04-22
rpm: require cli11-devel for rpm building

trixirt:
To build the application that use cli11 requires that the
cli11-devel rpm be installed on Fedora and RHEL.


trixirt:
The rpm only part of 
https://github.com/OPAE/opae-sdk/pull/2004

r-rojo (approved):

trixirt:
This never made it to master and is being carried in 
https://github.com/OPAE/opae-sdk/pull/2051

PR #2015, opened 2021-04-23
doc: add 1st version of hps doc

r-rojo:


ofs/apps/cpeng/readme.md
```diff
@@ -0,0 +1,88 @@
+# hps #
+
+## SYNOPSIS ##
+
+`hps OPTIONS SUBCOMMAND SUBCOMMAND_OPTIONS`
```
fpgamatt, line 5: I'm not sure a base command name of hps is the best choice.  I worry there might be some conflict with something on https://rocketboards.org/.  I also think that HPS has a particular meaning to a particular SOC implementation, and we might want to think about more generic name.

ofs/apps/cpeng/readme.md
```diff
@@ -0,0 +1,88 @@
...
+  `cpeng`
+
+    Run copy engine command. The copy engine command is used to program copy
+    engine AFU registers to copy an image file from the host into the FPGA DDR.
+    The copy engine AFU is logic that can be used to copy an HPS image into the
+    FPGA DDR. When the HPS boots, the first stage boot loader will load an image
+    from a specific offset in DDR that will be used to transition into the second
+    stage boot loader and subsequently boot into the embedded Linux that is also
+    part of this image.
+
```
fpgamatt, line 49: The copy engine is a specific hardware implementation of something that essentially allows for downloading firmware to a SOC on the board.  This flow is very similar to the Partial Reconfiguration flow and the secure manager flow.  In all cases we are downloading some binary to be consumed by some portion of the card.

Since Partial Reconfiguration and secure downloads are both handled by the fpgasupdate command, shouldn't we put this functionality there as well?  The image being download is a FIT image.  The command, fpgasupdate, would have to notice this type and branch to code that talks to the copy engine.

tswhison (approved):

PR #2021, opened 2021-05-05
Ramya/he enhancement rev2

ramyavex:
Added fpgaperf_counter library and fpgaperf class to host_exerciser source code to report similar metrics as fpgadiag did in OFS EA.

Implemented fpgaperf_counter shared library module works similar to perf tool.


ramyavex:
fixing the static analysis error

PR #2023, opened 2021-05-05
Ramya/he enhancement rev2

ramyavex:
Implemented fpgaperf_counter shared library module works similar to perf tool


ramyavex:
fix sanity check error

PR #2024, opened 2021-05-05
samples:Host Exerciser Enhancement

ramyavex:
Added fpgaperf_counter library and fpgaperf class to host exerciser source code to report similar metrics as fpgadiag does in OFS EA.

Implemented fpgaperf_counter shared library module works similar to perf tool.

It reads the FPGA performance counter values before workload and after workload, calculate delta and prints.

[fpgaperf_counter_log.txt](https://github.com/OPAE/opae-sdk/files/6429268/fpgaperf_counter_log.txt)
[he_enhancement_log.txt](https://github.com/OPAE/opae-sdk/files/6429269/he_enhancement_log.txt)


tswhison (commented):
Couple of high-level things need to be taken care of before a more in-depth review.  Please make sure to add the Intel copyright notice to the top of all files, including CMakeLists.txt.  Use existing files as examples.  Secondly, the code formatting is mostly incorrect according to our standards.  Please use checkpatch.pl from the Linux kernel tree and fix all errors that are reported.  Thirdly, since you are linking against opae-c, you can take advantage of the OPAE_INFO, OPAE_ERR, and OPAE_DBG printing macros.  No need to re-invent your own in the header.  Thanks!

ramyavex:
Updated the review comments

    Added header to CMakelist.txt
    Formatted the source using checkpatch.pl
    Used OPAE_ERR instead of PERF_ERR.

    Working on unit test cases.

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,461 @@
...
+	pmus->fme = fme;
+
+	return FPGA_OK;
+}
+
+fpga_result fpgaperfcounterstart(void)
+{
+	DIR *dir = NULL, *event_dir = NULL, *format_dir = NULL;
+	struct dirent *entry = NULL;
+	char dir_name[MAX] = "";
```
anandaravuri, line 357: Initialize like  char path[SYSFS_PATH_MAX] = { 0 };  dir_name ,event_name

ramyavex:
Added the unit test cases for fpgaperf_counter library.
Attached the output logs.

[output_for_hw_test_log.txt](https://github.com/OPAE/opae-sdk/files/6460932/output_for_hw_test_log.txt)
output_of_unit_test_fpgaperf

samples/host_exerciser/host_exerciser_cmd.h
```diff
@@ -26,12 +26,55 @@
...
+  }
+  fpga_result perfstop(void)
+  {
+	fpga_result res = FPGA_OK;
+
+	res = fpgaperfcounterstop();
+	if (res != FPGA_OK)
+		return FPGA_EXCEPTION;
+
+	return FPGA_OK;
```
tswhison, line 66:
```c
return fpgaperfcounterstop();
```

ramyavex:
Fixed the review comments

    Added mutex lock and unlock mechanism for global structure
    For print function added FILE * parameter
    Modified OPAE_ERR format
    Modified the strtol to return last byte instead of character
    Modified the return statements

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,508 @@
...
+	int is_start;
+	int num_formats;
+	int num_generic_events;
+	int num_read_events;
+	struct format_type *formats;
+	struct generic_event_type *generic_events;
+};
+
+/* Not static so other tools can access the PMU data */
+struct fpga_perf_counter *fpga_perf_pmus = NULL;
```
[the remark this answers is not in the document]
ramyavex: Added opae mutex lock and unlock mechanism for global structure

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,508 @@
...
+  
+	if (read(fpga_perf_pmus->generic_events[0].fd, rdft, sizeof(buf)) == -1) {
+		OPAE_ERR("read fpga perf counter failed");
+		return FPGA_EXCEPTION;
+	}
+
+	for (loop = 0; loop < (int)rdft->nr; loop++) {
+		for (inner_loop = 0; inner_loop < fpga_perf_pmus->num_generic_events; inner_loop++) {
+			if (rdft->values[loop].id == fpga_perf_pmus->generic_events[inner_loop].id) {
+				if (is_start == 1)
```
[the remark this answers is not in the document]
ramyavex: Modified

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,508 @@
...
+	if (ioctl(fpga_perf_pmus->generic_events[0].fd, PERF_EVENT_IOC_DISABLE, PERF_IOC_FLAG_GROUP) == -1) {
+		OPAE_ERR("PERF_EVENT_IOC_DISABLE ioctl failed: %s", strerror(errno));
+		return FPGA_EXCEPTION;
+	}
+	fpga_perf_pmus->is_start = 0;
+	/* read the performance counter for the workload*/
+	res = read_fab_counters(fpga_perf_pmus->is_start);
+	if (res != FPGA_OK)
+		return FPGA_EXCEPTION;
+	return FPGA_OK;
```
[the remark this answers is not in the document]
ramyavex: fixed

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,508 @@
...
+}
+
+void fpgaperfcounterprint(void)
```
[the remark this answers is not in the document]
ramyavex: added FILE * parameter for fpgaperfcounterprintf()

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,508 @@
...
+
+/* parse the each format and get the shift val */
+fpga_result prepare_format(DIR *dir, char *dir_name)
+{
+	int format_num = 0, result = -1;
+	struct dirent *format_entry = NULL;
+	FILE *file = NULL;
+	char sysfspath[DFL_BUFSIZ_MAX] = { 0,};
+
+	if ((dir == NULL) || (dir_name == NULL)){
```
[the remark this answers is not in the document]
ramyavex: fixed

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,508 @@
...
+					OPAE_ERR("PERF_EVENT_IOC_ID ioctl failed: %s", strerror(errno));
+					return FPGA_EXCEPTION;
+				}
+			}
+		}
+	}
+	/* reset the performance counter to 0 */
+	if (ioctl(fpga_perf_pmus->generic_events[0].fd, PERF_EVENT_IOC_RESET, PERF_IOC_FLAG_GROUP) == -1) {
+		OPAE_ERR("PERF_EVENT_IOC_RESET ioctl failed: %s", strerror(errno));
+		return FPGA_EXCEPTION;
```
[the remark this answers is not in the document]
ramyavex: fixed

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,508 @@
...
+}
+
+void fpgaperfcounterfree(void)
+{
+	if (fpga_perf_pmus->formats != NULL)
+		free(fpga_perf_pmus->formats);
+	if (fpga_perf_pmus->generic_events != NULL)
+		free(fpga_perf_pmus->generic_events);
+	if (fpga_perf_pmus != NULL)
+		free(fpga_perf_pmus);
```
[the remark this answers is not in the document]
ramyavex: fixed

samples/host_exerciser/host_exerciser_cmd.h
```diff
@@ -26,12 +26,55 @@
...
+public:
+  fpga_result perfinit(uint16_t segment, uint8_t bus, uint8_t device, uint8_t function)
+  {
+	fpga_result res = FPGA_OK;
+
+	res = fpgaperfcounterinit(segment, bus, device, function);
+	if (res != FPGA_OK)
+		return FPGA_EXCEPTION;
+
+	return FPGA_OK;
```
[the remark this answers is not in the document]
ramyavex: fixed

samples/host_exerciser/host_exerciser_cmd.h
```diff
@@ -26,12 +26,55 @@
...
+  }
+  fpga_result perfstart(void)
+  {
+	fpga_result res = FPGA_OK;
+
+	res = fpgaperfcounterstart();
+	if (res != FPGA_OK)
+		return FPGA_EXCEPTION;
+
+	return FPGA_OK;
```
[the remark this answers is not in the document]
ramyavex: fixed

samples/host_exerciser/host_exerciser_cmd.h
ramyavex: fixed

tswhison (commented):
Use of global variable fpga_perf_pmus is proving difficult with respect to locking.  Instead of making this variable a global, why not make the first parameter to each function here a pointer to a struct fpga_perf_counter?  This would eliminate the need to perform locking around the global, and is much cleaner.

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,644 @@
...
+	opae_mutex_unlock(res, &fpga_perf_lock);
+	return FPGA_EXCEPTION;
+}
+
+fpga_result fpgaPerfCounterPrint(FILE **f)
+{
+	uint64_t loop = 0;
+	int res;
+
+	if (!*f)
```
tswhison, line 595: This is still dereferencing a NULL pointer.  If someone calls..
```c
fpgaPerfCounterPrint(NULL);
```

samples/fpgaperf_counter/fpgaperf_counter.h
```diff
@@ -0,0 +1,93 @@
...
+fpga_result fpgaPerfCounterEnum(fpga_token token);
+
+/* 
+ * Enable the performance counter, read the start value and update
+ * the fpga_perf_counter start value.
+ * 
+ * @returns FPGA_OK on success. FPGA_EXCEPTION if an internal exception occurred
+ * while trying to access the handle.
+ */
+fpga_result fpgaPerfCounterStartrecord(void);
```
tswhison, line 57: fpgaPerfCounterStartRecord

samples/fpgaperf_counter/fpgaperf_counter.h
```diff
@@ -0,0 +1,93 @@
...
+
+/*
+ * Disable the performance counter, read the stop value and update
+ * the fpga_perf_counter stop value.
+ *
+ * @returns FPGA_OK on success. FPGA_EXCEPTION if an internal exception occurred
+ * while trying to access the handle.
+ */
+fpga_result fpgaPerfCounterStoprecord(void);
```
tswhison, line 66: fpgaPerfCounterStopRecord

tests/fpgaperf/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+        opae-c
+)
+
+opae_test_add(TARGET test_fpgaperf_c
+    SOURCE test_fpgaperf_counter_c.cpp
+    LIBS
+        fpgaperf-static
+)
+
+target_link_libraries(test_fpgaperf_c "-ludev")
```
tswhison, line 39: Remove this line and place udev below line 30.

samples/fpgaperf_counter/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+
+opae_add_shared_library(TARGET fpgaperf_counter
+	SOURCE fpgaperf_counter.c
+	LIBS
+		${CMAKE_THREAD_LIBS_INIT}
+		opae-c
+	COMPONENT opaesamplelib
+)
+
+target_link_libraries(fpgaperf_counter "-ludev")
```
tswhison, line 35: Remove this line and place udev below line 31.

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,639 @@
...
+#include "fpgaperf_counter.h"
+
+
+#define DFL_PERF_FME 		"/sys/bus/pci/devices/*%x*:*%x*:*%x*.*%x*/fpga_region/region*/dfl-fme.*"
+
+#define DFL_PERF_SYSFS		"/sys/bus/event_source/devices/dfl_fme"
+
+#define DFL_PERF_STR_MAX	256
+
+#define PERF_EVNET_PATTERN	"event=(0x[0-9a-fA-F]{2}),evtype=(0x[0-9a-fA-F]{2}),portid=(0x[0-9a-fA-F]{2})"
```
tswhison, line 57: PERF_EVENT_PATTERN

afu-test/afu_test.h
```diff
@@ -189,6 +189,7 @@ class afu {
     }
 
     auto tokens = fpga::token::enumerate({filter});
+    perf_token_ = tokens[0];
```
tswhison, line 192: Hi Ramya, this change is causing a segfault in test_dummy_afu.  Could you please implement a fix?  Thanks!

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,639 @@
...
+	return FPGA_EXCEPTION;
+}
+
+fpga_result fpgaPerfCounterPrint(FILE *f)
+{
+	uint64_t loop 	= 0;
+	int res		= 0;
+
+	if (!f)
+		return FPGA_EXCEPTION;
```
r-rojo, line 590: Should log a message and return FPGA_INVALID_PARAM

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,639 @@
...
+#include <linux/perf_event.h>
+#include <linux/hw_breakpoint.h>
+#include <asm/unistd.h>
+#include <opae/log.h>
+#include "opae_int.h"
+#include <opae/properties.h>
+#include <opae/utils.h>
+#include <opae/fpga.h>
+#include <regex.h>
+#include "fpgaperf_counter.h"
```
r-rojo, line 48: Should probably undefine _GNU_SOURCE after including linux headers and also reorder the includes by:
1. prototype file (fpgaperf_counter.h)
2. C system headers (stdio, stdlib, ...)
3. Other standard libraries (libudev.h, linux/perf_event.h, ...)
4. standard opae headers (<opae/...> that are part of OPAE installation
5. other project headers 

Each header group should have its headers listed in alphabetical order with a space between each group.

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,639 @@
...
+	}
+	if (ioctl(g_fpga_perf->perf_events[0].fd, PERF_EVENT_IOC_RESET,
+				PERF_IOC_FLAG_GROUP) == -1) {
+		OPAE_ERR("PERF_EVENT_IOC_RESET ioctl failed: %s", strerror(errno));
+		ret = FPGA_EXCEPTION;
+		goto out;
+	}
+out:
+	udev_unref(udev);
+	udev_device_unref(dev);
```
r-rojo, line 363: I'm not sure it it matters or not but I would release resources in the opposite order that they were acquired, especially if one resource depends on another acquired before it.

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,639 @@
...
+		return FPGA_NOT_FOUND;
+	}
+	if (pglob.gl_pathc == 1) {
+		char *ptr = strstr(pglob.gl_pathv[0], "fme");
+		if (!ptr) {
+			ret = FPGA_INVALID_PARAM;
+			goto out;
+		}
+		errno = 0;
+		fpga_id = strtoul(ptr + 3, &endptr, 16);
```
r-rojo, line 459: Is this the number at the end? I'm not sure it's in hex.

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,639 @@
...
+	uint64_t cpumask;
+	uint64_t num_format;
+	perf_format_type *format_type;
+	uint64_t num_perf_events;
+	perf_event_type *perf_events;
+} fpga_perf_counter;
+
+/* Not static so other tools can access the PMU data */
+fpga_perf_counter *g_fpga_perf = NULL;
+
```
r-rojo, line 108: If other tools need to access PMU data then maybe we should provide API functions to get them.
Also, are we assuming we can only have one of these at a time? What if I have multiple cards and I have an application what wants to monitor perf data in multiple threads?

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,639 @@
...
+	res = fpgaPropertiesGetFunction(props, function);
+	if (res != FPGA_OK) {
+		OPAE_ERR("Failed to get Function");
+		return res;
+	}
+
+	return res;
+}
+
+fpga_result fpgaPerfCounterEnum(fpga_token token)
```
r-rojo, line 416: Does this function assume that  the device this token represents is on the same bdf as the fme? or does it assume that this token is a token for an FME?

samples/fpgaperf_counter/fpgaperf_counter.c
ramyavex: fixed

afu-test/afu_test.h
ramyavex: getting a token from handler. Its fixed testing error

samples/fpgaperf_counter/fpgaperf_counter.c
ramyavex: changed as per the comment

samples/fpgaperf_counter/fpgaperf_counter.c
ramyavex: fixed

samples/fpgaperf_counter/fpgaperf_counter.c
ramyavex: changed the order of releasing resources

samples/fpgaperf_counter/fpgaperf_counter.c
ramyavex: changed to decimal value

samples/host_exerciser/host_exerciser.h
```diff
@@ -454,6 +457,12 @@ class host_exerciser : public test_afu {
     }
     return offset;
   }
+  
+  token::ptr_t get_parent_token()
+  {
+    auto tokens = token::enumerate({properties::get(handle_)});
```
tswhison, line 463: How is this getting the parent token?

samples/host_exerciser/host_exerciser_cmd.h
```diff
@@ -26,12 +26,40 @@
...
+  }
+  fpga_result perfstop(fpga_perf_counter *fpga_perf)
+  {
+      return fpgaPerfCounterStopRecord(fpga_perf);
+  }
+  fpga_result perfprint(fpga_perf_counter *fpga_perf)
+  {
+      FILE *file = stdout;
+
+      return fpgaPerfCounterPrint(file, fpga_perf);
```
tswhison, line 55:
```c
return fpgaPerfCounterPrint(stdout, fpga_perf);
```

tests/fpgaperf/fpgaperf_counter.h
```diff
@@ -0,0 +1,153 @@
+// Copyright(c) 2021, Intel Corporation
```
tswhison, line 1: This header seems to be duplicated.  Instead of doing that, we should add the necessary include directory when building the tests.

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,646 @@
...
+	do {							\
+		errno = 0;					\
+		_v = strtoul(_p + _m, NULL, _b);		\
+		if (errno) {					\
+			OPAE_MSG("error parsing int");		\
+		}						\
+	} while (0)
+
+/* mutex to protect fpga perf pmu data structures */
+pthread_mutex_t fpga_perf_lock = PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP;
```
tswhison, line 78: Since we are no longer tracking a global variable, but are instead passing an fpga_perf_counter * to the API functions, we no longer need this global mutex (fpga_perf_lock).  Instead, let's add a pthread_mutex_t to struct fpga_perf_counter and lock/unlock it as needed.

samples/fpgaperf_counter/fpgaperf_counter.h
```diff
@@ -61,6 +70,28 @@ typedef struct {
 	perf_events_type *perf_events;
 } fpga_perf_counter;
 
+/**
+ * Initialize the pthread mutex.
```
r-rojo, line 74: I think this is an implementation detail that doesn't really summarize what this function does.
IOW, it's too specific .

samples/fpgaperf_counter/fpgaperf_counter.h
```diff
@@ -61,6 +70,28 @@ typedef struct {
...
+ * @param[inout] fpga_perf  Returns the fpga_perf_counter struct
+ *
+ * @returns FPGA_OK on success. FPGA_INVALID_PARAM if any of the supplied
+ * parameters is invalid. FPGA_EXCEPTION if an internal exception occurred
+ * while trying to access the handle.
+ */
+fpga_result fpgaPerfCreateHandle(fpga_perf_counter *fpga_perf);
+
+/**
+ * Destroy the pthread mutex.
```
r-rojo, line 85: I think this is an implementation detail that doesn't really summarize what this function does.
IOW, it's too specific .

tests/fpgaperf/test_fpgaperf_counter_c.cpp
```diff
@@ -0,0 +1,192 @@
...
+			EXPECT_EQ(fpgaClose(dev_), FPGA_OK);
+			dev_ = nullptr;
+		}
+		for (auto &t : tokens_) {
+			if (t) {
+				EXPECT_EQ(fpgaDestroyToken(&t), FPGA_OK);
+				t = nullptr;
+			}
+		}
+		delete fpga_perf;
```
r-rojo, line 106: If the assertion on line 88 fails then line 89 won't execute (so we shouldn't delete)

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -0,0 +1,718 @@
...
+	struct udev_list_entry *attrs = udev_device_get_sysattr_list_entry(dev);
+	struct udev_list_entry *le = NULL;
+	fpga_perf->num_perf_events = 0;
+	
+	udev_list_entry_foreach(le, attrs) {
+		const char *attr = udev_list_entry_get_name(le);
+		if (strstr(attr, "events"))
+			fpga_perf->num_perf_events++;
+	}
+	fpga_perf->perf_events = calloc(fpga_perf->num_perf_events,
```
r-rojo, line 216: I this this could be a memory leak if we call parse_perf_event multiple times using the same fpga_perf_counter structure.

samples/fpgaperf_counter/fpgaperf_counter.h
r-rojo: Also, I'm a little confused why we have two API functions that initialize part of the data structure.

r-rojo (commented):
I think we should evaluate the lifecycle of the data structures
(how/when structures are allocated and initialized) and where the responsibility lies (user vs API library).

samples/fpgaperf_counter/fpgaperf_counter.h
r-rojo: Or maybe this really needs to be an internal function (not an API function) that is called during the initialization of the fpga_perf_counter structure.

samples/host_exerciser/host_exerciser_cmd.h
```diff
@@ -117,11 +151,41 @@ class host_exerciser_cmd : public test_command
...
+        if (res != FPGA_OK) {
+                return -1;
+        }
+
+        res = fpgaperf.perfget(token_, fpga_perf);
+        if (res != FPGA_OK) {
+                fpgaperf.perfdestroy(fpga_perf);
+                fpgaperf.destroyhandle(fpga_perf);
+                delete fpga_perf;
+                fpga_perf = nullptr;
```
r-rojo, line 177: I'm seeing a pattern here when we want to abort and have to cleanup the fpga_perf structure.
Perhaps this could be refactored/wrapped by a function or a c++ object that does lifecycle management?

r-rojo (commented):
I now understand that the class fpga_perf is a sort of wrapper around the fpga perf APIs.
This class is used without any member variables and its functions behave mostly like static functions. IOW, the functions don't operate on the calling object and only operate on the in/out parameters. Also, the way your code is structured, every time you call a function from this class, you check the return code and if it is non-zero, you have to destroy the C structure before aborting.
I think maybe another way to wrap the perf API in a C++ object that helps you out could look like the snippet below. 
 

```c++
class fpgaperf
{
public:

static std::shared_ptr<fpgaperf> get(token::ptr_t token)
{
  auto p = std::make_shared<fpgaperf>();
  if (fpgaPerfCounterGet(token->c_type(), &p->counter_)){
    // error message or something
    p.reset();
  }
  return p;
}

~fpgaperf()
{
  fpgaPerfCounterDestroy(&counter_);
  // maybe log message if result is non-zero
}

fpga_result start()
{
  return fpgaPerfCounterStartRecord(&counter_);
}
fpga_result stop()
{
  return fpgaPerfCounterStopRecord(&counter_);
}

fpga_result print()
{
  return fpgaPerfCounterPrint(stdout, &counter_);
}

private:
  fpgaperf();
  fpga_perf_counter counter_;
};
```

The nice thing about doing it this way is that the C structure is destroyed when its C++ wrapper instance goes out of scope. In this example I've made the default constructor private so the only way to get the C++ instance is by calling the static method, get.
So the calling function would look something like this:

```c++

auto perf = fpgaperf::get();
if (!perf) {
  // raise error
}

// use the perf object
perf->start();
// stuff
perf->stop();

```

_note_: if you do it this way, you might also want to make the copy constructor private...

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -450,31 +455,41 @@ fpga_result fpgaPerfCreateHandle(fpga_perf_counter *fpga_perf)
 	return FPGA_OK;
 }
 
-fpga_result fpgaPerfDestroyHandle(fpga_perf_counter *fpga_perf)
+/* Reset the magic number and destroy the mutex created */
+fpga_result fpga_perf_mutex_destroy(fpga_perf_counter *fpga_perf)
```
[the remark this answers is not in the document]
ramyavex: We can't make because, we are calling this one in unit test.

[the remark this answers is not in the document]
tswhison: You can use STATIC (in all caps).  For the test build, STATIC is compiled out; but in non-test builds STATIC will evaluate to static.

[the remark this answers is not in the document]
ramyavex: Updated with STATIC

tswhison (dismissed):
👍

samples/host_exerciser/host_exerciser_cmd.h
```diff
@@ -36,26 +36,48 @@ namespace host_exerciser {
...
+     *of fpgaperf, and as such it cannot instantiate it. Created temporary
+     *class that inherits from fpgaperf.*/
+    struct shared_enabler : public fpgaperf {};
+    auto p = std::make_shared<shared_enabler>();
+    if (fpgaPerfCounterGet(token->c_type(), p->counter_) != FPGA_OK) {
+        std::cout << "Failed to get the fpga perf counter" << std::endl;
+        p.reset();
+        return NULL;
+    }
+    return p;
```
r-rojo, line 51: I think this is overcomplicating things a bit.
The example I gave was mostly for reference and so I hadn't tested it.
If you're getting compiler errors with make_shared, then I suggest using something simpler:
```c++
  std::shared_ptr<fpgaperf> p(new fpgaperf());
  if (fpgaPerfCounterGet(token->c_type(), p->counter_) != FPGA_OK) {
    p.reset();
  }
  return p;

```

samples/host_exerciser/host_exerciser_cmd.h
```diff
@@ -36,26 +36,48 @@ namespace host_exerciser {
...
+      return fpgaPerfCounterStartRecord(counter_);
   }
-  fpga_result perfprint(fpga_perf_counter *fpga_perf)
+  fpga_result stop()
   {
-      return fpgaPerfCounterPrint(stdout, fpga_perf);
+      return fpgaPerfCounterStopRecord(counter_);
   }
-  fpga_result perfdestroy(fpga_perf_counter *fpga_perf)
+  fpga_result print()
```
r-rojo, line 71: When moving to c++, try to simplify the code.
For example, print seems like it should return a void (unless you thing there's value in capturing the error code).
If you happen to catch errors while performing some lower level operation, then you can print the message here.
Also, you might want to switch to using c++ exceptions for error notification and handling instead of error codes.

r-rojo (approved):

PR #2025, opened 2021-05-05
tools/extra/opae.io requires libedit header files

michael-adler:


michael-adler:
@tswhison and @r-rojo Note that this does not fix the missing CMake ordering dependence that is broken. In parallel builds, CMake is building opae.io before pybind11 is ready. I don't know CMake well enough to fix that. I'm having to disable parallel builds to compile successfully.

tswhison (approved):

michael-adler:
@r-rojo will you work on pybind11 dependence to fix the parallel build?

PR #2031, opened 2021-05-11
continuing with rpm building

trixirt:
The first commit addresses a lot of noise in the build log when the .git directory is not there.
The second disables pacsign building so other rpm errors can be addressed


michelleyho (approved):

trixirt:
Most of these changes are in 
https://github.com/OPAE/opae-sdk/pull/2051

PR #2050, opened 2021-05-25
tools: argsfilter: install sources to /usr/src/opae/argsfilter

hundeboll-silicom:
Building samples installed with the opae-devel package requires access to
the argsfilter header and source, so include those in said package.

Signed-off-by: Martin Hundebøll <mhu@silicom.dk>


tswhison (changes requested):
But, remember the Fedora rule not to package static libraries?

hundeboll-silicom:
> But, remember the Fedora rule not to package static libraries?

Which is why I package the source (`argsfilter.c` and `argsfilter.h`) in `/usr/src/opae/argsfilter`... That way one can build the samples with:

```
$ gcc -I /usr/src/opae/argsfilter/ \
        /usr/src/opae/argsfilter/argsfilter.c \
        /usr/src/opae/samples/hello_fpga/hello_fpga.c \
        -l opae-c -l uuid
```

tswhison (approved):
Ah yes.. Thanks!

anandaravuri (approved):

michelleyho (approved):

PR #2059, opened 2021-06-04
Mgerlach/bcce max10

fpgamatt:
A couple of small commits to support a new device id and kernel sysfs path.


michelleyho (approved):

tswhison (changes requested):
The files being changed in the opae-libs folder must be submitted as a separate pull request at github.com/OPAE/opae-libs.

fpgamatt:
It looks to like only one file is in the opae-libs dir/repo.  I will fix.

PR #2076, opened 2021-06-15
opae-libs: pull latest master into subtree

tswhison:


tswhison:
Need to fix the coding style.

PR #2082, opened 2021-06-21
fpgasupdate idle timeout increased for IOFS-Rel1

vsivar1x:
fpgasupdate idle timeout increased for IOFS-Rel1.

Signed-off-by: Sivaramakrishnan, VenkataramananX <venkataramananx.sivaramakrishnan@intel.com>


vsivar1x:
would be a valid test to run two fpgasupdate commands on a system that has more than one accelerator card, targeting the different accelerators; but fpgasupdate is behaving as expected in the single-card case.

PR #2090, opened 2021-06-25

ramyavex:
Modified the source to access the sysfs path for formats and events
using glob instead of udev library.
With udev library in Centos8 machine not able to get events and format attributes for dfl fme.  

### DIFF
--- a/samples/fpgaperf_counter/fpgaperf_counter.c
+++ b/samples/fpgaperf_counter/fpgaperf_counter.c
@@ -105,29 +105,41 @@ STATIC fpga_result fpga_perf_check_and_lock(fpga_perf_counter *fpga_perf)
 }
 
 /* parse the each format and get the shift val */
-STATIC fpga_result parse_perf_format(struct udev_device *dev, fpga_perf_counter *fpga_perf)
+STATIC fpga_result parse_perf_format(struct udev_device *dev,
+		fpga_perf_counter *fpga_perf, const char *attr)
 {
 	regex_t re;
-	regmatch_t matches[3] 	= { {0} };
-	char err[128] 		= { 0 };
-	int reg_res 		= 0;
-	uint64_t loop 		= 0;
-	uint64_t value 		= 0;
+	regmatch_t matches[3]			= { {0} };
+	char err[128] 				= { 0 };
+	int reg_res 				= 0;
+	uint64_t loop 				= 0;
+	uint64_t value 				= 0;
+	char attr_path[DFL_PERF_STR_MAX]	= { 0,};
+	char attr_value[128]			= { 0,};
+	int  gres				= 0;
+	size_t inner_loop			= 0;
+	FILE *file				= NULL;
+	glob_t pglob;
 
 	if (!dev || !fpga_perf) {
 		OPAE_ERR("Invalid input parameters");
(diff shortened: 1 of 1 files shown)


r-rojo (commented):
Seems like parse_perf_format and parse_perf_event are very similar.
We should avoid code duplication as much as possible.
What's different about them? Is it mainly the  attribute directory (glob pattern) and the regex pattern?
If that's the case, perhaps you can refactor to one function and parameterize both the glob pattern and the regex pattern.

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -104,142 +104,142 @@ STATIC fpga_result fpga_perf_check_and_lock(fpga_perf_counter *fpga_perf)
...
 		if (!fpga_perf->format_type) {
-			fpga_perf->num_format = 0;
-			OPAE_ERR("Failed to allocate Memory");
-			return FPGA_NO_MEMORY;
+			fpga_perf->format_type = calloc(fpga_perf->num_format,
+						sizeof(perf_format_type));
+			if (!fpga_perf->format_type) {
+				fpga_perf->num_format = 0;
+				OPAE_ERR("Failed to allocate Memory");
+				return FPGA_NO_MEMORY;
```
tswhison, line 151: Need a
```c
globfree(&pglob);
```
here as well?

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -104,142 +104,142 @@ STATIC fpga_result fpga_perf_check_and_lock(fpga_perf_counter *fpga_perf)
 	return FPGA_OK;
 }
 
-/* parse the each format and get the shift val */
-STATIC fpga_result parse_perf_format(struct udev_device *dev, fpga_perf_counter *fpga_perf)
+/* parse the each format and get the shift val
+ * parse the events for the particular device directory */
+STATIC fpga_result parse_perf_attributes(struct udev_device *dev,
+			fpga_perf_counter *fpga_perf, const char *attr)
```
r-rojo, line 110: Shouldn't the regex pattern be parameterized here too?

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -275,11 +275,11 @@ STATIC fpga_result fpga_perf_events(char* perf_sysfs_path, fpga_perf_counter *fp
 		PARSE_MATCH_INT(ptr, 0, fpga_perf->type, 10);
 
 	/* parse the format value */
-	ret = parse_perf_format(dev, fpga_perf);
+	ret = parse_perf_attributes(dev, fpga_perf, "format");
 	if (ret != FPGA_OK)
 		goto out;
 	/* parse the event value */
-	ret = parse_perf_event(dev, fpga_perf);
+	ret = parse_perf_attributes(dev, fpga_perf, "events");
```
r-rojo, line 282: Should the regex pattern used the same for calling this on line 278 and 282? Shouldn't the pattern depend on the attribute kind or is only one regex pattern valid?

r-rojo (approved):
We can refactor later.

r-rojo (approved):

PR #2092, opened 2021-06-26
host_exerciser: changes required for IRQ bringup

tswhison:

### DIFF
--- a/samples/host_exerciser/host_exerciser.h
+++ b/samples/host_exerciser/host_exerciser.h
@@ -324,6 +324,7 @@ public:
         "The Interrupt Vector Number for the device")
         ->transform(CLI::Range(0, 3))->default_val("0");
 
+    app_.add_option("--no-perf", no_perf_, "disable perf counters")->default_val("false");
   }
 
   virtual int run(CLI::App *app, test_command::ptr_t test) override
@@ -448,6 +449,7 @@ public:
   bool he_continuousmode_;
   uint32_t he_interleave_;
   uint32_t he_interrupt_;
+  bool no_perf_;
 
   std::map<uint32_t, uint32_t> limits_;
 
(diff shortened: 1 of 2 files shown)


samples/host_exerciser/host_exerciser.h
```diff
@@ -324,6 +324,7 @@ class host_exerciser : public test_afu {
         "The Interrupt Vector Number for the device")
         ->transform(CLI::Range(0, 3))->default_val("0");
 
+    app_.add_option("--no-perf", no_perf_, "disable perf counters")->default_val("false");
```
anandaravuri, line 327: can you change to --perf ? instead of --no-perf.

samples/host_exerciser/host_exerciser_cmd.h
```diff
@@ -270,21 +278,24 @@ class host_exerciser_cmd : public test_command
...
 
 
-
-
         std::cout << "Test Completed" << std::endl;
-        //stop performance counter
-        if (perf->stop() != FPGA_OK) {
-            std::cout << "Failed to stop the fpga perf counter" << std::endl;
-            return -1;
+        if (perf) {
```
anandaravuri, line 282: add perf print and stop into same if statement.
line 283 and 293 in sane if statement,
line 286 remove  return -1 and line 297

tools should compare  Compare buffer even perf fails ?

samples/host_exerciser/host_exerciser.h
tswhison: fixed

samples/host_exerciser/host_exerciser_cmd.h
tswhison: I think you want to stop the perf counters are soon as you detect end-of-test.  Otherwise, the counter values can be skewed, if they are time sensitive.

anandaravuri (approved):

PR #2100, opened 2021-07-06
tools: add n6010 smart NIC support  to fpgainfo

anandaravuri:


tools/fpgainfo/board.c
```diff
@@ -66,6 +66,7 @@ static platform_data platform_data_table[] = {
 	{ 0x8086, 0x0b2c, "libboard_d5005.so", NULL },
 	{ 0x8086, 0xaf00, "libboard_d5005.so", NULL },
 	{ 0x8086, 0xbcce, "libboard_d5005.so", NULL },
```
tswhison, line 68: What about this one (0x8086, 0xbcce)?  Does it need to be removed?

anandaravuri: 0x8086, 0xbcce -> IOFS R1 on D5005 card
0x1c2c, 0xbcce -> next gen N6010 card , vender id is different(silicom)

tswhison (changes requested):
I think this should be closed.  The decision was made not to change the Vendor ID.

PR #2102, opened 2021-07-07

ramyavex:
Added help message and Error message when application
running as a non-root user

### DIFF
--- a/samples/fpgaperf_counter/fpgaperf_counter.c
+++ b/samples/fpgaperf_counter/fpgaperf_counter.c
@@ -469,6 +469,15 @@ fpga_result fpgaPerfCounterGet(fpga_token token, fpga_perf_counter *fpga_perf)
 
 	memset(fpga_perf, 0, sizeof(fpga_perf_counter));
 
+	//check if its being run as root or not
+	uid_t uid = getuid();
+	if (uid != 0) {
+		fpga_perf->previlege = false;
+		return FPGA_OK;
+	} else {
+		fpga_perf->previlege = true;
+	}
+
 	ret = get_fpga_sbdf(token, &segment, &bus, &device, &function);
 	if (ret != FPGA_OK) {
 		OPAE_ERR("Failed to get sbdf");
@@ -558,6 +567,10 @@ fpga_result fpgaPerfCounterStartRecord(fpga_perf_counter *fpga_perf)
 		return FPGA_INVALID_PARAM;
 	}
 
+	if (!fpga_perf->previlege) {
+		return FPGA_OK;
+	}
+
 	if (fpga_perf_check_and_lock(fpga_perf)) {
 		OPAE_ERR("Failed to lock perf mutex");
 		return FPGA_EXCEPTION;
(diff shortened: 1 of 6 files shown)


samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -469,6 +469,15 @@ fpga_result fpgaPerfCounterGet(fpga_token token, fpga_perf_counter *fpga_perf)
 
 	memset(fpga_perf, 0, sizeof(fpga_perf_counter));
 
+	//check if its being run as root or not
```
tswhison, line 472: Hi Ramya, rather than checking here in the perf counter library, we should add the privilege check in the host_exerciser app.

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -558,6 +567,10 @@ fpga_result fpgaPerfCounterStartRecord(fpga_perf_counter *fpga_perf)
 		return FPGA_INVALID_PARAM;
 	}
 
+	if (!fpga_perf->previlege) {
```
tswhison, line 570: Then, instead of checking here and returning FPGA_OK (it should technically be an error the way we are defining it.), we would remove these checks and rely on the check in host_exerciser before the perf counters are queried.

samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -643,6 +660,12 @@ fpga_result fpgaPerfCounterPrint(FILE *f, fpga_perf_counter *fpga_perf)
...
```
tswhison, line 663: We should move this printout into the application.

samples/fpgaperf_counter/fpgaperf_counter.h
```diff
@@ -65,6 +65,7 @@ typedef struct {
 typedef struct {
 	pthread_mutex_t lock;
 	uint64_t magic;
+	bool previlege;
```
tswhison, line 68: Then, this flag can be removed.  Add the privilege check in host_exerciser_cmd.h:174, just before the perf counters are queried.

tswhison (approved):

samples/host_exerciser/host_exerciser_cmd.h
```diff
@@ -174,6 +174,12 @@ class host_exerciser_cmd : public test_command
 
 	fpgaperf::ptr_t perf(nullptr);
         if (host_exe_->perf_) {
+            uid_t uid = getuid();
+            if (uid != 0) {
+                std::cout <<"\nFailed to read Perf counter due to unprivileged user access"<<std::endl
+                <<"check --help for more information on setting the capabilities for binary\n" <<std::endl;
+                return -1;
+            }
```
r-rojo, line 182: Can't we still run but without perf counters?

tswhison: That would mean (in effect) editing the command line to the test.  Is this ok from a testing perspective, or is it better to error out and return a non-zero value?

r-rojo: Perhaps in the long term we can enhance it to make performance counters optional via command line and this check

r-rojo (approved):

PR #2109, opened 2021-07-13
Ramya/host exerciser rev5

ramyavex:
Checking the capability set or not for the binary.
If capability not set with non-root user then display the error. 
If capability set with non-root user then read the performance counter and display.

Added libcap-devel for capability support.

### DIFF
--- a/opae-libs/.github/workflows/ccpp.yml
+++ b/opae-libs/.github/workflows/ccpp.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Get Packages
       uses: mstksg/get-package@v1
       with:
-        apt-get: uuid-dev libjson-c-dev libhwloc-dev lcov linux-headers-generic
+        apt-get: uuid-dev libjson-c-dev libhwloc-dev lcov linux-headers-generic libcap-dev
     - name: CMake Configure
       run: cmake -B .build -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} -DOPAE_ENABLE_MOCK=ON -DOPAE_BUILD_TESTS=ON
     - name: make
--- a/opae-libs/cmake/modules/FindCap.cmake
+++ b/opae-libs/cmake/modules/FindCap.cmake
@@ -25,15 +25,18 @@
 ## ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
 ## POSSIBILITY OF SUCH DAMAGE
 
+find_package(PkgConfig)
+pkg_search_module(CAP libcap)
+
 find_path(LIBCAP_INCLUDE_DIRS
-  capability.h
-  PATHS /usr/include/sys
-  NO_DEFAULT_PATH)
+  NAMES sys/capability.h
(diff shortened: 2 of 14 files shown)


samples/host_exerciser/host_exerciser_cmd.h
```diff
@@ -175,7 +181,10 @@ class host_exerciser_cmd : public test_command
 	fpgaperf::ptr_t perf(nullptr);
         if (host_exe_->perf_) {
             uid_t uid = getuid();
-            if (uid != 0) {
+            caps = cap_get_file("/usr/bin/host_exerciser");
```
tswhison, line 184: Hi Ramya, the file may not always be under /usr/bin/.  Do we need the full path?

samples/host_exerciser/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ opae_add_executable(TARGET host_exerciser
         opae-c
         opae-cxx-core
         afu-test
+        cap
```
tswhison, line 33: Before we can link with this new library, we need to implement a finder.  Please see opae-sdk/opae-libs/cmake/modules/Find*

opae-libs/.github/workflows/ccpp.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     - name: Get Packages
```
tswhison, line 22: This file lives in opae-libs, so it must be updated at https://github.com/OPAE/opae-libs.git.

opae-libs/cmake/modules/FindCap.cmake
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/cmake -P
```
tswhison, line 1: This file is being added to opae-libs, so it must be added at https://github.com/OPAE/opae-libs.git.

opae-libs/cmake/modules/OPAE.cmake
```diff
@@ -52,6 +52,7 @@ if(NOT libuuid_FOUND)
 endif()
```
tswhison, line 52: This file lives in opae-libs, so it must be updated at https://github.com/OPAE/opae-libs.git.

tswhison (changes requested):

opae-libs/.github/workflows/ccpp.yml
ramyavex: Done the changes and pushed to opae-libs.git

opae-libs/cmake/modules/FindCap.cmake
ramyavex: Done the changes and pushed to opae-libs.git

opae-libs/cmake/modules/OPAE.cmake
ramyavex: Done the changes and pushed to opae-libs.git

tswhison:
Hi Ramya, you'll need to unstage the changes to the files that were submitted to opae-libs.  Either unstage them here in this pull request or else create a new pull request without them.

ramyavex:
creating a new pull request to resolve conflicts

PR #2126, opened 2021-07-21

ramyavex:
Checking the capability set or not for the binary.
If capability not set with non-root user then display the error.
If capability set with non-root user then read the performance counter and display.

Added libcap-devel for capability support.


samples/CMakeLists.txt
```diff
@@ -41,6 +41,10 @@ endif()
 # Add the samples ####################################################
 ######################################################################
 
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
```
tswhison, line 44: Is this required, given that it has been included in OPAE.cmake?

opae.spec
```diff
@@ -84,6 +85,7 @@ mkdir -p %{buildroot}%{_usr}/src/opae/opae-libs/cmake/modules
 for s in FindHwloc.cmake \
          OPAE.cmake \
          FindUUID.cmake \
```
tswhison, line 87: Add FindUdev.cmake here?

tswhison (approved):

tswhison (dismissed):

tswhison (approved):

r-rojo (commented):
is this meant to determine if the process is being run by root or if the effective permission the process is being run has access to the counters?

r-rojo (approved):

samples/fpgaperf_counter/fpgaperf_counter.h
```diff
@@ -47,6 +47,7 @@ extern "C" {
 
 #define FPGA_PERF_MAGIC 	0x46504741584c474c
 #define DFL_PERF_STR_MAX	256
+#define CAP_FPGA_PERFMON	38
```
r-rojo, line 50: if this could be defined in other versions of the libcap header, then maybe we should check if it's defined before redefining it here.

r-rojo (approved):

tswhison (approved):

PR #2129, opened 2021-07-22
add pmci-bus to support PMCI for fpgasupdate

Figo-zhang:
Adding pmci-bus to support PMCI for fpgasupdate on
Arrow Creek.

Signed-off-by: ac_adp <tianfei.zhang@intel.com>


python/opae.admin/opae/admin/fpga.py
```diff
@@ -218,6 +218,16 @@ def spi_bus(self):
                     return spi
         return self.find_one('spi*/spi_master/spi*/spi*')
 
+    @property
+    def pmci_bus(self):
+        if os.path.basename(self.sysfs_path).startswith('dfl'):
+            patterns = ['dfl*.*/*n6010bmc-secure*']
```
tswhison, line 224: Rather than **n6010bmc-secure**, could we just use **-secure**?  This would make things forward compatible.

python/opae.admin/opae/admin/fpga.py
```diff
@@ -218,6 +218,16 @@ def spi_bus(self):
...
+            for pattern in patterns:
+                pmci = self.find_one(pattern)
+                if pmci:
+                    return pmci
+        return self.find_one('n6010bmc-secure*')
```
tswhison, line 229: Same here.  Explicitly naming n6010bmc causes a compatibility issue.

PR #2135, opened 2021-07-23
cpeng: check for hugepages

r-rojo:
Add try/catch around allocate call and if an exception is caught, then
compare the byte size requested to the syttem page size. If the amount
is bigger than system page size, print more info about maybe requiring
hugepages reserved.

### DIFF
--- a/ofs/apps/cpeng/hps.cpp
+++ b/ofs/apps/cpeng/hps.cpp
@@ -119,7 +119,7 @@ public:
     } catch (opae_exception &ex) {
       log_->error("could not allocate {} bytes of memory", chunk);
       if (chunk > pg_size) {
-	auto hugepage_sz = chunk <= MB(2) ? "2MB" : "1GB";
+        auto hugepage_sz = chunk <= MB(2) ? "2MB" : "1GB";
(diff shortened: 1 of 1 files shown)


ofs/apps/cpeng/hps.cpp
```diff
@@ -106,7 +113,17 @@ class cpeng : public opae::afu_test::command
...
     // otherwise, use the smaller of chunk_ and file size
     size_t chunk = chunk_ ? std::min(static_cast<size_t>(chunk_), sz) : sz;
-    auto buffer = shared_buffer::allocate(afu->handle(), chunk);
+    shared_buffer::ptr_t buffer(0);
+    try {
+      buffer = shared_buffer::allocate(afu->handle(), chunk);
+    } catch (opae_exception &ex) {
+      log_->error("could not allocate {} bytes of memory", chunk);
+      if (chunk > pg_size) {
+	auto hugepage_sz = chunk <= MB(2) ? "2MB" : "1GB";
```
tswhison, line 122: indent is off

r-rojo: ah! fixed :)

tswhison (approved):

PR #2152, opened 2021-07-30
Rhel rpm

trixirt:
Fix building on rhel 8.2
Verify also building on fedora 31

### DIFF
--- a/opae-libs/CMakeLists.txt
+++ b/opae-libs/CMakeLists.txt
@@ -115,6 +115,9 @@ mark_as_advanced(OPAE_PRESERVE_REPOS)
 option(OPAE_BUILD_LIBOFS "Enable building of the libofs library and framework" ON)
 mark_as_advanced(OPAE_BUILD_LIBOFS)
 
+option(OPAE_BUILD_AFU_TEST "Enable building of afu_test" ON)
+mark_as_advanced(OPAE_BUILD_AFU_TEST)
+
 if(OPAE_BUILD_TESTS)
     set(OPAE_TEST_TAG "master" CACHE STRING "Desired branch for opae-test")
     mark_as_advanced(OPAE_TEST_TAG)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,7 +61,11 @@ endif()
 ############################################################################
 opae_add_subdirectory(opae-libs)
 opae_add_subdirectory(external)
-opae_add_subdirectory(afu-test)
+
+if (OPAE_BUILD_AFU_TEST)
+  opae_add_subdirectory(afu-test)
+endif()
+
 opae_add_subdirectory(platforms)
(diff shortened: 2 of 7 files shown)


r-rojo:
Is the intent here to remove dependencies on CLI11 and spdlog?
I thought that modern fedora versions included these packages. Does fedora31 not include them?
The afu_test is reused in a lot of our tools because it sets up things like command line parsing and logging so removing inclusion of afu-test would essentially break the tools that derive from it.

trixirt:
rhel does not have these devel packages.
and using external/ breaks its build

r-rojo:
@trixirt , can you elaborate by what you mean by "break its build"?
Also, are you saying that some packages that go into fedora are rejected by rhel? I was under the impression that making it into fedora was a requirement to get into rhel.

trixirt:
The linux-dfl-backport driver is an out a tree kernel driver running on rhel from 8.2+
It requires an opae that also runs on rhel from 8.2+

When the devel packages come from the distro they are a version that works with the distro version
When they come from external/ they are a single tag that works with a subset of distros.

The specific problem is
CMake Error at spdlog/download/spdlog-prefix/tmp/spdlog-gitclone.cmake:106 (message):
  Failed to copy script-last-run stamp file:
  '.../opae-sdk/packaging/opae/rpm/rpmbuild/BUILD/opae-2.0.1-1/external/spdlog/download/spdlog-prefix/src/spdlog-stamp/spdlog-gitclone-lastrun.txt'

embedding version logic for external/ cmake's is problematic, you would have to keep up a distro x distro version x external pkg version matrix.
In this case it was noted that most of external/ is used to build only afu_test and a few other derivative tests.  Removing these will not effect the in production use case.

michelleyho (approved):

opae-libs/CMakeLists.txt
```diff
@@ -115,6 +115,9 @@ mark_as_advanced(OPAE_PRESERVE_REPOS)
 option(OPAE_BUILD_LIBOFS "Enable building of the libofs library and framework" ON)
 mark_as_advanced(OPAE_BUILD_LIBOFS)
 
+option(OPAE_BUILD_AFU_TEST "Enable building of afu_test" ON)
```
tswhison, line 118: Hi Tom, since this file lives in opae-libs, there needs to be a separate pull request filed at github.com/OPAE/opae-libs.git.

tswhison:
@trixirt In order to set up CI around this, we need a RHEL .iso.  How can we get one?

tswhison:
Hi @trixirt, I was able to acquire the necessary images and license..  When I try to run create from this branch, I get errors about missing dependencies doxygen and hwloc-devel.  I think the doxygen dependency can be worked around, but how are you able to build fpgabist in rhel8 w/o hwloc-devel?  Were you able to install it somehow?

PR #2159, opened 2021-08-05
Handle AFU JSON that describes multiple VFs

michael-adler:


michael-adler:
@anandaravuri and @tswhison , this comes from discussion in a meeting today and extending JSON to support multiple VFs.

@matthew-gerlach attended.

michael-adler:
The summary email:

It turned out to be even simpler than anticipated. We can reasonably argue that the proper level for a vector is on “accelerator-clusters”, which already is a vector! So,
 
        "accelerator-clusters": [
            {
                "accelerator-type-uuid": "222baa4c-0ab8-4574-bf74-aa35b945a223",
                "name": "ofs_pr_default",
                "total-contexts": 1
            }
        ],
 
becomes:
 
        "accelerator-clusters": [
            {
                "accelerator-type-uuid": "222baa4c-0ab8-4574-bf74-aa35b945a223",
                "name": "ofs_pr_default",
                "total-contexts": 1
            },
            {
                "accelerator-type-uuid": "a5c39b7d-5d89-4ce1-a95f-e71912109f7d",
                "name": "some_other_afu",
                "total-contexts": 1
            }
        ],
 
in a multi-VF AFU. We can argue that “total-contexts” is a relic from the days when we thought that the AFU would manage multiple virtual contexts without the VF mechanism. The JSON schema currently requires it, but we could drop that and remove “total-contexts” – treating its absence as meaning the default.
 
The updated tools now generate both an indexed version of preprocessor macros and, for compatibility, a non-indexed entry for backward compatibility. The C header generated for the above JSON by afu_json_info would contain:
 
#define AFU_ACCEL_NAME "ofs_pr_default"
#define AFU_ACCEL_NAME0 "ofs_pr_default"
#define AFU_ACCEL_NAME1 "some_other_afu"
#define AFU_ACCEL_UUID "222baa4c-0ab8-4574-bf74-aa35b945a223"
#define AFU_ACCEL_UUID0 "222baa4c-0ab8-4574-bf74-aa35b945a223"
#define AFU_ACCEL_UUID1 "a5c39b7d-5d89-4ce1-a95f-e71912109f7d"

anandaravuri (approved):

r-rojo (approved):
Do we have a schema file for this?

tswhison:
Hi @michael-adler , could you run one of these bitstreams through fpgaconf to see what if anything we need to change in libbitstream / fpgaconf?  I suspect there might be some work there as well.

michael-adler:
@tswhison I just tested fpgaconf and it works without problems. I don't think fpgaconf reads the AFU UUID. It only needs the interface UUID, which is unchanged.

We should consider dropping "total-contexts" as a required field in the schema.

PR #2169, opened 2021-08-17
n6010 image info

hundeboll-silicom:
This depends on https://github.com/OPAE/opae-libs/pull/233


Anders5-dk (dismissed):
I think you mean "Prints fpga image info" insrtead of "Prints fpga info info".

hundeboll-silicom:
> I think you mean "Prints fpga image info" insrtead of "Prints fpga info info".

I updated the commit to fix the typo, and also updated the wrong return value description.

tools/fpgainfo/board.c
```diff
@@ -690,3 +690,25 @@ fpga_result fpga_boot_info(fpga_token token)
...
+fpga_result fpga_image_info(fpga_token token)
+{
+	fpga_result res = FPGA_OK;
+	void *dl_handle = NULL;
+
+	fpga_result (*fpga_image_info)(fpga_token token);
+
+	res = load_board_plugin(token, &dl_handle);
+	if (res != FPGA_OK) {
+		OPAE_MSG("Failed to load board plugin\n", dlerror());
```
tswhison, line 703: This needs a format specifier (%s) for the dlerror().  I think that dlerror() can also return NULL, so there needs to also be a NULL check.

hundeboll-silicom: I've updated the format string, and added a NULL check

tswhison: Could you push the changes?  I don't see them.

hundeboll-silicom: Sorry, the change was still chilling in my working tree :(

tswhison (approved):

PR #2195, opened 2021-08-23
Tools: Update Host Exerciser sample with DSM  performance counter.

anandaravuri:
-OFS release doesn’t support any global perf counters, so removed global perf counters code from Host Exerciser.

  -Host Exerciser  reads Performance Counter  from DSM Status buffer.
      DSM Status fromat:
	Field Name	Range	Width	Description
	Reserved	[511:288]	224	Reserved
	Ab_error_info	[287:256]	32	Arbiter error info field
	Reserved	[255:232]	24	Reserved
	Penalty_end	[231:224]	8	Test end overhead in # clks
	Reserved	[223:208]	16	Reserved
	Penalty_start	[207:192]	16	Test start overhead in # clks
	Num_writes	[191:160]	32	Total number of Writes sent / Total Num CX sent
	Num_reads	[159:128]	32	Total number of Reads sent
	Reserved	[127:104]	24	Reserved
	Num_ticks	[103:64]	40	Number of clocks
	ErrorVector	[63:32]	32	Errors detected
	Reserved	[31:16]	16	Reserved
	Dsm_number	[15:1]	15	Unique id for each dsm status write
	TestCompleted	[0]	1	Test completion flag

	host_exerciser  --perf true  lpbk
	[2021-08-23 14:14:36.624] [lpbk] [info] starting test run, count of 1
	Input Config:0
	Allocate SRC Buffer
	Allocate DST Buffer
	Allocate DSM Buffer
	Start Test
	Test Completed
	Host Exerciser swtest msg:0

	****Host Exerciser Performance Counter****
        Host Exerciser numReads:1024
        Host Exerciser numWrites:1025
        Host Exerciser numPendReads:0
       Host Exerciser numPendWrites:0
       Number of clocks:7283
       Total number of Reads sent:1024
       Total number of Writes sent :1022
       Bandwidth: 2.25 GB/s

       [2021-08-23 14:14:36.626] [lpbk] [info] Test lpbk(1): PASS

### DIFF
--- a/opae.spec.in
+++ b/opae.spec.in
@@ -120,10 +120,6 @@ mv %_topdir/tmpBRoot $RPM_BUILD_ROOT
 mkdir -p /etc/ld.so.conf.d
 echo "@LDCONFIG_DIR@" > /etc/ld.so.conf.d/opae-c.conf
 ldconfig
-if [ -x %{_sbindir}/setcap ];
-then
-  %{_sbindir}/setcap "38,cap_sys_ptrace,cap_syslog+eip" %{_bindir}/host_exerciser
-fi
 
 %postun
 @CPACK_RPM_SPEC_POSTUNINSTALL@
--- a/samples/host_exerciser/CMakeLists.txt
+++ b/samples/host_exerciser/CMakeLists.txt
@@ -30,14 +30,7 @@ opae_add_executable(TARGET host_exerciser
         opae-c
         opae-cxx-core
         afu-test
-        fpgaperf_counter
         ${libjson-c_LIBRARIES}
         ${libuuid_LIBRARIES}
-        ${LIBCAP_LIBRARIES}
     COMPONENT samplebin
-)
-
-target_include_directories(host_exerciser
-    PRIVATE
-        ${OPAE_SDK_SOURCE}/samples/fpgaperf_counter
 )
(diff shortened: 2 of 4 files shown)


tswhison (approved):

michelleyho (approved):

michael-adler:
@anandaravuri Nice improvement since it runs without root.

Since it doesn't require root, why not make "--perf true" the default?

I still can't figure out how to make host_exerciser run long enough to be of value. There is a switch for continuous mode, but that causes it to run until timeout and then doesn't print statistics. It also has a nasty habit of crashing the machine, since OFS doesn't handle traffic that's already inflight during an abort. 1024 transactions is not enough for a real measurement.

anandaravuri:
> @anandaravuri Nice improvement since it runs without root.
> 
> Since it doesn't require root, why not make "--perf true" the default?
> 
> I still can't figure out how to make host_exerciser run long enough to be of value. There is a switch for continuous mode, but that causes it to run until timeout and then doesn't print statistics. It also has a nasty habit of crashing the machine, since OFS doesn't handle traffic that's already inflight during an abort. 1024 transactions is not enough for a real measurement.

[Adler] 
Max number of cache lines  1024 in HE Loopback

michael-adler:
> > @anandaravuri Nice improvement since it runs without root.
> > Since it doesn't require root, why not make "--perf true" the default?
> > I still can't figure out how to make host_exerciser run long enough to be of value. There is a switch for continuous mode, but that causes it to run until timeout and then doesn't print statistics. It also has a nasty habit of crashing the machine, since OFS doesn't handle traffic that's already inflight during an abort. 1024 transactions is not enough for a real measurement.
> 
> [Adler]
> Max number of cache lines 1024 in HE Loopback

It can't loop around? Seriously?

I have NLB working. Perhaps a better option.

PR #2210, opened 2021-08-31
packager: move to Python setuptools project

tswhison:
Moving packager to a Python setuptools project allows it to be installed
from the source tree, eg
$ pip install opae-sdk/python/packager


michael-adler:
@tswhison Interesting. I was, just today, trying to figure out how to deal with the Python tools in /platforms in this repository because I want an easy way for HW developers to use a few scripts without having to teach them to compile the whole OPAE SDK.

Will take a look at this soon. Maybe those scripts should get setuptools wrappers, too.

r-rojo:
Just as an FYI, you can also pip install from a VCS URL:

For example, to use git:
```bash
pip install  "git+https://github.com/OPAE/opae-sdk.git#egg=packager&subdirectory=python/packager"
```

You can see more docs here:
https://pip-python3.readthedocs.io/en/latest/reference/pip_install.html#git

r-rojo (approved):

PR #2219, opened 2021-09-09
Tools: add  hssi stats clear command and print active port stats

anandaravuri:
- add hssi stats command line argument  --clear or -C ,  reset MAC statistics
   SAL command to clear MAC statistic registers value per port. User can choose to reset either Tx or Rx MAC statistic only or reset both Tx and Rx MAC statistic together by setting bit 16 (Tx) and 17 (Rx) on HSSI Control/Address Register.

- added new CSR field enabled ports  to HSSI Sub system feature list CSR
   Bit 21:6
   Physical Port Enable
   [6] - Port 0 Enable
   [7] - Port 1 Enable
   :
   [21] - Port 15 Enable


tools/extra/hssi/ethernet/hssistats.py
```diff
@@ -37,6 +37,8 @@
 import mmap
 from ethernet.hssicommon import *
 
+# Sleep 50 millisceonds afte clearing stats.
```
tswhison, line 40: milliseconds after

Also, you're dividing by 1 million.  Wouldn't that be microseconds?

tswhison (dismissed):

tools/extra/hssi/ethernet/hssistats.py
tswhison: Fix typos:
```
milliseconds after
```

anandaravuri: fixed

tswhison (approved):

PR #2194, opened 2021-08-21
Rpm fixes fedora

trixirt:
several fixes to get rpm building again on fedora 33

PR #2221, opened 2021-09-11
Rfc tests package

trixirt:
Only the last commit is new, it builds on this pr https://github.com/OPAE/opae-sdk/pull/2194

This pr introduces a new rpm package *-tests.rpm where the opae unit test can be packaged as well as other misc tests and not for normal release packages.  There are some problems building the unit tests now so they are not included in the file list. So an existing test n5010-ddr-test is used to start the -tests rpm.


tswhison:
Thanks for this, Tom.  As discussed, we will likely not package the unit tests, as they are driven by ctest - there is no scalable way to run them after installing them.

PR #2267, opened 2021-09-29
tools: fix static code analysis Buffer Overflow - Array Index Out of …

anandaravuri:
…Bounds


rchriste1 (approved):

tswhison (approved):

Anders5-dk (approved):
Ok. But I think it would be much easier to read the code if the for loop was expanded into the four individual lines.

PR #2281, opened 2021-10-05
Tools: add ofsreg tool to dump pcie, hssi and bmc registers

anandaravuri:
usage: ofsreg.py [-h] [--pcie-address PCIE_ADDRESS] [-r {pcie,bmc,hssi,all}]
optional arguments:
  -h, --help            show this help message and exit
  --pcie-address PCIE_ADDRESS, -P PCIE_ADDRESS
                        sbdf of device to program (e.g. 0000:04:00.0).
                        Optional when one device in system.
  -r {pcie,bmc,hssi,all}, --reg {pcie,bmc,hssi,all}
                        choose reg dump. Default: pcie

 - PCIe: Memory map pcie  bar0  , parse DFH and print csr’s
 - HSSI: Find HSSI feature and dump csr’s


tswhison:
I think we should come at this from a different angle.  We have tools that are capable of dumping PCIe CSRs.  Why not update them, instead of recreating that functionality here?

PR #2293, opened 2021-10-11
packaging: fix RHEL & Fedora RPM builds

tswhison:
- Split out a separate .spec file for Fedora, which is less restrictive in
terms of package availability.
- Update Requires and BuildRequires for both .spec files.
- Parameterize the create script to identify which flavor to build.
- Implements pull requests #2152 and #2194.

### DIFF
--- a/opae-libs/cmake/modules/OPAEGit.cmake
+++ b/opae-libs/cmake/modules/OPAEGit.cmake
@@ -1,5 +1,5 @@
 #!/usr/bin/cmake -P
-## Copyright(c) 2017-2020, Intel Corporation
+## Copyright(c) 2017-2021, Intel Corporation
 ##
 ## Redistribution  and  use  in source  and  binary  forms,  with  or  without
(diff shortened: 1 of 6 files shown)


opae.spec.fedora
```diff
@@ -0,0 +1,328 @@
...
+         Findjson-c.cmake \
+         OPAECompiler.cmake \
+         OPAEGit.cmake \
+         OPAEPackaging.cmake 
+do
+  cp "opae-libs/cmake/modules/${s}" %{buildroot}%{_usr}/src/opae/opae-libs/cmake/modules
+  chmod a+x %{buildroot}%{_usr}/src/opae/opae-libs/cmake/modules/$s
+done
+
+mkdir -p %{buildroot}%{_usr}/src/opae/samples
```
anandaravuri, line 120: please  remove samples folder  and scource code  from spec file
Fedora rom spec create SRC rpm

PR #2295, opened 2021-10-12
opae.admin: updates for sysfs path changes

tswhison:
*-secure.*.auto becomes *-sec-update.*.auto, so update the glob
to *-sec*.*.auto to account for both.

Rename the 'security' property accessor to 'secure_update'. This
corresponds to the base of the *-sec*.*.auto directory tree.

Rename the 'secure_dev' property accessor to 'upload_dev',
reflecting that it is the means by which to determine the
/dev/fpga_image_load* for loading secure images.

### DIFF
--- a/python/opae.admin/opae/admin/fpga.py
+++ b/python/opae.admin/opae/admin/fpga.py
@@ -372,11 +372,11 @@ class port(region):
         return self.node('afu_id').value
 
 
-class secure_dev(region):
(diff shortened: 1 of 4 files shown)


r-rojo (approved):
Is this backwards compatible?

tswhison:
Yes.  The globs were updated so that they would be.

PR #2330, opened 2021-10-21
doc: Add ToD feature usage document

raghavendraak:
Signed-off-by: raghavendraak <raghavendrax.anand.khadatare@intel.com>

### DIFF
--- a/samples/n5010-test/n5010-test.c
+++ b/samples/n5010-test/n5010-test.c
@@ -107,6 +107,12 @@ static fpga_result fpga_open(struct n5010 *n5010)
 		return res;
 	}
 
+	if (num == 0) {
+		res = FPGA_NOT_FOUND;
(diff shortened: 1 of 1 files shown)


doc/src/tod/tod.md
```diff
@@ -0,0 +1,200 @@
...
+1.  Install the linuxptp package:
+```console
+# sudo yum install linuxptp
+```
+2.  Check PTP device is created successfully by the ToD driver. 
+
+ToD driver registering as PHC device (clock_name: dfl_tod) to the Linux PTP stack and exposing to the Linux kernel to synchronize the system clock to its ToD information.
+```console
+# cat /sys/class/ptp/ptp0/clock_name
+# dfl_tod
```
tswhison, line 99: Remove the # here.  I think dfl_tod is the output of the cat command.

PR #2343, opened 2021-10-29
Tools: add SDM provision status info to  fpgainfo security

anandaravuri:
- fpgainfo reads SDM provision status and  prints values
- change bmc and security  output space format


tswhison (approved):
👍

PR #2356, opened 2021-11-03
tools: fix static code analysis  overflow of the buffer

anandaravuri:
-  Width is not specified for 's' conversion specifier.
   This can result in an overflow of the buffer provided in argument 3 of a call to 'fscanf'

### DIFF
--- a/samples/fpgaperf_counter/fpgaperf_counter.c
+++ b/samples/fpgaperf_counter/fpgaperf_counter.c
@@ -173,7 +173,7 @@ STATIC fpga_result parse_perf_attributes(struct udev_device *dev,
 			globfree(&pglob);
 			return FPGA_EXCEPTION;
 		}
-		if (fscanf(file, "%s", attr_value) != 1) {
+		if (fscanf(file, "%127s", attr_value) != 1) {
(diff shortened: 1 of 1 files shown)


samples/fpgaperf_counter/fpgaperf_counter.c
```diff
@@ -173,7 +173,7 @@ STATIC fpga_result parse_perf_attributes(struct udev_device *dev,
 			globfree(&pglob);
 			return FPGA_EXCEPTION;
 		}
-		if (fscanf(file, "%s", attr_value) != 1) {
+		if (fscanf(file, "%127s", attr_value) != 1) {
```
tswhison, line 176: attr_value is 128 bytes?

anandaravuri: Vulnerable code example
  void main() {
      char s[16];
      scanf("%s",s);
 }

Fixed code example
  void main() {
      char s[16];
      scanf("%15s",s);
 }
In the fixed code, width of the string is provided correctly, ensuring that the destination buffer won't overflow.

tswhison (approved):

PR #2367, opened 2021-11-05
packaging, rpm: remove exclude clause from tar

tswhison:
Preserves any .so's archived in the repo so that they can be
properly distributed with python projects.

### DIFF
--- a/packaging/opae/rpm/create-tarball.sh
+++ b/packaging/opae/rpm/create-tarball.sh
@@ -21,7 +21,6 @@ tar --transform=$trans \
   --exclude=build \
   --exclude=.* \
   --exclude=*~ \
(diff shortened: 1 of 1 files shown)


trixirt:
*.so's should not be in the repo, including them breaks security.

tswhison:
> *.so's should not be in the repo, including them breaks security.

We've discussed this before.  The situation hasn't changed.  We are in compliance with our legal teams' requirements.

trixirt:
delivering a *.so in a source tarball will make the tarball toxic.
Can the source be added to the third party dir like the others  and built as-needed?

tswhison:
Let's take the discussion to our meeting this week.

PR #2372, opened 2021-11-11
Update fpgainfo events to latest register specs

sc-ele:
Update fpgainfo events to latest register specs.

The failing check does not seem to have anything to do with my code. It complains about missing doxygen, kernel headers not installed, etc.


tools/libboard/board_n6000/board_event_log.c
```diff
@@ -218,6 +219,27 @@ static void bel_print_header(const char *label, struct bel_header *header)
 	printf("  " BEL_LABEL_FMT "%s", 50, label, time_str);
 }
 
+static void reserved_field(void *p, uint32_t value, size_t first, size_t last)
+{
+	UNUSED_PARAM(p);  // Need dummy paramater to cheat the coding style check!
```
tswhison, line 224: Please don't do this.  The coding style check is there for a reason.  What is the issue?

tools/libboard/board_n6000/board_event_log.c
```diff
@@ -218,6 +219,27 @@ static void bel_print_header(const char *label, struct bel_header *header)
...
+	uint32_t mask = UINT_MAX >> (32 - (last - first));
+	uint32_t field = (value >> first) & mask;
+
+	if (field != 0)
+		printf("      *** RESERVED FIELD [%lu:%lu] IS NOT ZERO: 0x%X\n", last - 1, first, field);
+}
+
+static void reserved_bit(void *p, uint32_t value, size_t offset)
+{
+	UNUSED_PARAM(p);  // Need dummy paramater to cheat the coding style check!
```
tswhison, line 235: Please don't do this.  Why is the coding style check being subverted?

tswhison (changes requested):

sc-ele:
In the calls to these functions, I wanted to align the value parameter in the same column as the surrounding calls as it looks much nicer and helps with readability. Coding style check complains about this, it does not allow spaces where I put a dummy NULL parameter. It does not understand esthetics.

Sure, I can change it to the ugly look if you insist.

/Esa
________________________________

sc-ele:
I have changed the code slightly so that the first label parameter is actually used.

Don't know why the code checks fail, don't seem to have anything to do with my code but the build environment as far as I can see.

tswhison (approved):

PR #2383, opened 2021-11-16
spdlog: update external version to 1.9.2

tswhison:


tswhison:
Not updating now, as spdlog fails CI for CentOS7.X and Fedora3X.

PR #2387, opened 2021-11-16
PACSign: output more info when libcrypto not found

tswhison:
When the openssl hsm_manager fails to find libcrypto.so, output a
message describing the packages required to install the library.


michael-adler:
Well, the message is better but not actually true. There must be a way to search for a library with a minimum version.

michael-adler:
> It's true for distros where no libcrypto.so is found. Installing openssl-devel will remedy that. It will not work as a remedy for all scenarios, particularly when the distro doesn't natively support a secure version of the library; but isn't it better than the prior stack trace.

Any user of a distro where both 1.1.1 and an older version are both available will be quite confused.

anandaravuri (approved):

r-rojo (approved):

PR #2413, opened 2021-12-03
Rrojo/walk dump

r-rojo:
Add the following options to the 'walk' command:
* dump: dump registers in between dfh's
* count: stop walk after count is reached
* delay: wait 'delay' seconds after each mmio read

### DIFF
--- a/tools/extra/opae.io/opae/io/utils.py
+++ b/tools/extra/opae.io/opae/io/utils.py
@@ -288,7 +288,7 @@ def walk(region, offset=0, show_uuid=False, dump=False):
         if show_uuid:
             print(f'    uuid: {read_guid(region, offset_+0x8)}')
         if dump and not hdr.eol:
-            sz = hdr.next - offset_ if not hdr.eol else region.size-offset_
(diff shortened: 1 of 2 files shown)


tools/extra/opae.io/opae/io/utils.py
```diff
@@ -281,12 +281,25 @@ def dfh_walk(region, offset=0, header=dfh0, guid=None):
...
+    for (offset_, hdr), i in enumerate(dfh_walk(region, offset=offset)):
+        if count and i >= count:
+            break
         print(f'offset: 0x{offset_:04x}, value: 0x{hdr.value:04x}')
         print(f'    dfh: {hdr}')
         if show_uuid:
             print(f'    uuid: {read_guid(region, offset_+0x8)}')
+            if delay:
+                time.sleep(delay)
+        if delay:
```
tswhison, line 295: Can the delay at 293 be omitted, because we're delaying here again?

r-rojo: I had intended a read after every mmio read but I think I have to have a delay (for the header) before the delay for reading the uuid.

tswhison (approved):

PR #2456, opened 2022-01-12
rpm: install hssi_pkt_filt_*.h

trixirt:
Fixes this error on rhel 8.5

error: Installed (but unpackaged) file(s) found:
   /usr/src/opae/samples/hssi/hssi_pkt_filt_100g_cmd.h
   /usr/src/opae/samples/hssi/hssi_pkt_filt_10g_cmd.h

Signed-off-by: Tom Rix <trix@redhat.com>

PR #2457, opened 2022-01-12
rhel rpm: add missing hssi files

tswhison:


tswhison:
This is a copy of #2456, because that one is from a fork, which does not play well with CI.

anandaravuri (approved):

PR #2465, opened 2022-02-02
opae.io: fix dump for binary format

r-rojo:
Convert register values to bytes when dumping to binary format.


fpgamatt (approved):
The fix makes sense based on the error I saw.

tswhison (approved):

PR #2479, opened 2022-02-11
opae.io: handle sysfs exceptions

tswhison:
Writing to /sys/bus/pci/drivers/vfio-pci/bind will fail if the
driver enumeration doesn't claim the device.

In recent kernels, eg 5.15, writing the same VID/DID multiple times
to /sys/bus/pci/drivers/vfio-pci/new_id will result in EEXIST. Catch
EEXIST and continue to process the device, else exit on other OSError.


rweight (approved):
The changes look good to me!

anandaravuri (approved):

r-rojo (approved):

PR #2509, opened 2022-03-04
opae-sdk: remove ase rpm from spec file

anandaravuri:

### DIFF
--- a/opae.spec.fedora
+++ b/opae.spec.fedora
@@ -16,7 +16,7 @@ Requires:  systemd
 Requires:  tbb
 Requires:  hwloc
 Requires:  spdlog
-Requires:  edit
+Requires:  libedit
 
 URL:            https://github.com/OPAE/%{name}-sdk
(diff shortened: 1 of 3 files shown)


opae.spec.fedora
```diff
@@ -16,7 +16,7 @@ Requires:  systemd
 Requires:  tbb
 Requires:  hwloc
 Requires:  spdlog
-Requires:  libedit
+Requires:  edit
```
tswhison, line 19: We want libedit.  Remember from before?

tswhison (dismissed):
You could have simply done a
$ git revert
on the commit where I added these changes.  It would have been cleaner, and would have removed the issue of injecting new changes that are unrelated, eg the libedit change.

opae.spec.fedora
anandaravuri: rpmlint error opae.x86_64: E: explicit-lib-dependency libedit
changed libedit -> edit

tswhison: package libedit and edit are different.  We discussed this before.  Please change it back to libedit.

anandaravuri: changed  back to libedit.

tswhison (approved):

PR #2553, opened 2022-04-12
gtest: INSTANTIATE_TEST_CASE_P is deprecated

tswhison:
Replace with INSTANTIATE_TEST_SUITE_P, per the gtest docs.
Set the gtest variables correctly when manually fetching gtest.

Link with debug gtest libraries when CMAKE_BUILD_TYPE is Debug or
RelWithDebInfo.

Fix bug in object_api tests.  Reset cleanup_size to 0 at the beginning
of each test, because main() is called multiple times.

### DIFF
--- a/cmake/modules/OPAE.cmake
+++ b/cmake/modules/OPAE.cmake
@@ -1,5 +1,5 @@
 #!/usr/bin/cmake -P
-## Copyright(c) 2020, Intel Corporation
+## Copyright(c) 2020-2022, Intel Corporation
 ##
 ## Redistribution  and  use  in source  and  binary  forms,  with  or  without
 ## modification, are permitted provided that the following conditions are met:
@@ -57,7 +57,7 @@ include(FindUdev)
 
 include(OPAECompiler)
 if(OPAE_BUILD_TESTS)
-    find_package(GTest 1.11.0)
+    find_package(GTest ${GTEST_VERSION})
 endif(OPAE_BUILD_TESTS)
 include(OPAETest)
 include(OPAEPackaging)
--- a/cmake/modules/OPAE.cmake
+++ b/cmake/modules/OPAE.cmake
@@ -57,7 +57,7 @@ include(FindUdev)
 
 include(OPAECompiler)
 if(OPAE_BUILD_TESTS)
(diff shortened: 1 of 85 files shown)


anandaravuri (approved):

cmake/modules/OPAE.cmake
```diff
@@ -57,7 +57,7 @@ include(FindUdev)
 
 include(OPAECompiler)
 if(OPAE_BUILD_TESTS)
-    find_package(GTest 1.8.0)
+    find_package(GTest 1.11.0)
```
r-rojo, line 60: Should we store this in a cmake variable? That way when we change versions we only have to change it in one place?

tests/framework/mock/opae_fixtures.cpp
```diff
@@ -0,0 +1,46 @@
...
+// INTERRUPTION)  HOWEVER CAUSED  AND ON ANY THEORY  OF LIABILITY,  WHETHER IN
+// CONTRACT,  STRICT LIABILITY,  OR TORT  (INCLUDING NEGLIGENCE  OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+#include "opae_fixtures.h"
+
+namespace opae {
+namespace testing {
+
+uint32_t opae_p::num_tokens_for(fpga_properties filter)
```
r-rojo, line 31: What is the intent of this class?

cmake/modules/OPAE.cmake
tswhison: Where should we store it?

tests/framework/mock/opae_fixtures.cpp
tswhison: I'm intending for it to replace most of our test fixtures which, today, contain boilerplate that is replicated over and over.

cmake/modules/OPAE.cmake
tswhison: done

r-rojo (approved):

PR #2555, opened 2022-04-14
build: add config cmake files

r-rojo:
* Modify cmake functions for creating targets to export the targets as
  opae-targets
* Change CMakeLists.txt to configure opae-config and opae-config-version
  files and install them to <prefix>/share/cmake
* Add template files opae-config.cmake.in and
  opae-config-version.cmake.in. These files will be configured from the
  build and when installed, external projects can use
  find_package(opae ...) to find opae includes/libraries.

TODO:
* add installed config files to RPM spec files
* scan codebase for other directives to add targets without using cmake functions in OPAECompiler.cmake


anandaravuri (approved):

tswhison (approved):
Neato!

michelleyho (approved):

PR #2567, opened 2022-04-20
opae.io: return pci_address in lower case

r-rojo:
Change pci_address parser function to always return pci address in lower
case becasue sysfs operates in lower case hex.


anandaravuri (approved):

rlee5-intel (approved):

tswhison (approved):
;)  quick draw McGraw..

PR #2570, opened 2022-04-21
tests: refactor opae-c tests for new fixtures

tswhison:
Refactor the opae-c unit tests to use the new opae_base_p<> /
opae_p<> template classes.


michelleyho (approved):

libraries/plugins/xfpga/error.c
```diff
@@ -199,12 +199,13 @@ const char *errors_exclude[NUM_ERRORS_EXCLUDE] = {
 };
 
 /* files that can be cleared by writing their value to them */
-#define NUM_ERRORS_CLEARABLE 6
+#define NUM_ERRORS_CLEARABLE 7
```
r-rojo, line 202: This change seems like it should be independent of this PR.

tswhison: It's required in order for the tests to pass.  Have a look at the opae-c errors API test.

r-rojo: Is this macro used by the tests or is it used in the xfpga (dfl) implementation of the errors API? Does this reflect what the driver (dfl) exports or what tests expect? Our mock sysfs structure should reflect what the driver exports. If the tests fail then it's either because our mock structure is incorrect or the API implementation doesn't truly reflect the interfaces exported. If this fixes the API then it should be independent of a "test refactor" PR.

r-rojo (approved):

PR #2577, opened 2022-04-26
pluginmgr: add API to register a plugin

r-rojo:
This adds a prototype in adapter.h and implements it in pluginmgr.c
This function is meant to register a plugin given only the plugin module
name. This allows client code to register a plugin and not have to rely
on built-in plugin loading and configuration.

### DIFF
--- a/libraries/libopae-c/pluginmgr.c
+++ b/libraries/libopae-c/pluginmgr.c
@@ -821,12 +821,12 @@ out_unlock:
 
 int opae_plugin_mgr_register_plugin(const char *name, const char *cfg)
 {
+	int lock_res = 0;
 	opae_api_adapter_table *adapter = opae_plugin_mgr_alloc_adapter(name);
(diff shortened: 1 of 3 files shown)


libraries/libopae-c/pluginmgr.c
```diff
@@ -821,29 +821,31 @@ int opae_plugin_mgr_for_each_adapter
...
-
 	int res = opae_plugin_mgr_configure_plugin(adapter, cfg);
 	if (res) {
 		opae_plugin_mgr_free_adapter(adapter);
 		OPAE_ERR("failed to configure plugin \"%s\"", name);
 		return res;
 	}
 
+	if (opae_mutex_lock(lock_res, &adapter_list_lock))
+		return res;
```
tswhison, line 839: Should it be
```c
return res;
```
or
```c
return lock_res;
```

tswhison (approved):

PR #2580, opened 2022-04-27
build: add OPAEPlugin.cmake

r-rojo:
* define cmake function called opae_add_shared_plugin that creates a
  shared library that registers a plugin as defined by its PLUGIN
  argument. To make the shared library it:
  * writes a file with C code implementing a library constructor that
    registers a plugin
  * adds any source files as listed after the SOURCE keyword
  * links this to opae-c-archive

The end result is a shared library that implements the OPAE API but has
the named plugin registered when the library loads.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -577,5 +577,6 @@ install(FILES
 
 install(FILES
     ${OPAE_SDK_SOURCE}/cmake/modules/OPAEGit.cmake
+    ${OPAE_SDK_SOURCE}/cmake/modules/OPAEPlugin.cmake
     DESTINATION lib/opae-${OPAE_VERSION}/modules
 )
--- a/cmake/modules/OPAEPlugin.cmake
+++ b/cmake/modules/OPAEPlugin.cmake
@@ -1,0 +1,51 @@
+#!/usr/bin/cmake -P
+## Copyright(c) 2022, Intel Corporation
+##
+## Redistribution  and  use  in source  and  binary  forms,  with  or  without
+## modification, are permitted provided that the following conditions are met:
+##
+## * Redistributions of  source code  must retain the  above copyright notice,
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -577,5 +577,6 @@ install(FILES
 
 install(FILES
     ${OPAE_SDK_SOURCE}/cmake/modules/OPAEGit.cmake
+    ${OPAE_SDK_SOURCE}/cmake/modules/OPAEPlugin.cmake
```
tswhison, line 580: Need to add to .rpm specs and .deb install files.

tswhison, line 580: Need to add to .rpm specs and .deb install files.

r-rojo: this should be covered by including the cmake installed directory in lib:
```spec
%{_prefix}/lib/opae-%{version}
```

tswhison (approved):

michael-adler:
Is this a solution to the libopae-c.so that wraps the ASE library automatically?

r-rojo:
@michael-adler , yes. By linking in libopae-c-archive.a, a library implements the API. This solution also gives a way to "hardcode" a plugin in the library constructor.

michael-adler:
Would it work as a shared library, e.g. libopae-c-ase.so, so that existing scripts continue to work?

michelleyho (approved):

r-rojo:
yes, the result of using this cmake function is a .so that you can link with to resolve symbols introduced by including <opae/fpga.h>.
I built a simple example that includes opae headers but links with libopae-c-ase. This example only calls fpgaEnumerate but yo can see here that by doing so it reached into ase plugin code.

image

Here is the backtrace:
image

michael-adler (approved):
Nice!

PR #2583, opened 2022-04-28
xfpga: sysobject: use minimum sysfs file size

tswhison:
An issue was identified with a sysfs file that contained a single 0.
In that case, the obj->max_size value was being set to 2, which caused
a subsequent fpgaObjectWrite64() call (with size > 2) to fail. Setting
the default size large enough to accomodate most sysfs entries prevents
this issue from occurring.

### DIFF
--- a/libraries/plugins/xfpga/sysfs.c
+++ b/libraries/plugins/xfpga/sysfs.c
@@ -2163,6 +2163,7 @@ out_free:
 	return res;
 }
 
+#define MIN_SYSOBJECT_FILESIZE 256
 #define MAX_SYSOBJECT_FILESIZE 0x40000
 static ssize_t find_eof(int fd)
 {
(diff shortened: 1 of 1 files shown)


anandaravuri (approved):

libraries/plugins/xfpga/sysfs.c
```diff
@@ -2451,8 +2454,8 @@ fpga_result make_sysfs_object(char *sysfspath, const char *name,
 	}
 	obj->handle = handle;
 	obj->type = FPGA_SYSFS_FILE;
-	obj->buffer = calloc(objstat.st_size, sizeof(uint8_t));
-	obj->max_size = objstat.st_size;
+	obj->max_size = MIN_SYSOBJECT_FILESIZE;
```
r-rojo, line 2457: what if the object.st_size is bigger than MIN_SYSOBJECT_FILESIZE?

tswhison: That was the nature of the bug.  The objstat.st_size for a file containing "0" was 2.  But, the subsequent write size was for a string that contained "0x400000000" (which is greater than 2, and so failed the length check).  Should it be set to max(objstat.st_size, MIN_SYSOBJECT_FILESIZE)?

r-rojo: That's what I was expecting. max of the file size as reported by stat or the minimum.

r-rojo (approved):

PR #2585, opened 2022-04-29
opae.io: add 32-bit support for actions

r-rojo:
* Add --access-mode CLI argument
  * Setting access mode sets utils.ACCESS_MODE accordingly
* Modify opae.io actions to read/write registers in 32 bit mode

### DIFF
--- a/binaries/opae.io/opae/io/utils.py
+++ b/binaries/opae.io/opae/io/utils.py
@@ -59,6 +59,7 @@ VENDOR_DEVICE_PATTERN = r'(?P<vendor>[\da-f]{0,4}):(?P<device>[\da-f]{0,4})'
 VENDOR_DEVICE_RE = re.compile(VENDOR_DEVICE_PATTERN, re.IGNORECASE)
 
 PICKLE_FILE = '/var/lib/opae/opae.io.pickle'
+ACCESS_MODE = 64
 
 
 class pcicfg(Enum):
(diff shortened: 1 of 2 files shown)


binaries/opae.io/opae/io/utils.py
```diff
@@ -259,12 +261,21 @@ def define(cls,
     def commit(self, value=None):
         if value is not None:
             self.value = value
-        self.wr(self.offset, self.value)
+        if ACCESS_MODE == 32 and self.width == 64:
+            self.region.write32(self.offset, self._upper_mask & value)
```
tswhison, line 265: Shouldn't self._upper_mask be reversed?  I think you are writing the low bits here.

r-rojo: you're right! fixed!

binaries/opae.io/opae/io/utils.py
```diff
@@ -259,12 +261,21 @@ def define(cls,
...
+        else:
+            self.wr(self.offset, self.value)
 
     def update(self):
         if self.region is None:
             raise OSError(os.EX_OSERR, 'no region open')
-        self.value = self.rd(self.offset)
+        if ACCESS_MODE == 32 and self.width == 64:
+            lo = self.region.read32(self.offset)
+            self.value = self.reaion.read32(self.offset+4) << 32
```
tswhison, line 275: s,reaion,region,

tswhison (approved):

PR #2592, opened 2022-05-09
opae.io: refactor `ls` action

r-rojo:
* add pci.py file for pci related functionality
  * add device class that encapsulates some sysfs related functions for
    discovering/controlling pci devices.
  * implement `ls` function that lists pci devices and can filter out
    given a filter function.
  * move pci related operations from utils.py into pci.py
* refactor utils to
  * store known fpga devices in a dictionary keyed off a pci_id named
    tuple that names vendor,device, as well as
    subsystem_vendor, and subsystem_device.
  * implement lsfpga to create a custom filter function that checks for
    known devices as defined in the named tuple dictionary.
* change pymain.h to reference data structures in pci.py


binaries/opae.io/opae/io/utils.py
```diff
@@ -412,18 +381,18 @@ def register(self, offset, name='csr', bits=[('value', c_uint64, 64)]):
...
-                (0x1c2c, 0x1000): "Silicom SmartNIC N5010",
-                (0x1c2c, 0x1001): "Silicom SmartNIC N5011"}
+fpga_devices = {pci.pci_id(0x8086, 0x09c4): "Intel PAC A10 GX",
+                pci.pci_id(0x8086, 0x09c5): "Intel PAC A10 GX VF",
+                pci.pci_id(0x8086, 0x0b2b): "Intel PAC D5005",
+                pci.pci_id(0x8086, 0x0b2c): "Intel PAC D5005 VF",
+                pci.pci_id(0x8086, 0x0b30): "Intel PAC N3000",
+                pci.pci_id(0x8086, 0x0b31): "Intel PAC N3000 VF",
+                pci.pci_id(0x8086, 0xaf00): "Intel N6000 ADP",
+                pci.pci_id(0x8086, 0xaf01): "Intel N6000 ADP VF",
```
fpgamatt, line 391: I don't think 0xaf00/0xaf01 should be in this table.  Those id were accidentally used for a bit during IOFS-EA, but have been retired.

binaries/opae.io/opae/io/utils.py
```diff
@@ -412,18 +381,18 @@ def register(self, offset, name='csr', bits=[('value', c_uint64, 64)]):
...
+                pci.pci_id(0x8086, 0xbcce): "Intel N6000 ADP",
+                pci.pci_id(0x8086, 0xbccf): "Intel N6000 ADP VF",
```
fpgamatt, line 393: bcce/bccf is used by n6000, c61000, and d5005 at some point.  You could just remove N6000 from the string, but I think this table should have the full 4-tupple of VID:PID:SSVID:SID.

binaries/opae.io/opae/io/pci.py
```diff
@@ -0,0 +1,167 @@
...
+    0x0B: "Processors ",
+    0x0C: "Serial bus controllers ",
+    0x0D: "Wireless controller",
+    0x0E: "Intelligent I/O controllers",
+    0x0F: "Satellite communication controllers",
+    0x10: "Encryption/Decryption controllers",
+    0x11: "Data acquisition and signal processing controllers",
+    0x12: "Processing accelerators",
+    0x13: "Non-Essential Instrumentation"
+}
```
fpgamatt, line 120: It seems like this would be defined somewhere like pciutils, but then again I'm extremely lazy (in the perl sense).

binaries/opae.io/opae/io/utils.py
r-rojo: will remove

binaries/opae.io/opae/io/utils.py
```diff
@@ -412,18 +381,18 @@ def register(self, offset, name='csr', bits=[('value', c_uint64, 64)]):
...
-                (0x8086, 0xbccf): "Intel N6000 ADP VF",
-                (0x1c2c, 0x1000): "Silicom SmartNIC N5010",
-                (0x1c2c, 0x1001): "Silicom SmartNIC N5011"}
+fpga_devices = {pci.pci_id(0x8086, 0x09c4): "Intel PAC A10 GX",
+                pci.pci_id(0x8086, 0x09c5): "Intel PAC A10 GX VF",
+                pci.pci_id(0x8086, 0x0b2b): "Intel PAC D5005",
+                pci.pci_id(0x8086, 0x0b2c): "Intel PAC D5005 VF",
+                pci.pci_id(0x8086, 0x0b30): "Intel PAC N3000",
+                pci.pci_id(0x8086, 0x0b31): "Intel PAC N3000 VF",
+                pci.pci_id(0x8086, 0xbcce, 0x8086, 0x1770): "Intel N6000 ADP",
```
anandaravuri, line 390: add 0x8086 0xbcce 0x8086 0x17d4 Intel Acceleration Development Platform C6100

0x8086 0xbcce 0x8086 0x138d   Intel Acceleration Development Platform  D5005
0x8086 0xbccf 0x8086 0x138e    Intel Acceleration Development Platform D5005 VF

binaries/opae.io/opae/io/pci.py
r-rojo: Perhaps but I don't want to have another dependency if I can avoid it.

binaries/opae.io/opae/io/utils.py
```diff
@@ -448,54 +423,63 @@ def get_conf():
...
-            continue
-        if device and int(did, 16) != int(device, 16):
-            continue
-        id_tuple = (int(vid, 0), int(did, 0))
-        if not device_ids or id_tuple in device_ids:
-            if not driver or (driver and drv and driver == drv):
-                yield (pcidir.parts[-1], vid, did, device_ids.get(id_tuple, ''), drv)
+
+    def filter_fn(d: pci.device):
+        for k,v in kwargs.items():
```
tswhison, line 455: Is kwargs a parameter here?  Should the function signature account for it?
```python
def filter_fn(d: pci.device, **kwargs):
```

r-rojo: no, `filter_fn` is a closure and defined inside of the outer function which means its scope includes the outer function.

tswhison: Perhaps a comment that clarifies what's going on?

tswhison (approved):

PR #2604, opened 2022-05-18
pacsign: enable on CentOS 8 Stream / fix bug

tswhison:
Allow the version check to perform exact match or match based on
regex pattern.


rchriste1 (approved):
This works on CentOS 8.5

anandaravuri (approved):

PR #2616, opened 2022-06-03
mmio: ensure x86 arch for AVX512 mmio

tswhison:
Enables building on ARM/other platforms that don't support these
instructions.

### DIFF
--- a/libraries/plugins/vfio/opae_vfio.c
+++ b/libraries/plugins/vfio/opae_vfio.c
@@ -677,12 +677,14 @@ fpga_result vfio_fpgaOpen(fpga_token token, fpga_handle *handle, int flags)
 	_handle->mmio_size = size;
 
 	_handle->flags = 0;
+#if defined(__i386__) || defined(__x86_64__) || defined(__ia64__)
 #if GCC_VERSION >= 40900
 	__builtin_cpu_init();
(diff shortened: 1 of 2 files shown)


anandaravuri (approved):

libraries/plugins/xfpga/mmio.c
```diff
@@ -319,13 +319,21 @@ fpga_result __XFPGA_API__ xfpga_fpgaReadMMIO64(fpga_handle handle,
 	return result;
 }
 
+#if (defined(__i386__) || defined(__x86_64__) || defined(__ia64__)) && GCC_VERSION >= 40900
 static inline void copy512(const void *src, void *dst)
```
r-rojo, line 323: I'm wondering if we should change the signature of this function to return either 0 on success and non-zero on failure.
Or maybe put some logging information to warn when the dummy version is called. I would like to have some sort of indication that calling that version is essentially a no-op.

r-rojo (approved):

libraries/plugins/xfpga/mmio.c
tswhison: We'll get FPGA_NOT_SUPPORTED in the case of non-Intel architecture, before copy512() is actually called.

PR #2623, opened 2022-06-13
opae.io: save progress on uservices demo

tswhison:

### DIFF
--- a/binaries/opae.io/opae/io/run.py
+++ b/binaries/opae.io/opae/io/run.py
@@ -1,5 +1,5 @@
 import json
-from flask import Flask, request
+from flask import Flask, request, abort
 from opae import fpga
 
 DEFAULT_SRV_PORT = 8080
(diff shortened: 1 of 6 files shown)


binaries/opae.io/opae/io/run.py
```diff
@@ -46,32 +48,44 @@ def token_info(t: fpga.token):
...
     tokens = fpga.enumerate()
     return json.dumps([token_info(t) for t in tokens])
 
 
+@app.route('/api/v0/fpga/<object_id>/pr', methods=['POST'])
+def partial_reconfig(object_id):
+    file = request.files['gbs']
+    if file:
+        print(f'received {file.filename} to PR object_id={object_id}')
+        t = fpga.enumerate(object_id=int(object_id, 16))
```
r-rojo, line 71: Should we check if t is empty here?

r-rojo (approved):

binaries/opae.io/opae/io/run.py
tswhison: ok

PR #2630, opened 2022-06-21
opae-sdk: add board c6100 module

anandaravuri:
- prints board specific info  example bmc info , image info and events.
- fpgainfo tool dynamically loads board c6100 module if finds device id 0xbccc and sub device id 0x17d4


tswhison:
Unit tests for new board module?

tswhison:
Thanks for adding the test cases.  I checked the coverage, and it is still down 2 points with your additions; however, your board_c6100 module is only at 20.4% (likewise, board_n6000 is only at 21%).  Could you please add some more test cases to bring the number up?

fpgamatt (approved):

michelleyho (approved):

PR #2646, opened 2022-07-06
changing packages in create

gwencasey96:
replaced catchall python package on line 61 with 3 individual packages. Original breaks Fedora Rawhide, proposed change

### DIFF
--- a/packaging/opae/rpm/create
+++ b/packaging/opae/rpm/create
@@ -58,7 +58,9 @@ declare -a PKG_DEPS=(\
 'gcc-c++' \
 'json-c-devel' \
 'libuuid-devel' \
-'python3*-devel' \
+'libedit-devel' \
+'python3-devel' \
+'systemd-devel' \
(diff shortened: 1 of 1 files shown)


anandaravuri (approved):

packaging/opae/rpm/create
```diff
@@ -58,7 +58,9 @@ declare -a PKG_DEPS=(\
 'gcc-c++' \
 'json-c-devel' \
 'libuuid-devel' \
-'python3*-devel' \
+'libedit-devel' \
```
tswhison, line 61: Hi @gwencasey96](https://github.com/gwencasey96), we should move libedit-devel down to the Fedora-specific list (lines 71-81), since libedit-devel is not required for the minimized rhel9 build, currently.  I'll also try a regression in my environment to make sure things are ok.

tswhison:
Closing this, as CI is hung (fork). See #2656 instead.

PR #2656, opened 2022-07-15
packaging: remove python3*-devel wildcard

tswhison:
It was reported that the wildcard doesn't work with Fedora 37. Move
to an explicit check on the name instead.

Add libedit-devel and systemd-devel to the list of prerequisites for
the full package build.


tswhison:
Submitting this instead of #2646, as that one was submitted from a fork, which causes our CI to hang.

gwencasey96:
I thought you mentioned the python3*-devel was needed for rhel 8? Sorry I didn't get a chance to confirm. Will Rhel 8 and 9 be added to this CI? Also I don't believe I had direct access so I submitted from a fork; please let me know if there's a preferred way for me to communicate changes.

tswhison:
The addition of python36-devel and python38-devel as explicit options satisfies the RHEL builds.  Let me see about adding you and Tom as developers.

tswhison:
@gwencasey96 I updated your account to have write access.  Hopefully that will solve it.

packaging/opae/rpm/create
```diff
@@ -50,6 +50,27 @@ declare -r RPM_SPEC="opae.spec.${WHICH_RPM}"
 
 # Check for prerequisite packages.
 
+declare -ra PYTHON_PKG_DEPS=(\
+'python3-devel' \
+'python36-devel' \
+'python38-devel'\
```
anandaravuri, line 56: missing 3.9,3.10, and 3.11 ?
Fedora37 python3.11 version
fedora36 python3.10 version

default pyhon packages changes every Fedora release
can we use python3*-devel ?

tswhison: We can't use python3*-devel.  Per Gwen, it breaks on Fedora37.  What are the exact package names in Fedora36 and 37?

anandaravuri: Fedora37  cat /etc/redhat-release
Fedora release 37 (Rawhide)
python
Python 3.11.0b3 (main, Jun 24 2022, 00:00:00) [GCC 12.1.1 20220507 (Red Hat 12.1.1-1)] on linux

Fedora35  cat /etc/redhat-release
Fedora release 35 (Thirty Five)
 python
Python 3.10.4 (main, Mar 25 2022, 00:00:00) [GCC 11.2.1 20220127 (Red Hat 11.2.1-9)] on linux


Python in Fedora 36, Python 3.10,

anandaravuri (approved):

PR #2659, opened 2022-07-18
opae-sdk: add support installed  pybind11 packages

anandaravuri:
-Use  installed  pybind11 for building OPAE-SDK  If user installed pybind11-devel package on system.
-Clone pybind11 devel  packages If user is not installed pybind11-devel


binaries/fpgadiag/CMakeLists.txt
```diff
@@ -89,7 +89,10 @@ add_fpgadiag_app(nlb3 src/nlb3_main.cpp toolfpgadiagapps)
 add_fpgadiag_app(nlb7 src/nlb7_main.cpp toolfpgadiagapps)
 
 if (OPAE_WITH_PYBIND11)
-    find_package(pybind11) 
+    if (POLICY CMP0074)
+      cmake_policy(SET CMP0074 NEW)
+    endif()
+    find_package(pybind11)
```
tswhison, line 95: This doesn't fix the warning, it merely masks it.  Please address the cause of the warning.
It might be helpful to run the build with $ make VERBOSE=1 to see the include paths being used.

tswhison (changes requested):

PR #2660, opened 2022-07-19
(cherry pick) unit test: fix hardware test failures (#2658)

tswhison:
On Ubuntu platforms, calling dlopen() caused the opae_init() library
constructor to be called a 2nd time (while the first call was in
progress). Use pluginmgr.c's initialized variable to avoid the
re-entrant call to opae_init().

dfl_n6000_sku0 hardware has only 1 MMIO region, similar to sku1.


rlee5-intel (approved):

anandaravuri (approved):

rlee5-intel:
Merging as admin

PR #2667, opened 2022-07-25
opae.io: add --safe option to walk command

tswhison:
Verifies that the next offset field when walking the DFL is aligned
per the active ACCESS_MODE. If it is not properly aligned, raise
an exception without dereferencing the offset.


fpgamatt (approved):
My only nit would be to have a "subject" line in the commit message like what is done in kernel code.  I think we should "up our game" with opae commits.

anandaravuri (approved):

r-rojo (approved):

PR #2685, opened 2022-08-08
packaging: (Cherry-pick) remove python3*-devel wildcard (#2656)

anandaravuri:
It was reported that the wildcard doesn't work with Fedora 37. Move
to an explicit check on the name instead.

Add libedit-devel and systemd-devel to the list of prerequisites for
the full package build.


tswhison (approved):
LGTM

PR #2686, opened 2022-08-08
opae.io: (Cherry-pick) add --safe option to walk command (#2667)

anandaravuri:
Verifies that the next offset field when walking the DFL is aligned
per the active ACCESS_MODE. If it is not properly aligned, raise
an exception without dereferencing the offset.


tswhison (approved):
LGTM

PR #2697, opened 2022-08-19
Don't use the deprecated and removed os.errno

chipb-intel:
This API was apparently undocumented, and it was removed altogether in
the 3.7 timeframe: https://bugs.python.org/issue33666

### DIFF
--- a/platforms/scripts/afu_synth_setup
+++ b/platforms/scripts/afu_synth_setup
@@ -131,7 +131,7 @@ def commands_list(cmd, cwd=None, stdout=None):
     try:
         subprocess.check_call(cmd, cwd=cwd, stdout=stdout, env=getCmdEnv())
     except OSError as e:
-        if e.errno == os.errno.ENOENT:
+        if e.errno == errno.ENOENT:
             msg = cmd[0] + " not found on PATH!"
             errorExit(msg)
         else:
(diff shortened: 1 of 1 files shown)


chipb-intel:
@michael-adler FYI -- here's a fix for part of the CI surprise failure.

anandaravuri (dismissed):

chipb-intel:
@anandaravuri @tswhison Any ideas who owns these required "analyze (3.{6..8})" checks which aren't running?

tswhison:
@chipb-intel We've noticed this problem is unique to pull requests that are made from forks.  We are not sure why this is the case.

tswhison (dismissed):

chipb-intel:
Bah. I was blind to the fact that these Python scripts lack the `*.py` extension, so this PR wasn't triggering the "real" Python static checks at all. One more fix with #2708 should get this PR correctly launching *all* its checks.

michael-adler:
Maybe the tests could be improved by looking for the shebang? Forcing a script's extension to indicate it's implementation language is a strange convention. We don't impose this on normal binaries.

chipb-intel:
> Maybe the tests could be improved by looking for the shebang? Forcing a script's extension to indicate it's implementation language is a strange convention. We don't impose this on normal binaries.

While I completely agree, I also don't want to drag this simple two-liner fix into a goose chase of fixing *all* the static checks right at this moment.

chipb-intel:
@OPAE/ase Okay, it looks like the checks are all sorted finally. Could I get one final thumbs-up?

michael-adler (approved):
I agree -- I didn't mean to suggest holding this up.

tswhison (approved):

PR #2698, opened 2022-08-22
Ahjorth/optional pcie address in eth tools

Anders5-dk:
Add support for optional pcie_address in binary/hssi/ethernet/* tools.
Add command rejection in binaries/fpgadiag/opae/diag/* tools if more than one board is selected.


Anders5-dk:
Please note that I have only changed the hssistats and fpgastats commands yet. If the changes are approved, they shall be replicated to all the other commands in the two folders.
In that case I also recommend changing the return data type from class FpgaFinder.find_eth_group() from dict to list since the data is used that way.

anandaravuri (approved):

PR #2706, opened 2022-08-26
Add a CSR field indicating PCIe atomics support to avoid problems on older machines

michael-adler:
Remains backward compatible with older systems. Unfortunately, this means
assuming that older machines do work with atomics. On future OFS systems,
"host_exerciser --testall true" will skip atomic tests if unsupported.


tswhison (approved):
LGTM

anandaravuri (approved):

PR #2707, opened 2022-08-26
ci: trigger Python static checks on PRs

chipb-intel:


michelleyho (dismissed):

chipb-intel:
@tswhison I *think* this will fix the job triggering in my other PR? Seems to be what's different in those jobs, but I'm still getting more familiar with the Github Actions way of doing things.

tswhison (approved):
This is great.  Thanks a bunch!

PR #2709, opened 2022-08-29
opae-sdk:  add support to installed pybind11 to build python bindings

anandaravuri:
-if OPAE SW build flow check pybind11-devel installed on system, OPAE build  uses installed pybind11 packages.


tswhison (commented):
The find_package(pybind11) calls seem redundant.  Can't we just call it once in the main CMakeLists.txt?

tswhison (approved):
Cool!

PR #2716, opened 2022-09-03
cmake: fix missing include directories in opae_add_static_library()

pcolberg:
This synchronizes the target include directories with those of other
targets, which resolves a build failure when libjson-c is installed
in a non-standard path and configured via CMAKE_PREFIX_PATH.

Signed-off-by: Peter Colberg <peter.colberg@intel.com>


tswhison (approved):
Nice catch, thanks!

PR #2721, opened 2022-09-08
opaevfio, opae.io: bind vfio-pci driver to single PF instead of all PFs

pcolberg:
On Linux kernel >= 3.15, use driver_override to bind the
vfio-pci driver, which ensures the driver binds only to the
given physical function (PF). On older kernels, the driver
will instead bind to all PFs of the corresponding device.

See also https://github.com/DPDK/dpdk/commit/2fc3502935700243d9a6d903166e6fd11e429843


tswhison (commented):
In the pull description and in the code comments you mention PF.  Is there anything precluding this change from also binding a VF device (was this tested)?  This is a common use case, eg the host exercisers are VF devices.

tswhison (approved):
Thanks!

PR #2740, opened 2022-09-27
scripts: add opae-clean.py

tswhison:
Add utility script opae-clean.py to aid in scrubbing systems when a local install from source is present. A common issue encountered with local installs from source is that cleaning a system of installed artifacts is nearly impossible once the source/build tree has been removed. Using opae-clean.py, particularly with the --deep flag can help remedy this situation.

The script can be run from any directory, so long as the relative positions to the project files are unchanged. For RedHat-based installs, the script scans opae.spec.fedora in the root of the git tree. For Debian-based installs, packaging/opae/deb/*.(install|dirs) are used.

A typical usage is to run the script without parameters, or with the --deep parameter. The script will report what files and directories it thinks are likely candidates to scrub; but at this point, nothing will be removed. After you have carefully examined the list of files and directories reported, the script can be re-executed with the -D switch to remove the target files and directories (given that sufficient privilege is held to do so).


tswhison:
The use case is quite clear.  It happens frequently on our lab systems.  Someone will do
```bash
$ cmake ..
$ make
$ sudo make install
```
without setting CMAKE_INSTALL_PREFIX, and install binaries into /usr/local, irrespective of whether someone has already installed RPMs to /usr.  This case is particularly problematic because of the load order of shared module libraries and python modules.

The problem is that, when another user comes along, they are not aware of the prior user's build tree.  And they have no way of removing an installation that was made from $ sudo make install.  The system is effectively broken with no way for them to recover.

Viewing opae.spec.fedora, our debian packaging in packaging/opae/deb, or running the script on a system as I described would have illustrated the difficulty being addressed.  There is no set $PREFIX than can be applied.  We have files landing not only in /usr, but also /etc/sysconfig, /etc/opae/, and others.

While what you are describing behind using a common $PREFIX is theoretically possible, I'm harnessed to the reality today that this is a problem that has manifested numerous times in house.  In lieu of any such capability, which would require a deep refactor of the entire build system, we are where we are with this.

You didn't speak to whether the script itself has merit.  I would appreciate your review of the python script content.

anandaravuri (approved):

PR #2741, opened 2022-09-27
user clk: read user clock dfh revision from uio driver

anandaravuri:

### DIFF
--- a/libraries/plugins/xfpga/usrclk/fpga_user_clk.c
+++ b/libraries/plugins/xfpga/usrclk/fpga_user_clk.c
@@ -136,6 +136,23 @@
 
 #define USRCLK_FEATURE_ID             0x14
 
+ // DFHv0
+struct dfh {
+	union {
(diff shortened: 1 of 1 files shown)


tswhison (approved):
LGTM.  What sort of testing was done?

michael-adler (approved):
Tested on Agilex and S10, confirmed both are detected and max. frequency matches the target. Loaded a GBS at 635 MHz on Agilex and the proper frequency was set.

Would it make sense to extend the CSRs to indicate properties of the clock so that you don't have to infer from the HW family?

michael-adler (approved):

michael-adler:
@tswhison I think this is waiting on you as a code owner to approve. When I did a rebase in the GUI here it dropped your approval.

tswhison:
Thanks @michael-adler.  I think that testing is still ongoing with this.

tswhison (approved):

PR #2742, opened 2022-09-28
config: change 1c2c:1001 to N5011 PF

tswhison:
The previous implementation erroneously treated 1001 as the n5010 VF. Also, removes default fpga boot sequence treatment from n5010, based on feedback from Silicom.


anandaravuri (dismissed):

Anders5-dk (changes requested):
default fpga boot sequence treatment shall also be removed from n5011

Anders5-dk (approved):

anandaravuri (approved):

PR #2747, opened 2022-09-30
ase: add hook to enable ASE with new config format

tswhison:
(re)-enables ASE as of the new config file format. Hook to allow enumerating the ASE device when WITH_ASE is set in the environment.


michael-adler (approved):

tswhison:
Override "api" approval as opae-sim is broken without this change.  Change was confirmed to work in house.

PR #2751, opened 2022-10-03
sim: changes to support ASE for OpenCL MMD.

tswhison:


tswhison:
Supports https://github.com/OPAE/opae-sim/pull/32

libraries/libopae-c/pluginmgr.c
```diff
@@ -281,14 +281,22 @@ STATIC int opae_plugin_mgr_detect_platforms(bool with_ase)
 
 	if (with_ase) {
 		opae_pci_device ase_pf = {
-			.name = "ase",
+			.name = "ase_pf",
```
michael-adler, line 284: We have existing tests that look for a single device and then check the vendor_id/device_id for 0x8086/0xa5e. OPAE is, unfortunately, returning device_id 0xa5f first. The simulation works, but the SW thinks it isn't ASE.

Would this new code still work if you made the device_id always 0xa5e and varied subsystem-device_id instead?

tswhison: We can give it a try.  I can't think of a reason OTTOMH why it wouldn't.  It will involve a change here and a change in opae_ase.cfg.  Let me know.

michael-adler (dismissed):
The updated version works for me and existing SW properly detects ASE. Thank you.

I can't evaluate whether the update will work for the MMD.

pajgaonk (approved):

include/opae/types_enum.h
```diff
@@ -126,9 +126,12 @@ typedef enum {
 	/** FPGA_IFC_VFIO indicates that the plugin interface is the
 	 * vfio-pci driver. */
 	FPGA_IFC_VFIO,
-	/** FPGA_IFC_SIM indicates that the plugin interface is the
-	 * AFU Simulation Environment. */
-	FPGA_IFC_SIM
+	/** FPGA_IFC_SIM_DFL indicates that the plugin interface is the
+	 * AFU Simulation Environment simulating DFL drivers. */
+	FPGA_IFC_SIM_DFL,
```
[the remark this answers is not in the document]
tswhison: Yes, this is true.  ASE would be the only plugin to implement FPGA_IFC_SIM and the replacement FPGA_IFC_SIM_DFL/FPGA_IFC_SIM_VFIO.  So, we just need to coordinate the checkin between the two repos.

PR #2758, opened 2022-10-07
deb packaging: add support for Ubuntu 20.04

tswhison:
Remove explicit mentions of .so deps from the control file
and let ${shlibs:Depends} do its thing.

Verified using Ubuntu 22.04 LTS and Ubuntu 20.04 LTS.

Co-authored-by: Peter Colberg <Peter.Colberg@intel.com>

### DIFF
--- a/packaging/opae/deb/control
+++ b/packaging/opae/deb/control
@@ -20,8 +20,7 @@ Package: opae
 Section: libs
 Priority: optional
 Architecture: any
-Depends: libuuid1, libjson-c5,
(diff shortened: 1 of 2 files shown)


tswhison:
I see.  Thanks for the heads-up.  Would setting a reasonable default to DIST_KEY when lsb_release is not detected be an appropriate safeguard?

bmuddebi (approved):
Checked out this branch on Ubuntu 20.04,  Able create 20.04 deb packages and installed , works perfectly fine

anandaravuri (approved):

PR #2763, opened 2022-10-20
Rchriste/n5014 2

rchriste1:
These two commits add support for Silicom N5013 and N5014 cards. It has been tested with the real cards.


opae.cfg
```diff
@@ -268,6 +268,126 @@
...
+            "configuration": {
+              "cool-down": 30,
+              "get-aer":     [ "setpci -s %s ECAP_AER+0x08.L",
+                               "setpci -s %s ECAP_AER+0x14.L" ],
+              "disable-aer": [ "setpci -s %s ECAP_AER+0x08.L=0xffffffff",
+                               "setpci -s %s ECAP_AER+0x14.L=0xffffffff" ],
+              "set-aer":     [ "setpci -s %s ECAP_AER+0x08.L=0x%08x",
+                               "setpci -s %s ECAP_AER+0x14.L=0x%08x" ],
+              "sensor-overrides": [],
+              "monitor-seu": true
```
tswhison, line 311: Does this device support Single Event Upset via the FME error registers?  Just checking to be sure, because as of Arrow Creek, SEU is detected via the flight data recorder.

opae.cfg
```diff
@@ -268,6 +268,126 @@
...
```
tswhison, line 371: Is SEU supported via the FME error registers?  If not, then let's set this instance and the above instance to false.

tswhison (commented):
Thanks for your submission.  The configuration file code looks great.  I would like Ananda to have a chance to review the fpgainfo updates.

opae.cfg
rchriste1: The N501x cards do not support the flight data recorder. But I checked with our FPGA team and it seemd the SEU bit is not implemented either in the old OFS version used. So I have changed the monitoring to false.

I had also made a json syntax error which I have fixed. So I did a forced push.

tswhison (approved):

anandaravuri (approved):

PR #2774, opened 2022-11-14
build: add CMAKE_INSTALL_PREFIX length check

tswhison:
Linux mandates a limit on the length of a command that may appear in
the "shebang" line of a script. In our internal environments, this
limit has been exceeded, causing ill-behaved builds.

Fail the build if the length of CMAKE_INSTALL_PREFIX exceeds the
127 character limit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,9 +83,11 @@ endif(OPAE_MINIMAL_BUILD)
 
 string(LENGTH ${CMAKE_INSTALL_PREFIX} INSTALL_PREFIX_LEN)
 if (${INSTALL_PREFIX_LEN} GREATER 127)
-    message(WARNING "The length of CMAKE_INSTALL_PREFIX (${INSTALL_PREFIX_LEN}) has exceeded"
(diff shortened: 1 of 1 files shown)


tswhison:
Sure, what did you have in mind?

CMakeLists.txt
```diff
@@ -77,6 +77,17 @@ if (OPAE_MINIMAL_BUILD)
     set(CMAKE_INSTALL_PREFIX "/usr" CACHE STRING "" FORCE)
 endif(OPAE_MINIMAL_BUILD)
 
+############################################################################
+## CMAKE_INSTALL_PREFIX check ##############################################
+############################################################################
+
+string(LENGTH ${CMAKE_INSTALL_PREFIX} INSTALL_PREFIX_LEN)
+if (${INSTALL_PREFIX_LEN} GREATER 127)
```
michael-adler, line 85: Should this be an error instead of just a message? We already know the commands will fail mysteriously. I'm worried that just a message will be lost in the noise of everything else CMake is spitting out.

michael-adler (approved):
Thank you!

PR #2775, opened 2022-11-16
CODEOWNERS: transition from OPAE to OFS teams

pcolberg:
Signed-off-by: Peter Colberg <peter.colberg@intel.com>


tswhison (approved):
Thanks!

PR #2783, opened 2022-12-08
ci: update python versions

anandaravuri:
- Remove Python 3.6, 3.7  and add 3.10,3.11 to versions to use for static analysis.

Signed-off-by: lab <ananda.ravuri@intel.com>

### DIFF
--- a/.github/workflows/no-python-analysis.yml
+++ b/.github/workflows/no-python-analysis.yml
@@ -10,6 +10,6 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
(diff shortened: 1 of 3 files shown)


anandaravuri:
> Thanks @anandaravuri!
> 
> What is the reasoning behind supporting these particular versions? Is it simply to support the latest 4 major versions at any point?
> 
> According to the [Status of Python Versions](https://devguide.python.org/versions/), 3.7, 3.8, and 3.9 are currently receiving security updates, while 3.10 and 3.11 are also receiving bug fixes.

@pcolberg 
Fedora37/38 python default version  are 3.10 and 3.11 , make sure OPAE SW  builds on latest fedora python versions.

tswhison (approved):

PR #2786, opened 2022-12-14
rename ofs.uio.py to ofs_uio.py and fix sig fault

anandaravuri:
- install opae packages, ofs.uo runtime python search in wrong dist-packages folders python interprets uio.ofs.uio:main  as uio/ofs/uio.py :main in setup.py rename ofs.uio.py to ofs_uio.py python interprets uio.ofs_uio:main as uio/ofs_uio.py :main in setup.py

  - user calls pyopae_uio  close() without open() cause sigfault. memset struct opae_uio to zero in pyopaeui.h


binaries/ofs.uio/setup.py
```diff
@@ -31,7 +31,7 @@
     packages=find_namespace_packages(include=['*']),
     entry_points={
         'console_scripts': [
-            'ofs.uio = uio.ofs.uio:main',
+            'ofs.uio = uio.ofs_uio:main',
```
tswhison, line 34: Good catch!

tswhison (approved):

PR #2798, opened 2022-12-22
ci: update to lycheeverse/lychee-action@v1

tswhison:
Use lychee instead of peter-evans/link-checker, which is deprecated.

### DIFF
--- a/.github/workflows/ccpp-tests.yml
+++ b/.github/workflows/ccpp-tests.yml
@@ -112,12 +112,13 @@ jobs:
       with:
         name: docs
         path: mybuild_docs/sphinx/html
-    - name: Link Checker
-      uses: peter-evans/link-checker@v1
+    - name: (lychee) Link Checker
+      uses: lycheeverse/lychee-action@v1
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ccpp-tests.yml
```diff
@@ -112,12 +112,13 @@ jobs:
       with:
         name: docs
         path: mybuild_docs/sphinx/html
-    - name: Link Checker
-      uses: peter-evans/link-checker@v1
+    - name: (lychee) Link Checker
+      uses: lycheeverse/lychee-action@v1
```
anandaravuri, line 116: https://github.com/lycheeverse/lychee-action
use latest version lycheeverse/lychee-action@v1.5.4

anandaravuri (approved):

.github/workflows/ccpp-tests.yml
tswhison: https://github.com/lycheeverse/lychee-action

"If you always want to use the latest features but avoid breaking changes, you can replace the version with
lycheeverse/lychee-action@v1"

PR #2809, opened 2023-01-07
cfg search: log error when "configs" not detected

tswhison:
* Search for ASE config in both /usr/share/opae and /usr/local/share/opae to avoid conflicts caused by mismatched CMAKE_INSTALL_PREFIX.
* OPAE_ERR() when parsing the config JSON fails to find the "configs" array in order to help detect the case when an old configuration is encountered.
* OPAE_DBG() when the ASE configuration file is found. This was already being done for the non-ASE case.

### DIFF
--- a/libraries/libopae-c/cfg-file.c
+++ b/libraries/libopae-c/cfg-file.c
@@ -297,15 +297,15 @@ STATIC libopae_config_data default_libopae_config_table[] = {
 };
 
 libopae_config_data *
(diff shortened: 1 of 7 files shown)


anandaravuri (dismissed):

anandaravuri (approved):

michael-adler (approved):
Thank you for adding better error checking. It wasn't failing in my env, so I didn't check the new tests. I did confirm that ASE still works.

PR #2810, opened 2023-01-09
fix rpmlint non-executable-script errors

anandaravuri:
Error: opae-devel.x86_64: E: non-executable-script /usr/lib/opae-2.2.0/modules/OPAEGit.cmake 644 /usr/bin/cmake -P Add execute permissions to  .cmake

### DIFF
--- a/opae.spec.rhel
+++ b/opae.spec.rhel
@@ -83,29 +83,6 @@ cp LICENSE %{buildroot}%{_datadir}/opae/LICENSE
 cp COPYING %{buildroot}%{_datadir}/opae/COPYING
 
 mkdir -p %{buildroot}%{_usr}/src/opae/cmake/modules
-for s in \
-    FindCLI11.cmake \
-    FindCap.cmake \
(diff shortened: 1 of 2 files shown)


tswhison (commented):
Could you please implement the same change in opae.spec.rhel for consistency?

opae.spec.rhel
```diff
@@ -159,6 +145,10 @@ for file in %{buildroot}%{python3_sitelib}/ethernet/{hssicommon,hssiloopback,hss
    chmod a+x $file
 done
 
+# packager
+for file in %{buildroot}%{python3_sitelib}/packager/tools/{afu_json_mgr,packager}.py; do
+   chmod a+x $file
+done
```
tswhison, line 151: I think we already have this in the fedora spec, is that right?

tswhison (approved):

opae.spec.rhel
anandaravuri: yes,  packager is already there in fedora spec ,

PR #2813, opened 2023-01-10
fix rpmlint  static-library-without-debuginfo error

anandaravuri:
-change static library  opae-c-archive  to shared library


tswhison (changes requested):
* Couple of things wrong here..  libopae-c-archive.so is made a duplicate of libopae-c.so.  Why not just get rid of the libopae-c-archive.so target altogether?
* You need to go through the exercise of proving that ASE still works with a .so library.

PR #2814, opened 2023-01-11
build: re-enable parallel builds

tswhison:
Add the necessary dependencies so that parallel builds work again. Here are some observed data points. There seems to be a point of diminishing returns around -j 8:

cmake .. -DCMAKE_BUILD_TYPE=Debug -DOPAE_BUILD_TESTS=ON -DOPAE_ENABLE_MOCK=ON -DOPAE_BUILD_LEGACY=ON

make -j 4  5m5.137s
make -j 8  4m21.589s
make -j 12 5m35.980s
make -j 16 6m11.187s

Ah, `nproc` reports 8 cpu's. That explains it.


michael-adler (approved):
Much better. Thank you!

anandaravuri (approved):

PR #2816, opened 2023-01-13
fpgasupdate: break from transferring on error

tswhison:
Add a check for status.value != 'idle' to the transferring loop. This allows the script to break from the transferring loop when the cancel sysfs node is written externally.


rweight (approved):
Looks good to me

anandaravuri (approved):

PR #2818, opened 2023-01-17
remove opae-c-archive static library

anandaravuri:
- OPAE-ASE module statically link opae-c-archive static library,
- Removed static library from OPAE-SDK , OPAE-ASE module dynamically likes opae-c library


tswhison:
@michael-adler could you please try this change out in your environment?  This is the one that changes from .a to .so for the libopae-c.  Thanks!

michael-adler (approved):
ASE still works after this. I think the static library was part of a test system that we never really used.

tswhison (approved):

PR #2826, opened 2023-01-18
opae vfio : fix static code analysis issue - Memory Leak

anandaravuri:


libraries/plugins/vfio/opae_vfio.c
```diff
@@ -351,7 +351,7 @@ int pci_discover(void)
 
 	if (gres) {
 		OPAE_MSG("vfio-pci not bound to any PCIe enpoint");
-		return 0;
+		goto free;
```
tswhison, line 354: I think you should set res to 0 here if you want to preserve the previous behavior.  It's initialized to 1 on 347.

tswhison (approved):

PR #2836, opened 2023-02-01
fix pybin11 build error

anandaravuri:
Signed-off-by: anandaravuri <ananda.ravuri@intel>

### DIFF
--- a/libraries/pyopaeuio/CMakeLists.txt
+++ b/libraries/pyopaeuio/CMakeLists.txt
@@ -34,8 +34,8 @@ if (OPAE_WITH_PYBIND11)
     set(SETUP_PY_IN "${CMAKE_CURRENT_SOURCE_DIR}/setup.py")
     set(SETUP_PY    "${CMAKE_CURRENT_BINARY_DIR}/setup.py")
 
-    #configure_file(${SETUP_PY_IN} ${SETUP_PY})
-    configure_file(${SETUP_PY})
+    configure_file(${SETUP_PY_IN} ${SETUP_PY})
+
 
     add_custom_command(
         OUTPUT ${OUTPUT}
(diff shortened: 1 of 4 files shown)


libraries/pyopaeuio/CMakeLists.txt
```diff
@@ -24,11 +24,19 @@
 ## ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
 ## POSSIBILITY OF SUCH DAMAGE.
 
+set(PYOPAE_PYBIND11_VERSION "2.2.4")
```
tswhison, line 27: Can we remove this definition of PYOPAE_PYBIND11_VERSION and the one from pyopae, and just use the global one defined here? https://github.com/OFS/opae-sdk/blob/master/CMakeLists.txt#L425

libraries/pyopaeuio/CMakeLists.txt
```diff
@@ -24,11 +24,19 @@
...
 if (OPAE_WITH_PYBIND11)
     set(INCLUDE_DIRS "${OPAE_INCLUDE_PATH}:${pybind11_ROOT}/include")
     set(LINK_DIRS "${LIBRARY_OUTPUT_PATH}")
     set(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/timestamp)
 
+    set(SETUP_PY_IN "${CMAKE_CURRENT_SOURCE_DIR}/setup.py")
+    set(SETUP_PY    "${CMAKE_CURRENT_BINARY_DIR}/setup.py")
+
+    #configure_file(${SETUP_PY_IN} ${SETUP_PY})
+    configure_file(${SETUP_PY})
```
tswhison, line 38: This configure_file call will modify the file in place in the source directory.  Will that present an issue?

tswhison: Actually, this approach will dirty the source tree whenever a build is done.  We should probably avoid this.

tswhison (approved):
Awesome. Thanks!

Issue #2841, opened 2023-02-10
PACSign fails on RHEL9 because of OpenSSL version 3.0.1


rchriste1:
Running PACSign on RHEL 9 and 9.1 (RockyLinux9) fails with:
2023-02-10 08:55:25,241 - PACSign.log - DEBUG - OpenSSL version "OpenSSL 3.0.1 14 Dec 2021" fails to match "OpenSSL 3.0.2 15 Mar 2022"
Error: Module: openssl.py, Function: _find_openssl_so, Line: 330
       Failed to find crypto library (libcrypto.so).

Updating package openssl-devel does not provide OpenSSL 3.0.2, which is the only one PACSign accepts of 3.x.x.

If I add ('libcrypto.so.3', 'OpenSSL 3.0.1 14 Dec 2021') to the list of versions, it works. However, I am wondering if it is necessary that PACSign requires a specific version among 3.x.x. New versions are coming frequently. I think 3.0.2 was added to PACSign because it is the one in Ubuntu 22.04.

tswhison:
Hello rchriste1, thanks for pointing this out.  PACSign does require a particular interface from libcrypto.so in order to work, but it seems to work well with version >= 1.1.1.  Would you be able to provide the patch you got to work in your environment?

tswhison:
Fixed by #2842.

PR #2842, opened 2023-02-13
Fix PACSign fails on RHEL9 because of OpenSSL version 3.0.1 #2841

rchriste1:
The check for openssl version is relaxed to allow all versions of libcrypto.so.3 Ubuntu 22.04 uses 3.0.2, RHEL 9.0 and 9.1 uses 3.0.1. Several higher versions already exists.

Signed-off-by: Roger Christensen <rc@silicom.dk>


tswhison:
Resolves #2841

PR #2846, opened 2023-02-16
vfio: changes in support of OpenCL MMD for gzip workload

tswhison:
Add a hash map implementation to libopaemem. The hash map implementation
is modeled after the implementation found in xfpga, but is a more general
implementation, operating on void pointers instead of buffer structs.

Use the hash map to replace the list-based buffer tracking in
libopaevfio to increase performance.

Removes the linked-list tracking for the buffer info and instead stashes
the pointer to libopaevfio's underlying buffer info struct in the wsid
output parameter. This eliminates a list walk in the buffer alloc/free
paths.

### DIFF
--- a/include/opae/hash_map.h
+++ b/include/opae/hash_map.h
@@ -23,6 +23,17 @@
 // CONTRACT,  STRICT LIABILITY,  OR TORT  (INCLUDING NEGLIGENCE  OR OTHERWISE)
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,  EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
+
+/**
+ * @file opae/hash_map.h
+ * @brief A general-purpose hybrid array/list hash map implementation.
+ *
+ * Presents a generic interface for mapping key objects to value objects.
+ * Both keys and values may be arbitrary data structures. The user supplies
+ * the means by which the hash of values is generated and by which the
+ * keys are compared to each other.
+ */
+
 #ifndef __OPAE_HASH_MAP_H__
 #define __OPAE_HASH_MAP_H__
 #include <stdint.h>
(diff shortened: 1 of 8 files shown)


libraries/libopaemem/hash_map.c
```diff
@@ -0,0 +1,291 @@
...
+	if (!item) {
+		ERR("malloc() failed");
+		return FPGA_NO_MEMORY;
+	}
+
+	key_hash = hm->key_hash(hm->num_buckets,
+				hm->hash_seed,
+				key);
+
+	list = hm->buckets[key_hash];
```
michael-adler, line 127: Check that key_hash is < num_buckets

libraries/libopaemem/hash_map.c
```diff
@@ -0,0 +1,291 @@
...
+	if (!list) {
+		// Bucket was empty. Fill it.
+		hm->buckets[key_hash] = item;
+		return FPGA_OK;
+	}
+
+	// Bucket not empty. Do we have a key collision?
+	while (list) {
```
michael-adler, line 135: Is this step really necessary?

libraries/libopaemem/hash_map.c
```diff
@@ -0,0 +1,291 @@
...
+			if (hm->value_cleanup)
+				hm->value_cleanup(list->value, hm->cleanup_context);
+			list->value = value; // Replace value only.
+			return FPGA_OK;
+		}
+		list = list->next;
+	}
+
+	// No key collision. Add item to the list.
+	item->next = hm->buckets[key_hash];
```
michael-adler, line 148: If you are going to do the while loop at 135 to check for duplicates, you might as well push the new item at the end of the list. It will work better for typical buffer release patterns.

libraries/libopaemem/hash_map.c
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	return true;
+}
+
+uint32_t opae_u64_key_hash(uint32_t num_buckets, uint32_t hash_seed, void *key)
+{
+	UNUSED_PARAM(hash_seed);
+	uint64_t remote_id = (uint64_t)key;
+	uint64_t hash = remote_id % 17659;
```
michael-adler, line 276: What if someone wants > 17659 buckets? It would be better to just set num_buckets to prime. The way you have it now with 16384 buckets, you've set up a higher probability of using buckets between 0-1274 because of the way hash from line 276 will then map to num_buckets in 277.

libraries/libopaemem/hash_map.c
```diff
@@ -0,0 +1,291 @@
...
+	key_hash = hm->key_hash(hm->num_buckets,
+				hm->hash_seed,
+				key);
+
+	list = hm->buckets[key_hash];
+	if (!list)
+		return FPGA_NOT_FOUND;
+
+	while (list) {
+		if (!hm->key_compare(key, list->key))
```
michael-adler, line 205: Calling a function through a pointer is slow in a critical routine, both here and in opae_hash_map_find(). I understand the value of setting a function to generalize hash comparison.

How about a compromise? It makes the code more complicated, but you could check whether hm->key_compare == opae_u64_key_hash. If it is, call opae_u64_key_hash directly instead of through a pointer. The branch predictor will handle that case correctly -- much easier than predicting a call through a pointer. You could even mark opae_u64_key_hash inline and avoid the jump inside the loop entirely.

libraries/libopaemem/hash_map.c
tswhison: In this generalized approach, add means add or replace.  I'd prefer to keep it this way.

libraries/libopaemem/hash_map.c
tswhison: Added a check.

libraries/libopaemem/hash_map.c
michael-adler: You could make it a parameter in the initializer to avoid an O(n) search at every insert if the parent can assert that all keys are unique.

tswhison: Added flags parameter to opae_hash_map_init() and flag OPAE_HASH_MAP_UNIQUE_KEYSPACE so that the list walk can be avoided.

libraries/libopaemem/hash_map.c
tswhison: Added logic to track the previous list item so that insert can be done at the end of the list.

libraries/libopaemem/hash_map.c
tswhison: Implemented this optimization both for key_hash and key_compare.

libraries/libopaemem/hash_map.c
tswhison: Revised the hash function logic and creating num_buckets with a prime.

include/opae/hash_map.h
```diff
@@ -0,0 +1,118 @@
...
+uint32_t opae_u64_key_hash(uint32_t num_buckets,
+			   uint32_t hash_seed,
+			   void *key)
+{
+	UNUSED_PARAM(hash_seed);
+	uint64_t ukey = (uint64_t)key;
+	return (uint32_t)(ukey % num_buckets);
+}
+
+static inline
```
michael-adler, line 100: I don't know what the compiler does when it sees this in the header file as a static. Won't you get a separate instance of opae_u64_key_compare in each module, so that the vfio code will pass in the address of it's private copy? Then the comparison in hash_map.c will compare against it's private copy and won't match, which will cause the test to see whether to invoke opae_u64_key_compare fail and it will call through the pointer.

michael-adler (commented):
Looking much better.

I doubt you need the optimization around calling opae_u64_key_hash. It won't be called frequently and I suspect you can simplify your code and just call indirect through the pointer to it.

It's opae_u64_key_compare that gets called inside a loop and matters more.

libraries/libopaevfio/opaevfio.c
```diff
@@ -1123,6 +1144,20 @@ STATIC int opae_vfio_init(struct opae_vfio *v,
 
 	mem_alloc_init(&v->iova_alloc);
 
+	result = opae_hash_map_init(&v->cont_buffers,
+				    19441, // num_buckets
```
anandaravuri, line 1148: #define   OPAE_HASH_MAPS_SIZE 19441, // num_buckets

libraries/libopaemem/hash_map.c
```diff
@@ -0,0 +1,348 @@
...
+	hm->buckets = (opae_hash_map_item **)
+			opae_calloc(num_buckets,
+				    sizeof(opae_hash_map_item *));
+	if (!hm->buckets) {
+		ERR("calloc() failed");
+		return FPGA_NO_MEMORY;
+	}
+
+	hm->num_buckets = num_buckets;
+	hm->hash_seed = hash_seed;
```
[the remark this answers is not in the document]
tswhison: It is an optional parameter and is not strictly required.

libraries/libopaemem/hash_map.c
```diff
@@ -0,0 +1,348 @@
...
+	hm->key_hash = key_hash;
+	hm->key_compare = key_compare;
+	hm->key_cleanup = key_cleanup;
+	hm->value_cleanup = value_cleanup;
+
+	return FPGA_OK;
+}
+
+STATIC opae_hash_map_item *
+opae_hash_map_alloc_item(void *key, void *value)
```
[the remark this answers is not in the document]
tswhison: NULL can be a valid key value.

libraries/libopaevfio/opaevfio.c
tswhison: Not all hash maps will be the same size.  This is implementation-specific.

michael-adler (approved):
Looks good.

anandaravuri (approved):

pajgaonk (approved):

PR #2862, opened 2023-03-14
rsu: extend --force to also force disabling AER even DPC is supported

rchriste1:
On some servers e.g. Dell R750 and XR12 even DPC is supported, hardware error occurs and server may crash, when running the rsu command on Arrow Creek or Lightning Creek cards. Disabling AER even DPC is supported can workaround these issues. This extension allows that by using the --force option.

There is an ongoing validation by Dell of the Arrow Creek card. Intel, Dell and Silicom are working together on this. This issue has been investigated for some time without finding the root cause. It would thus be good to have this option as a workaround.

### DIFF
--- a/python/opae.admin/opae/admin/fpga.py
+++ b/python/opae.admin/opae/admin/fpga.py
@@ -574,17 +574,23 @@ class fpga_base(sysfs_device):
         root_port = self.pci_node.root
         force = kwargs.pop('force', False)
         supports_dpc = root_port.supports_ecap('dpc')
-        if supports_dpc or force:
-            if supports_dpc:
-                self.log.debug('ECAP_DPC supported')
-            elif force:
-                self.log.warn('forcing rsu routine')
-            self.rsu_routine(available_image, **kwargs)
-        else:
-            if root_port.supports_ecap('aer'):
-                self.log.debug('ECAP_AER is supported')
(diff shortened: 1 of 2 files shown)


tswhison (approved):
Thank you!

python/opae.admin/opae/admin/fpga.py
```diff
@@ -574,17 +574,23 @@ def safe_rsu_boot(self, available_image, **kwargs):
...
-                self.log.debug('ECAP_DPC supported')
-            elif force:
-                self.log.warn('forcing rsu routine')
-            self.rsu_routine(available_image, **kwargs)
-        else:
-            if root_port.supports_ecap('aer'):
-                self.log.debug('ECAP_AER is supported')
+        if root_port.supports_ecap('aer'):
+            self.log.debug('ECAP_AER is supported')
+            if (not supports_dpc) or force:
```
anandaravuri, line 579: remove Round () brackets   if (not supports_dpc)
python code style

anandaravuri (approved):

python/opae.admin/opae/admin/fpga.py
rchriste1: thanks

anandaravuri (approved):

PR #2884, opened 2023-03-22
bist: fix segfault encountered during fuzz testing

tswhison:


binaries/fpgabist/bist/bist_D5005.c
```diff
@@ -228,21 +228,36 @@ static void parse_args(struct config *config, int argc, char *argv[])
 		case 'B':    /* bus */
 			if (NULL == tmp_optarg)
 				break;
+			endptr = NULL;
```
anandaravuri, line 231: can you check errno after calling strtoul()?
errno = 0;
config->bus = (int) strtoul(tmp_optarg, &endptr, 0);

if ( (endptr != tmp_optarg + strlen(tmp_optarg)) || (errno!=0 )) {
				fprintf(stderr, "Invalid bus number.\n");
				exit(1);
			}

tswhison: Shouldn't be necessary.  $ main strtoul says: ".. In particular, if *nptr is not '\0' but **endptr is '\0' on return, the entire string is valid."

PR #2901, opened 2023-03-28
fix opaevfiotest Segmentation fault

anandaravuri:
Segmentation fault1: opaevfiotest  0000:3b:00.1 dfh
PF1 and PF2 doesn’t support Bar 2 , opae_vfio_region_get() function ignores function return code  causes Segmentation fault
Fix: print_dfh() function exits gracefully  if opae_vfio_region_get() fails.

Segmentation fault2: opaevfiotest  0000:3b:00.0 nlb0
opae_vfio_region_get() and opae_vfio_buffer_allocate() functions ignores function return code  causes Segmentation fault
Fix: nlb0 () function exits gracefully   if opae_vfio_region_get() and opae_vfio_buffer_allocate() fails.

### DIFF
--- a/libraries/libopaevfio/opaevfiotest.c
+++ b/libraries/libopaevfio/opaevfiotest.c
@@ -47,7 +47,8 @@ void print_dfhs(struct opae_vfio *v)
 	struct fme_dfh *mmio = NULL;
 	uint8_t *p = NULL;
 
-	opae_vfio_region_get(v, 0, (uint8_t **)&mmio, NULL);
+	if (opae_vfio_region_get(v, 0, (uint8_t **)&mmio, NULL))
+		return;
 
 	printf("FME\n"
 	       "dfh:         0x%016lx\n"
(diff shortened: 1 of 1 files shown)


tswhison:
Could you provide a description of the root cause? ie what was responsible for the seg fault (I believe hardware configuration was involved).

libraries/libopaevfio/opaevfiotest.c
```diff
@@ -47,7 +47,8 @@ void print_dfhs(struct opae_vfio *v)
 	struct fme_dfh *mmio = NULL;
 	uint8_t *p = NULL;
 
-	opae_vfio_region_get(v, 0, (uint8_t **)&mmio, NULL);
+	if (opae_vfio_region_get(v, 0, (uint8_t **)&mmio, NULL))
+		return;
```
tswhison, line 51: Better to print something when returning early.  As it is now, this will be a silent failure.

libraries/libopaevfio/opaevfiotest.c
```diff
@@ -57,7 +58,8 @@ void print_dfhs(struct opae_vfio *v)
 	       mmio->fme_version, mmio->afu_id_low,
 	       mmio->afu_id_high, mmio->next_afu);
 
-	opae_vfio_region_get(v, 2, (uint8_t **)&mmio, NULL);
+	if (opae_vfio_region_get(v, 2, (uint8_t **)&mmio, NULL))
+		return;
```
tswhison, line 62: Print something when returning early so that this isn't a silent failure.

tswhison (approved):

PR #2931, opened 2023-04-25
host_exerciser computed sum of reads+writes incorrectly

michael-adler:
host_exerciser assumed read and write traffic were identical when computing lpbk and trput bandwidth. Use the actual count of reads and writes instead of just doubling writes.


michael-adler:
> Thanks @michael-adler! This should be backported to OFS 2023.1, correct?

Yes, I think that would be a good idea. Leaving that to you for proper tagging.

PR #2938, opened 2023-04-28
xfpga: fix a long-standing bug in the PR path

tswhison:
Tokens must not be destroyed until their associated handle is closed.

### DIFF
--- a/libraries/plugins/xfpga/reconf.c
+++ b/libraries/plugins/xfpga/reconf.c
@@ -117,13 +117,12 @@ STATIC fpga_result validate_bitstream(fpga_handle handle,
 }
 
 
-// open child accelerator exclusively - it not, it's busy!
-STATIC fpga_result open_accel(fpga_handle handle, fpga_handle *accel)
+// open child accelerator exclusively - if not, it's busy!
+STATIC fpga_result open_accel(fpga_handle handle, fpga_token *token, fpga_handle *accel)
 {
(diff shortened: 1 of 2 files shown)


libraries/plugins/xfpga/reconf.c
```diff
@@ -159,20 +158,15 @@ STATIC fpga_result open_accel(fpga_handle handle, fpga_handle *accel)
...
-		goto destroy_token;
+		goto free_props;
 	}
 
-	result = xfpga_fpgaOpen(token, accel, 0);
+	result = xfpga_fpgaOpen(*token, accel, 0);
 	if (result != FPGA_OK) {
 		OPAE_ERR("Could not open accelerator for given slot");
-		goto destroy_token;
+		goto free_props;
```
anandaravuri, line 167: Open fails , destroy token   xfpga_fpgaDestroyToken(&token)

anandaravuri: result = xfpga_fpgaOpen(token, accel, 0); Line 165 fails, accel is not valid 
line 405: if (accel && xfpga_fpgaClose(accel) != FPGA_OK) {
fpga_handle accel is not valid

tswhison: It's initialized to NULL on 266.

anandaravuri (approved):

PR #2939, opened 2023-05-01
rsu, pci_device: fix fpga.enum returns a list

rchriste1:
fpga.enum returns a list, using a for loop to handle 0 or more elements.


tswhison (approved):
Thanks, Roger.  I don't know why we didn't see this during our testing.

PR #2940, opened 2023-05-02
rpm build: add unrestricted build mode

tswhison:
When given 'rhel' or 'fedora', the create script would check for the presence of certain prerequisite packages before calling rpmbuild. The RPM spec file would also require certain packages in the 'BuildRequires: ' clauses. This posed a problem for certain distros that renamed their packages to be other than the standard package names. eg python3-pybind11 became python39-pybind11 in one instance.

This change adds a new 'unrestricted' mode to the create script which 1) skips the checking of prerequisite packages by the create script and 2) removes the 'BuildRequires: ' clauses from the RPM spec. When using 'unrestricted', it is the user's responsibility to ensure that the SDK will build correctly so that the RPM can be created.

### DIFF
--- a/packaging/opae/rpm/create
+++ b/packaging/opae/rpm/create
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright(c) 2020-2022, Intel Corporation
+# Copyright(c) 2020-2023, Intel Corporation
 #
 # Redistribution  and  use  in source  and  binary  forms,  with  or  without
 # modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


rchriste1 (approved):
Hi Tim, this change solves the issue. My test build on CentOS 8.2 works with this patch. The code looks fine to me. Thanks.

anandaravuri (approved):
Can you update install docs  https://github.com/OFS/opae-sdk/blob/master/doc/src/install_guide/installation_guide.md

tswhison:
> Can you update install docs https://github.com/OFS/opae-sdk/blob/master/doc/src/install_guide/installation_guide.md

Separate pull request..

PR #2947, opened 2023-05-18
Fix GitHub Actions build-debs and build-rpms

a-thompson2:
The current build-debs and build-rpms workflows do not upload the built artifacts, because they are trying to upload from the wrong directory.

This change corrects the directory, so the artifacts are uploaded.


tswhison (approved):
Cool, thanks!

tswhison (approved):

PR #2969, opened 2023-06-23
fix ofs.uio  log format

anandaravuri:


binaries/ofs.uio/uio/ofs_uio.py
```diff
@@ -728,7 +728,7 @@ def main():
             for i in range(int(args.peek_dump[1], 16)):
                 value = uio.peek(args.region_index, args.bit_size, addr)
                 LOG.info('peek(0x%x): 0x%x', addr, value)
-                addr = addr + int(args.bit_size)/8
+                addr = addr + int(args.bit_size/8)
```
tswhison, line 731: If args.bit_size can safely be divided by 8, do you need the int( ) conversion?

tswhison (approved):
👍

fpgamatt (approved):

PR #2975, opened 2023-06-27
ofs_uio.py: add pause_daemon support

fpgamatt:
Add support for a pause_daemon which is a SW work around to the RX pause configuration registers being initialized improperly in the E-tile. Since these registers can only be written when the ethernet link is established, the daemon tries to run forever, periodically writing the registers.


tswhison (approved):
LGTM

anandaravuri (approved):

PR #2984, opened 2023-07-12
opae.io: add vf_token subcommand

fpgamatt:
Add the vf_token subcommand to set the PCI Virtual Function (VF) token when the vfio-pci driver is bound to the Physical Function (PF).

### DIFF
--- a/binaries/opae.io/main.h
+++ b/binaries/opae.io/main.h
@@ -26,6 +26,8 @@
 
 #pragma once
 
+#include <iostream>
+#include <sys/ioctl.h>
 #include <opae/vfio.h>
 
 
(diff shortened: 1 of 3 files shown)


binaries/opae.io/main.h
```diff
@@ -215,6 +216,33 @@ struct vfio_device {
...
   }
+
+  #define VFIO_VF_TOKEN_LEN (16)
+  int set_vf_token(const char *vf_token)
+  {
+    struct vfio_device_feature *df;
+    int ret;
+
+    df = (struct vfio_device_feature *)
+      malloc(sizeof(struct vfio_device_feature) + VFIO_VF_TOKEN_LEN);
```
tswhison, line 227: Rather than malloc/free, could we use new/delete?

binaries/opae.io/main.h
```diff
@@ -215,6 +216,33 @@ struct vfio_device {
...
+
+    if (!df) {
+        printf("no mem\n");
```
tswhison, line 230: Could we print to std::cerr on error instead?

binaries/opae.io/main.h
```diff
@@ -215,6 +216,33 @@ struct vfio_device {
...
+
+    df->argsz = sizeof(df) + VFIO_VF_TOKEN_LEN;
+    df->flags = VFIO_DEVICE_FEATURE_SET | VFIO_DEVICE_FEATURE_PCI_VF_TOKEN;
+    memcpy(df->data, vf_token, VFIO_VF_TOKEN_LEN);
+
+    ret = ioctl(v_->device.device_fd, VFIO_DEVICE_FEATURE, df);
+    free(df);
+
+    if (ret)
+      printf("ioctl failed %d\n", errno);
```
tswhison, line 242: std::cerr instead?

tswhison (commented):
Just a few nits.

binaries/opae.io/main.h
fpgamatt: I'm not sure new/delete can be used here. We are allocating memory for variable length data structure. In this case the 16 bytes of token data are appended to the rest of the basic ioctl data structure.

tswhison: I think you could do something like
```c++
 df = (struct vfio_device_feature *) new uint8_t[sizeof(struct vfio_device_feature) + VFIO_VF_TOKEN_LEN];
```

tswhison (dismissed):
Now that I think of it, since we switched to new [ ], the check on 230 is no longer necessary.  The standard C++ new operator will raise an exception if the allocation can't be fulfilled.  Feel free to remove 230-233 if you feel so inclined.

tswhison (approved):

PR #2983, opened 2023-07-11
multi-channel mem_tg test 

robert-purcaru:
- Added multi-channel capability to mem_tg test, specified in "-m, --mem-channel" argument, allowing bandwidth to be measured for several channels communicating in parallel.
    - User can specify a list of integers representing the 0 indexed channels to be run, or 'all' to run on all channels enumerated in MEM_TG_CTRL register.
    - Each channel gets a thread spawned for it that runs the preexistent configuration and run routines.
    - The results of each thread are logged in the same format as the previous version.

PR #2985, opened 2023-07-18
mem_tg test channel synchronization and documentation

robert-purcaru:
- Added documentation/instructions for #2983 
- Adjusted tg_test.h do the starting write to each channel around the same time when multiple channels are used.
- Fixed bug where summary statements could collide after tests completed.


tswhison (approved):
👍

PR #3014, opened 2023-09-20
Feature: add a method for binding shared virtual addressing

michael-adler:
### Description
Add fpgaBindSVA() to allocate and bind PASIDs to FPGA ports. Shared virtual addressing requires hardware support, both in the CPU and on the FPGA.

fpgaBindSVA() depends on /dev/dfl-pci-sva from the dfl-pci driver. IOVA-based pinned DMA remains available with older drivers and hardware that does not support PASID.

OPAE behavior is unchanged when fpgaBindSVA() is not called.

### DIFF
--- a/libraries/plugins/vfio/opae_vfio.c
+++ b/libraries/plugins/vfio/opae_vfio.c
@@ -716,6 +716,9 @@ fpga_result __VFIO_API__ vfio_fpgaOpen(fpga_token token, fpga_handle *handle, in
 	_handle->mmio_base = (volatile uint8_t *)mmio;
 	_handle->mmio_size = size;
 
+	_handle->sva_fd = -1;
+	_handle->pasid = -1;
+
 	_handle->flags = 0;
 #if defined(__i386__) || defined(__x86_64__) || defined(__ia64__)
 #if GCC_VERSION >= 40900
@@ -763,11 +766,11 @@ fpga_result __VFIO_API__ vfio_fpgaClose(fpga_handle handle)
 		OPAE_ERR("invalid token in handle");
 	}
 
-	if (h->sva_fd) {
+	if (h->sva_fd >= 0) {
 		// Release PASID and shared virtual addressing
 		opae_close(h->sva_fd);
-		h->sva_fd = 0;
-		h->pasid = 0;
+		h->sva_fd = -1;
+		h->pasid = -1;
 	}
 
 	close_vfio_pair(&h->vfio_pair);
@@ -1608,7 +1611,9 @@ fpga_result __VFIO_API__ vfio_fpgaBindSVA(fpga_handle handle, uint32_t *pasid)
 
 	h = handle_check_and_lock(handle);
 	ASSERT_NOT_NULL(h);
-	ASSERT_NOT_NULL(pasid);
+
+	if (pasid)
+		*pasid = -1;
 
 	struct opae_vfio *v = h->vfio_pair->device;
 	if (!v || !v->cont_pciaddr) {
@@ -1616,9 +1621,10 @@ fpga_result __VFIO_API__ vfio_fpgaBindSVA(fpga_handle handle, uint32_t *pasid)
 		goto out_unlock;
 	}
 
-	if (h->pasid) {
+	if (h->pasid >= 0) {
 		// vfio_fpgaBindSVA() was already called. Return the same result.
-		*pasid = h->pasid;
+		if (pasid)
+			*pasid = h->pasid;
 		goto out_unlock;
 	}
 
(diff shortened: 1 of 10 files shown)


michael-adler:
@rweight and @matthew-gerlach - These are the user space changes that go along with the dfl-pci driver updates.

Note that the changes here will work with old drivers. They won't find /dev/dfl-pci-sva and will be fine with that.

libraries/plugins/vfio/opae_vfio.c
```diff
@@ -762,6 +763,13 @@ fpga_result __VFIO_API__ vfio_fpgaClose(fpga_handle handle)
 		OPAE_ERR("invalid token in handle");
 	}
 
+	if (h->sva_fd) {
+		// Release PASID and shared virtual addressing
+		opae_close(h->sva_fd);
+		h->sva_fd = 0;
```
tswhison, line 769: 0 is actually a valid file descriptor.  Could we use -1 as the closed sentinel for this?

libraries/plugins/vfio/opae_vfio.c
```diff
@@ -1590,6 +1598,63 @@ fpga_result __VFIO_API__ vfio_fpgaGetIOAddress(fpga_handle handle,
...
+{
+	vfio_handle *h;
+	char path[PATH_MAX];
+	int fd;
+	int err;
+	fpga_result res = FPGA_OK;
+
+	h = handle_check_and_lock(handle);
+	ASSERT_NOT_NULL(h);
+	ASSERT_NOT_NULL(pasid);
```
tswhison, line 1611: If we want to allow pasid to be NULL, this will prevent that. Better to add a NULL check at 1621 and 1642.

libraries/plugins/vfio/opae_vfio.c
```diff
@@ -1590,6 +1598,63 @@ fpga_result __VFIO_API__ vfio_fpgaGetIOAddress(fpga_handle handle,
...
+
+	// Currently, the dfl-pci driver provides support for allocating a PASID
+	// and binding it to the FPGA in the IOMMU. At some point, VFIO and the
+	// IOMMUFD will also support shared virtual addressing. When ready, this
+	// function can be updated to call libopaevfio.
+
+	// dfl-pci creates a device at /dev/dfl-pci-sva/<pci_addr> for all
+	// ports that support shared virtual memory. If the device isn't found
+	// then assume PASID is not supported, either by the host or the FPGA.
+	snprintf(path, sizeof(path), "/dev/dfl-pci-sva/%s", v->cont_pciaddr);
```
tswhison, line 1633: How is there a /dev/dfl-pci-sva/\<address\> if the device is currently bound to vfio-pci?

libraries/plugins/vfio/opae_vfio.c
```diff
@@ -1590,6 +1598,63 @@ fpga_result __VFIO_API__ vfio_fpgaGetIOAddress(fpga_handle handle,
...
+	fd = opae_open(path, O_RDONLY);
+	if (fd > 0) {
```
tswhison, line 1635:
```suggestion
	if (fd >= 0) {
```
0 is a valid file descriptor.

libraries/plugins/vfio/opae_vfio.c
```diff
@@ -1590,6 +1598,63 @@ fpga_result __VFIO_API__ vfio_fpgaGetIOAddress(fpga_handle handle,
...
+		// Request a shared virtual addressing. On success, the PASID is
+		// returned.
+		int res = opae_ioctl(fd, DFL_PCI_SVA_BIND_DEV);
```
tswhison, line 1638: Could we use a different name than res?  res is also defined on 1607.

tswhison (commented):
We're adding a new top-level API, so to be complete we would also need to add C++ and Python bindings support.  Just FYI, if you wanted to include the support here; else I can take it on.  There are also doc updates that need to happen.

michael-adler:
@tswhison - Thanks for going through this. I'll come back to it once the low-level device mechanism is set with @rweight. The general framework here will remain, but the details down inside opae_vfio.c are in flux.

libraries/plugins/vfio/opae_vfio.c
michael-adler: That's the "magic" of the dfl-pci-sva. No device is actually bound to it. The dfl-pci-sva driver is monitoring PCIe devices (not drivers) and instantiating pseudo-devices under /dev/dfl-pci-sva when a port is found that supports shared virtual memory.

Ideally, this feature would be provided by vfio/iommufd itself. Unfortunately, they are nowhere near ready. The dfl-pci-sva driver solves the problem until vfio is ready. When vfio is ready, we'll still have the top-level fpgaBindSVA() method but the implementation in libopaevfio will change -- hidden from user code.

tswhison: That's very cool.

michael-adler:
@tswhison - I'm confused by the failed test at opae-sdk/tests/opae-v/test_opae_vfio_c.cpp:3246. I think I didn't change anything relevant.

I think the test causes vfio_fpgaDestroyEventHandle() to close fd 0. Seems like a bad idea.

michael-adler:
@tswhison - I still think that opae-sdk/tests/opae-v/test_opae_vfio_c.cpp:3246 is a broken test since it closes fd 0. I exposed that bug by causing another test to do the same thing. My case is fixed in the latest commit by adding flags for pasid so that when the handle is initialized to 0 it doesn't look like there's a file open.

tswhison:
Thanks for pointing this out.  I will submit a fix.

tswhison (approved):

PR #3018, opened 2023-09-27
Feature: fpgaBindSVA() support for Python bindings

tswhison:
### Description
Extend the Python handle class to include a bind_sva() method.

### Tests run:
$ python3
\>\>\> import opae.fpga
\>\>\> props = opae.fpga.properties(type=opae.fpga.ACCELERATOR)
\>\>\> tokens = opae.fpga.enumerate([props])
\>\>\> h = opae.fpga.open(tokens[0])
\>\>\> print(h.bind_sva())


libraries/pyopae/pyhandle.cpp
```diff
@@ -175,3 +175,12 @@ const char *handle_doc_write_csr64() {
       csr_space: The CSR space to write from. Default is 0.
   )opaedoc";
 }
+
+const char *handle_doc_bind_sva() {
+  return R"opaedoc(
+    Bind IOMMU shared virtual addressing.
+    When PCIe PASID, ATS and PRS capabilities are enabled, some platforms
+    support binding the IOMMU to user space virtual addresses.
+    Returns the non-zero PASID when supported, else 0.
```
michael-adler, line 184: The comment about non-zero/zero value of pasid made me go back and look at the C++ implementation from the last check in. Two things:

1. Zero is, technically, a legal value. If no fpga_status is available then it would be better to return -1 on error.
2. It looks like the C++ bind_sva() method prints an error (and exits?) when the BindSVA() call returns an error. It should not. Return -1 instead, without printing an error. It's reasonable for an application to call BindSVA() to see whether SVA is supported and then deal with the consequences if it is not.

tswhison: > 1. -1 on error.
Hmm, but the pasid parameter is a uint32_t.  Is it the case the 0xffffffff is not a valid pasid?  If it is, should we change the parameter to be a signed type?

michael-adler: Signed would be fine. pasid is 20 bits -- defined by the PCIe standard.

tswhison: If 0 through 0xffffffff are all valid pasid values, then we could think about changing the pasid parameter to be a uint64_t and then use -1, 0xffffffffffffffffUL.

michael-adler: 32 is enough and 0xffffffff is not a valid pasid. If you think the type would be better as a signed int32 instead of unsigned, fine with me.

anandaravuri (approved):

PR #3026, opened 2023-10-12
Update: add --exclude option to pci_device unplug

tswhison:
### Description
In a hw design utilizing two PFs, the unplug command must provide an option to preserve
the PF that provides the PR capability; otherwise, PR will not function.

### Tests run:
Manual verification of --exclude functionality.

### DIFF
--- a/python/opae.admin/opae/admin/tools/pci_device.py
+++ b/python/opae.admin/opae/admin/tools/pci_device.py
@@ -215,8 +215,8 @@ class unplug(object):
         debug = args.debug
 
         exclude_node = None
-        if myargs.exclude:
-            exclude_devs = pcie_device.enum(myargs.exclude)
+        if args.exclude:
+            exclude_devs = pcie_device.enum(args.exclude)
             if exclude_devs:
                 exclude_node = exclude_devs[0].pci_node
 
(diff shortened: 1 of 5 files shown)


python/opae.admin/opae/admin/tools/pci_device.py
```diff
@@ -208,12 +208,18 @@ def add_subparser(subparser):
         unplug = subparser.add_parser('unplug')
         unplug.add_argument('-d', '--debug', action='store_true',
                             default=False, help='enable debug output')
-        #unplug.add_argument('-e', '--exclude', default=None,
-        #                    help='do not unplug the device specified here')
+        unplug.add_argument('-e', '--exclude', default=None, type=pci_devices,
+                            help='do not unplug the device specified here')
```
fpgamatt, line 212: Would we ever want to exclude more than one pci device?

tswhison: Hmm, I think we envisioned the context of this to be 1 workload PF and 1 PR PF. I can't imagine a situation where PR would require more than 1 device to exclude.  If you can think of one, then I can adjust the option to accept a list.

fpgamatt: If there were multiple PR regions (e.g. PF1 and PF2 independently PR'able), then we'd need multple --excludes. When that happens, we can extend --exclude for a list.

fpgamatt (approved):

anandaravuri (approved):

Issue #3024, opened 2023-10-11
(security vulnerability) use of python pickle module in opae.io python scripts


umairsiddiqui-digitek:
i want to point out the use of python pickle module in python script of opae.io. 
https://github.com/OFS/opae-sdk/blob/4cb9d59102415b6393b784502c7fd1d37366066b/binaries/opae.io/opae/io/utils.py#L54C4-L54C4

 As opae.io binary run as root, there is a (possible) security risk of arbitrary code execution during unpickling

https://docs.python.org/3/library/pickle.html

> Warning The pickle module is not secure. Only unpickle data you trust.
> It is possible to construct malicious pickle data which will execute arbitrary code during unpickling. Never unpickle data that could have come from an untrusted source, or that could have been tampered with.
> 
> Safer serialization formats such as [json](https://docs.python.org/3/library/json.html#module-json) may be more appropriate if you are processing untrusted data. See [Comparison with json](https://docs.python.org/3/library/pickle.html#comparison-with-json).
>

tswhison:
Thank you for pointing this out.  The pickle file itself is also protected by root permissions.
```bash
-rw-r--r--. 1 root root 66 Oct 11 09:37 /var/lib/opae/opae.io.pickle
```
If someone is able to inject malicious data into the pickle file, that implies that they already have escalated privileges (or else, how could they have written to the file?)  From that perspective, do you think there is still some issue?

I will investigate converting it to JSON format.

tswhison:
Resolved by #3028.

PR #3028, opened 2023-10-12
Update: use json to save opae.io driver state

tswhison:
### Description
opae.io init unbinds the target device from its driver and writes
a breadcrumb out to file so that the driver can be restored during
opae.io release. Previously, we were using the pickle module to do
this.

This github issue #3024 was filed, noting that pickle is prone to
code execution injection issues.  The proposal was to use json
instead.

#3024

### Tests run:
Manual verification using opae.io init followed by opae.io release.

### DIFF
--- a/binaries/opae.io/opae/io/utils.py
+++ b/binaries/opae.io/opae/io/utils.py
@@ -29,7 +29,6 @@ import errno
 import grp
 import json
 import os
-import pickle
 import pwd
 import re
 import struct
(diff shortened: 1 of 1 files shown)


michael-adler (approved):
Neat that replacing pickle with json is basically a switch of I/O packages, still using their dump/load methods.

anandaravuri (approved):

binaries/opae.io/opae/io/utils.py
```diff
@@ -51,7 +50,7 @@
 
 import libvfio
 
-PICKLE_FILE = '/var/lib/opae/opae.io.pickle'
+JSON_FILE = '/var/lib/opae/opae.io.json'
```
anandaravuri, line 53: we have add opae.io.json Spec file? required or not required ?

tswhison: Not required. We don't ship that file. It is created if it doesn't exist.

PR #3036, opened 2023-11-06
Fix: address issues pointed out by Python scanners

tswhison:
### Description
Our Python scanners point out issues regarding the use of
the subprocess module. The issues tend to be in 2 main
categories:
1) The use of subprocess.Popen() directly is discouraged.
Instead, the scanners recommend using run, call, or checked_call.

2) The use of any of the subprocess calls with shell=True 
is forbidden, because it is prone to code injection attacks.
Instead, the scanners require shell=False or omitting shell
altogether.

Python scans when preparing the release.


michael-adler:
@tswhison - It appears your branch is far behind opae-sdk master. It would be easier to validate if you rebase your branch.

michael-adler (approved):
Tested an ASE config

anandaravuri (approved):

anandaravuri (approved):

PR #3038, opened 2023-11-07
Update: add --force, --enable-sriov to opae.io

tswhison:
### Description
Add --force and --enable-sriov options to the subparser for opae.io init. The --force option enables controlling the force parameter to vfio_init() that was previously hard-coded to False.

The --enable-sriov option, when given, causes opae.io init to write 'Y' to /sys/module/vfio_pci/parameters/enable_sriov.

The default value for both options is False.

- [x] Document Update Required? (Specify FIM/AFU/Scripts)

### Tests run:
$ cat /sys/module/vfio_pci/parameters/enable_sriov
N
$ sudo opae.io init -e -d 0000:ca:00.0 tswhison:tswhison
$ cat /sys/module/vfio_pci/parameters/enable_sriov
Y

### DIFF
--- a/binaries/opae.io/opae/io/utils.py
+++ b/binaries/opae.io/opae/io/utils.py
@@ -123,9 +123,23 @@ def chown_pci_sva(pci_addr, uid, gid):
         os.chown(sva_bind_dev, uid, gid)
 
 
-def vfio_init(pci_addr, new_owner='', force=False):
+def enable_sriov(enable):
+    sriov = '/sys/module/vfio_pci/parameters/enable_sriov'
+    if os.path.exists(sriov):
+        LOG.info('Enabling SR-IOV for vfio-pci')
+        try:
+            with open(sriov, 'w') as outf:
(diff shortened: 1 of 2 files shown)


binaries/opae.io/opae/io/utils.py
```diff
@@ -123,9 +123,23 @@ def chown_pci_sva(pci_addr, uid, gid):
         os.chown(sva_bind_dev, uid, gid)
 
 
-def vfio_init(pci_addr, new_owner='', force=False):
+def enable_sriov(enable):
+    sriov = '/sys/module/vfio_pci/parameters/enable_sriov'
+    if os.path.exists(sriov):
```
fpgamatt, line 128: The Linux kernel folks would suggest trying to minimize indentation by something like the following:

if not os.path.exists(sriov):
    LOG.error("Cannot enable sriov because {sriov} does not exist")
    return FALSE

try:...

PR #3040, opened 2023-11-09
Feature:support  Continues mode and latency iterations to CXL host exe…

anandaravuri:
…rciser

   -Add Continues mode test
   -Support command line argument latency iterations and run scenario test in loop
   -Support command line argument cache line count


tswhison (approved):
lgtm

PR #3049, opened 2023-11-22
disallow factory image update if boot page is also factory

anandhv:
*PR Title should start with one of the following tags: [Fix]/[Feature]/[Style]/[Update]*

*[Fix]- Bug Fix*

*[Feature]- for new feature*

*[Style]- Grammar or branding fix*

*[Update]-For an update to an existing feature*

------------- *Keep everything below this line* -------------------------

### Description
*Describe the issue, update, change or fix and **why***

### DIFF
--- a/python/opae.admin/opae/admin/tools/fpgasupdate.py
+++ b/python/opae.admin/opae/admin/tools/fpgasupdate.py
@@ -52,6 +52,9 @@ import shutil
 from opae.admin.fpga import fpga
 from opae.admin.utils.progress import progress
 from opae.admin.version import pretty_version
+import pacsign
+from pacsign import database
+from opae.admin.sysfs import sysfs_device, sysfs_node
 
 if sys.version_info[0] == 2:
     input = raw_input  # noqa pylint: disable=E0602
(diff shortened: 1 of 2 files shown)


python/opae.admin/opae/admin/tools/fpgasupdate.py
```diff
@@ -52,6 +52,9 @@
 from opae.admin.fpga import fpga
 from opae.admin.utils.progress import progress
 from opae.admin.version import pretty_version
+import pacsign
+from pacsign import database
```
tswhison, line 56: Probably don't want to insert a reference to pacsign in fpgasupdate, because they are two distinct packages; and there has been some talk of removing pacsign from the opae-sdk repo.  If all you need is database.CONTENT_FACTORY, then just create a CONTENT_FACTORY variable in fpgasupdate to avoid the dependency.

anandhv: Thanks, I have implemented this.

tswhison (approved):
👍

PR #3059, opened 2023-12-13
fix: remove clx he loopback sub command

anandaravuri:
CXL host exerciser doesn’t support loopback feature, so remove lpbk sub command


fpgamatt (commented):
Please update the commit message as to why loopback sub command is being removed.

anandaravuri:
> Please update the commit message as to why loopback sub command is being removed.

Ananda: updated commit meassge with comments

fpgamatt:
I don't see "CXL host exerciser doesn’t support loopback feature, so remove lpbk sub command" in the commit message. I even pulled your branch locally and checked.

tswhison (approved):
Thanks for the updated message.

PR #3060, opened 2023-12-14
FIX: flashupdate.py fix for intel PAC n3000 cards - boot_page

hwspeedy:
Issue:
fpgasupdate.py does not work on intel PAC n3000 devices due to added check for updating FPGA factory image when running factory image.

Found:
The boot_page is only implemented in PMCI driver not the SPI driver.

Changed:
Checked for SPI bus:
Seems that a check for spi_bus is used several places in the fpga.py.  I do not have knowledge - if the Intel PAC n3000 NIOS can supply which fpga image loaded in another way, so just disabled the factory update check for SPI interfaces.

Test: 
Success - programmed user image on Intel PAC n3000.


chipb-intel:
@tk-x86 -- Maybe something you could look at this week with Tim?

PR #3062, opened 2023-12-15
fix:set continuous mode bit for write hit/miss cases tests

anandaravuri:
CXL HE write hit/miss cases, sets continuous mode bit in rd_cfg instead of wr_cfg.
 set continuous mode bit write hit/miss case tests.

 Forced stop  control register ResetL bit is low causes timeout failures
 set ResetL too high for forced stop


tswhison:
Could you provide some description in the commit message indicating why the change is required?

fpgamatt (commented):
Can add to the commit message as to why this needs to be done and why the change involving removing setting he_rw_cfg_.continuous_mode_enable?

anandaravuri:
> Can add to the commit message as to why this needs to be done and why the change involving removing setting he_rw_cfg_.continuous_mode_enable?

Ananda: updated commit meassge with comments

PR #3066, opened 2024-01-04
fix: CXL host exerciser  segmentation fault non-root user mode

anandaravuri:
-CXL host exerciser  segmentation fault occurs in non-root user mode,
  exit application if clx device open fails.
 - CXL host exerciser  running pointer tests show wrong "data ping pong tests", Change to "Running pointer test started ......"


fpgamatt (commented):
Please separate this into to two commits. Change the commit message regarding segmentation fault of non-root user to something like "properly handle failing to open device file".

KrishnaSimmadhari (commented):
Looks good to me

PR #3068, opened 2024-01-04
fix: CXL host exerciser running pointer test output wrong  string

anandaravuri:
-CXL host exerciser running pointer tests show wrong "data ping pong tests".
   Change to "Running pointer test started ......"


fpgamatt (approved):
As Peter mentioned, you could have force pushed your changes to the same branch. You also can have a pull request with more than one commit.

KrishnaSimmadhari (approved):

PR #3069, opened 2024-01-04
[Feature] - Add fpgaGetChildren()

michael-adler:
### Description
The new method returns handles of children opened as a side effect of opening a parent AFU.

### DIFF
--- a/include/opae/access.h
+++ b/include/opae/access.h
@@ -73,6 +73,35 @@ fpga_result fpgaOpen(fpga_token token, fpga_handle *handle,
 		     int flags);
 
 /**
+ * Extract the handles of children of a previously fpgaOpen()ed resource.
+ * Only AFUs with feature parameters that name child AFU GUIDs will have
+ * children.
+ *
+ * Child AFU handles may be used to connect to child-specific MMIO regions.
+ * There is no need to close child handles with fpgaClose(). The handles
+ * will be closed automatically when the parent is closed. Child handles
+ * may not be passed to fpgaPrepareBuffer(). All shared memory management
+ * must be associated with the parent.
+ *
+ * @param[in]  handle       Handle to previously opened FPGA object
+ * @param[in]  max_children Maximum number of handles that may be returned
+ *                          in the `children` array.
+ * @param[out] children     Pointer to an array of child handles currently
+ *                          open. When NULL or `max_children` is 0, the
+ *                          number of children will still be returned in
+ *                          `num_children`.
+ * @param[out] num_children Number of children belonging to the parent.
+ *                          This number may be higher than `max_children`.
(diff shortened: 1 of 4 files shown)


michael-adler:
Please comment on this proposal. I think it's the only public API change needed.

When a parent AFU discovers and opens children, SW may need to get the handles of the children in order to access MMIO space of individual children.

I haven't yet added code to wrap the array of child handles returned, mostly because I don't know how to deallocate the wrapped handles on fpgaClose() of the parent. I'm thinking it would be best not to require that the parent close the children explicitly. Is it possible to destroy the child wrappers automatically?

include/opae/access.h
```diff
@@ -72,6 +72,35 @@ extern "C" {
...
+ * @param[in]  handle       Handle to previously opened FPGA object
+ * @param[in]  max_children Maximum number of handles that may be returned
+ *                          in the `children` array.
+ * @param[out] children     Pointer to an array of child handles currently
+ *                          open. When NULL or `max_children` is 0, the
+ *                          number of children will still be returned in
+ *                          `num_children`.
+ * @param[out] num_children Number of children belonging to the parent.
+ *                          This number may be higher than `max_children`.
+ * @returns                 FPGA_OK on success.
```
anandaravuri, line 95: If child AFU GUID not available or present,  fpgaGetChildren() going to return FPGA_NOT_FOUND ?  
fpga_handle (OPAE FPGA resources Handle ) is FPGA_ACCELERATOR type ?

michael-adler: No, fpgaOpen() will fail. I actually didn't want to expose child handles at all, since they are just needed for MMIO. Is there something better we should do? The only thing I could think of required adding a child ID parameter to all the MMIO methods, which seems like a bad idea.

Assuming we go with fpgaGetChildren(), what do you think about the problem of wrapping child handles and releasing them automatically? Is that possible?

fpgamatt:
For symmetry, wouldn't it better for fpgaClose() to close all the children, since fpgaOpen() opens all the children?

michael-adler:
> For symmetry, wouldn't it better for fpgaClose() to close all the children, since fpgaOpen() opens all the children?

fpgaClose() will actually close all the internal handles from fpgaOpen(). The messy part here is that fpgaGetChildren() is a deliberate exposure of the child handles, including creating wrappers. That exposure allocates a shell-level wrapper around the internal handle and that wrapper has to be deallocated by something. Either it has to be tracked inside the shell or the user has to do it. I'm worried that a user might call fpgaClose() on one of these wrappers and then would be in a race between the parent wrapper being deallocated and the exposed child. Declaring that the user has to do it after explicitly exposing a child handle avoids that race and simplifies the shell's data structures. It seems safer to me.

PR #3077, opened 2024-01-16
[Fix] SEGV when passing NULL buf_addr to fpgaPrepareBuffer()

michael-adler:
### Description
Mistaken pointer dereference. buf_addr is allowed to be NULL when testing the fpgaPrepareBuffer() capabilities.

Resolves a regression introduced in commit 4d757b6d.


boelkrug-intel (approved):
Looks good to me. Ship it!

pajgaonk (approved):

PR #3090, opened 2024-01-23
Improve VFIO device enumeration performance

michael-adler:
### Description
- Use results of earlier device walks instead of reopening devices.
- Add a lightweight check for busy devices.

### DIFF
--- a/include/opae/vfio.h
+++ b/include/opae/vfio.h
@@ -163,6 +163,26 @@ extern "C" {
 #endif
 
 /**
+ * Confirm that a device is available (not already open).
+ *
+ * Opens the PCIe device corresponding to the address given in pciaddr.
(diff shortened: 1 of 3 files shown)


michael-adler:
@anandaravuri and @pcolberg --

This dramatically improves performance of fpgaEnumerate(), especially on systems with a lot of ports. I tested on a machine with about 20 ports and an AFU with 2 children. Opening the AFU dropped from 12 to 2 seconds.

There are two changes you should check:
1. Check whether a port is busy by just opening the vfio group instead of going through the whole container setup and initialization.
2. Discover tokens for a device only once. Once walked, future references to the device use the old token list.

I'm confident about 1. Less so about 2.

anandaravuri (approved):

PR #3095, opened 2024-02-07
opae.io: Fix error handling when binding the vfio-pci driver fails

anandaravuri:
The 'opae.io init' command can fail to bind the vfio-pci driver when the IOMMU is disabled, leaving the system seemingly unable to bind another driver. The fix is to check if the IOMMU is enabled before even attempting to bind the vfio-pci driver.


fpgamatt:
Can we improve the commit message to something like the following?

opae.io: Fix error handling when binding the vfio-pci driver fails

The 'opae.io init' command can fail to bind the vfio-pci driver when the IOMMU is disabled, leaving the system seemingly unable to bind another driver. The fix is to check if the IOMMU is enabled before even attempting to bind the vfio-pci driver.

PR #3098, opened 2024-02-09
opae.io: Fix error system exit exception value

anandaravuri:
python os exception EX_NOTFOUND doesn’t supported, change to exception code to 1


anandaravuri:
> Thanks @anandaravuri! Could you clarify the commit message to say that EX_NOTFOUND is not available on Linux? It is valid Python, but not available on all platforms.

[Ananda]
opae.io init -d 0000:38:00.1 root:root
Binding to vfio-pci will fail because IOMMU is disabled.
AttributeError: module 'os' has no attribute 'EX_NOTFOUND'

At:
  /usr/lib/python3/dist-packages/opae/io/utils.py(150): vfio_init
  <string>(84): execute
  <string>(31): __call__
  <string>(481): main
  <string>(484): <module>

Replaced os.EX_NOTFOUND to 1 , opae.io exits with error code 1

PR #3117, opened 2024-04-11
hssi: suggest hssi{loopback,stats} when eth interface is absent

pcolberg:
While the D5005 FPGA card employs a separate kernel driver (s10hssi) to configure loopback and provide statistics via a network device, newer FPGA cards expose a UIO device used with hssiloopback and hssistats.

Closes: https://hsdes.intel.com/appstore/article/#/22019839240

Cc: @badanur

### DIFF
--- a/samples/hssi/hssi_100g_cmd.h
+++ b/samples/hssi/hssi_100g_cmd.h
@@ -416,7 +416,9 @@ public:
 
     if (eth_ifc == "") {
       std::cout << "No eth interface, so not "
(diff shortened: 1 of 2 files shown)


fpgamatt:
This change is helpful because it adds information to a weird error message, but can't we do better? Can't the hssi command figure out if the hssi in question is d5005 or a uio? That way no weird message would need to be printed out in the first place.

anandaravuri (approved):

PR #3120, opened 2024-04-15
Fpgamatt/subdevid 0001

fpgamatt:
*PR Title should start with one of the following tags: [Fix]/[Feature]/[Style]/[Update]*

*[Fix]- Bug Fix*

*[Feature]- for new feature*

*[Style]- Grammar or branding fix*

*[Update]-For an update to an existing feature*

------------- *Keep everything below this line* -------------------------

### Description
*Describe the issue, update, change or fix and **why***


fpgamatt:
The functionality moved to board_common exists on any board that has qsfp(s) and hssi subsystems.

We could call these boards something like pci_jtag_devkit.

michael-adler:
Can someone please explain why this very invasive change is needed and why something simpler won't work? I see complexity being added both here and to the HW side.

fpgamatt:
I would not classify this change as invasive. We are just adding a new board to the board plug-in framework of opae. This new board is a generic PCI devkit without a board BMC and is jtag only. Without this change, boards were incorrectly identifying themselves, and then opae would in turn mishandle the misidentified board.

michael-adler:
So is the rule: set subdevice ID to 1 if there is no BMC and to 0x1771 if there is a BMC?

fpgamatt:
There are currently quite a few subdevice ids being used. The intent is help identify the board with just lspci, and to address concerns from the kernel community of having just a generic DFL Device id with out any subdevice info. The subdevice of 0x1771 is actually a n6001, and 0x1770 is a n6000. 0x17d4 is used for C6100.

fpgamatt:
can I get an approve so this can be merged into master. The RTL has been merged.

michael-adler:
> can I get an approve so this can be merged into master. The RTL has been merged.

I can't help there. Not a code owner.

anandaravuri (approved):

PR #3121, opened 2024-04-24
host_exerciser: fix ASE simulation

michael-adler:
Don't check EMIF calibration when running in ASE. The required access token is not available.


fpgamatt (approved):
Thanks for a very clear commit message.

PR #3132, opened 2024-06-17
hssi: increate timeout to 1 second

michael-adler:
Some operations are taking longer than the current timeout, such as putting an HSSI port in loopback. Timeout is a failure, so waiting a little long isn't a problem. Actual completion observed has been over 100ms, so chose 1 second.

The current timeout logic could be improved but is not changed in this commit. Time passed is calculated as the sum of sleep times, which assumes MMIO reads are free and sleep is exactly the requested wait.

### Tests run:
Set loopback on 8x25 interfaces in iseries-dk, n6001


binaries/hssi/ethernet/hssicommon.py
```diff
@@ -53,8 +53,10 @@
 # mailbox register poll interval 1 microseconds
 HSSI_POLL_SLEEP_TIME = 1/1000000
 
-# mailbox register poll timeout 500 microseconds
-HSSI_POLL_TIMEOUT = 5/10000
+# Mailbox register poll timeout 1 second. Responses are typically much
+# faster than one second, but in rare cases can take a while. No reason
+# to make this short since timeout is generally a failure.
+HSSI_POLL_TIMEOUT = 1
```
fpgamatt, line 59: In the Linux Kernel the units of the timeout are usually given in the macro name
HSSI_POLL_TIMEOUT_MS or HSSI_POLL_TIMEOUT_SEC would be clearer.

fpgamatt (approved):

PR #3145, opened 2024-12-19
pci_device: make subprocess logging a debugging message and not error

michael-adler:


fpgamatt (commented):
commit message should say "not error" instead of "not info" to match the code change.

PR #3171, opened 2025-07-24
Update type/versions supported by st_dbg_ip_driver.c

fpgamatt:
The type and version fields in the dbg-ip were updated recently. Add support for origin and new type and versions.

### DIFF
--- a/binaries/mmlink/remote_dbg/streaming/st_dbg_ip_driver.c
+++ b/binaries/mmlink/remote_dbg/streaming/st_dbg_ip_driver.c
@@ -25,6 +25,7 @@
 // POSSIBILITY OF SUCH DAMAGE.
 #include <string.h>
 #include <stdint.h>
+#include <stdio.h>
 #include "st_dbg_ip_driver.h"
 #include "st_dbg_ip_allocator.h"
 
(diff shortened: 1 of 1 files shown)


binaries/mmlink/remote_dbg/streaming/st_dbg_ip_driver.c
```diff
@@ -283,15 +284,30 @@ int get_mgmt_support() {
     }
 }
 
+struct ver_type {
+    unsigned long type;
+    unsigned long ver;
+};
+struct ver_type supported_ver_types[] = {
+    { SUPPORTED_TYPE, SUPPORTED_VERSION, },
+    { 0x15244444d, 1, },
```
michael-adler, line 293: Should there be a name for these?

fpgamatt: I thought about having #defines for these numbers, but since they are only used in one place, I just put the numbers in the array.

michael-adler: Isn't there a named device associated with it that would be good to encode?

fpgamatt: As far as I know, the dbg_ip is the same, but the version and type numbers were updated.

michael-adler (approved):

